PR #56, opened 2025-08-15
PAE-127: define and test auditing

carlesandres:
Ticket: [PAE-127](https://eaflood.atlassian.net/browse/PAE-127)

## Description

- Documented how audits work in CDP
- Send audit event when an email has been sent successfully
- Allow to auditing to be disable via and env var

[PAE-127]: https://eaflood.atlassian.net/browse/PAE-127?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/common/helpers/mongodb.test.js
+++ b/src/common/helpers/mongodb.test.js
@@ -13,10 +13,6 @@ describe('#mongoDb', () => {
       await server.initialize()
     })
 
-    afterAll(async () => {
-      await server.stop({ timeout: 10000 })
-    })
-
     test('Server should have expected MongoDb decorators', () => {
       expect(server.db).toBeInstanceOf(Db)
       expect(server.mongoClient).toBeInstanceOf(MongoClient)


src/common/enums/event.js
```diff
@@ -22,3 +22,11 @@ export const LOGGING_EVENT_ACTIONS = {
   START_FAILURE: 'start_failure',
   START_SUCCESS: 'start_success'
 }
+
+export const AUDIT_EVENT_CATEGORIES = {
+  EMAIL: 'email'
+}
+
+export const AUDIT_EVENT_ACTIONS = {
+  EMAIL_SENT: 'email_sent'
```
nijk, line 31: Nitpick: I wonder if an action of `sent` would be sufficient?

src/common/helpers/notify.test.js
```diff
@@ -23,6 +26,10 @@ vi.mock('./logging/logger.js', () => ({
   })
 }))
 
+vi.mock('@defra/cdp-auditing', () => ({
+  audit: (...args) => mockAudit(...args)
+}))
```
nijk, line 31: Nitpick: I appreciate you're following convention here, but I think for terseness you can just do this:

```suggestion
vi.mock('@defra/cdp-auditing', () => ({
  audit: mockAudit
}))
```

src/common/helpers/notify.js
```diff
@@ -32,6 +35,14 @@ async function sendEmail(templateId, emailAddress, personalisation = {}) {
     await notifyClient.sendEmail?.(templateId, emailAddress, {
       personalisation
     })
+    audit({
+      event: {
+        category: AUDIT_EVENT_CATEGORIES.EMAIL,
+        action: AUDIT_EVENT_ACTIONS.EMAIL_SENT
+      },
+      template_id: templateId,
+      email_address: emailAddress
```
nijk, line 44: Nitpick: Can I just check that we're explicitly not putting the these values in `data` or `context` object? If so, what were the reasons for that decision?

CONTRIBUTING.md
```diff
@@ -376,6 +376,25 @@ therefore helping yourself and other engineers.
...
+For those operations requiring an audit trial, we use the [CDP Auditing library](https://github.com/DEFRA/cdp-libraries/tree/main/packages/cdp-auditing#readme).
+
+The library provides a simple `audit` function accepts either a string or a serialisable object.
+
+```js
+import { audit } from '@defra/cdp-auditing'
+audit('This is a test of auditing')
+// {"log.level":"audit","time":"2025-07-04T11:46:33.278Z","msg":"This is a test of auditing"}
+
+audit({ id: '1234', event: 'login', outcome: 'denied' }, 'Login failed')
```
nijk, line 390: I wonder if these examples should follow our implementation in code a little more closely and I imagine that we'll only be using the serialisable object approach

CONTRIBUTING.md
```diff
@@ -376,6 +376,25 @@ therefore helping yourself and other engineers.
...
+// {"log.level":"audit","time":"2025-07-04T11:46:33.278Z","id":"1234","event":"login","outcome":"denied","msg":"Login failed"}
+```
+
+Behind the scenes, audits are just logs with a "log.level" field set to "audit", which will be used to send them to an audit stream instead of the same OpenSearch pipeline with the other logs.
+
+The library also provides an `enableAuditing` function that accepts a boolean to easily switch auditing on and off.
```
nijk, line 396: Is the any reason you can think of to turn this into a feature flag or ENV var? For instance, might we want to not send audit logs from specific environments?

nijk (commented):
Nice work, a few questions and thoughts for you

src/common/helpers/notify.js
carlesandres: It makes sense to put our custom payload in such an object. I've changed it now. 
I think we should we standardise it for this project and document it.

CONTRIBUTING.md
carlesandres: I've made those changes.

PR #90, opened 2025-09-04
PAE-205: Add github actions to run backend journey tests on each PR

atzn-dev:
## Description

This PR would enforce the backend journey tests to run on each Pull Request that is made. This is to ensure that the integration tests are not failing prior to merging to main.


nijk (approved):
Nice work 👏

PR #1040, opened 2026-04-09
PAE-1184: action items from ADR-30

dansweeting:
Ticket: [PAE-1184](https://eaflood.atlassian.net/browse/PAE-1184)
## Summary

Follow-up from [ADR 0030](https://github.com/DEFRA/epr-re-ex-service/blob/PAE-1184-registered-only-edge-cases-adr/docs/architecture/decisions/0030-registered-only-edge-cases.md) (registered-only edge cases). Three low-risk improvements that don't require product decisions:

- **Fix `AccreditationOther` typedef**: the status union listed `'archived'` which doesn't exist in `REG_ACC_STATUS`. Replaced with `'cancelled'` to match the domain model.
- **Add `diagnostics.wasteReceivedRecordsExcluded` to `aggregateReportDetail`**: counts received-type waste records excluded because they lack the expected date field. This makes the silent data-loss path observable — after a registered-only → accredited transition, historical records have `MONTH_RECEIVED_FOR_*` but the accredited category looks up `DATE_RECEIVED_FOR_*`, so those records are silently dropped from reports.
- **Log warning on date field mismatch**: the GET report-detail route logs at warn level when excluded records are detected, including org/registration context.
- **Add transition tests**: three new tests in `aggregate-report-detail.test.js` that explicitly demonstrate and document the data-loss behaviour for both reprocessor and exporter transitions, plus an integration test in `get-detail.test.js` confirming the diagnostics surface through the API.

[PAE-1184]: https://eaflood.atlassian.net/browse/PAE-1184?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/domain/organisations/accreditation.js
```diff
@@ -66,7 +66,7 @@
 /**
  * @typedef {AccreditationBase & {
  *  accreditationNumber?: string;
- *  status: 'created'|'rejected'|'archived';
+ *  status: 'created'|'rejected'|'cancelled';
```
loveboat: Could this be a `Pick<Status, 'created'|'rejected'|'cancelled'>` (or where they're from) to keep them in sync with the origin of these status values

src/reports/domain/aggregation/aggregate-report-detail.js
```diff
@@ -126,6 +104,19 @@ export function aggregateReportDetail(
...
 
+  // Count received records excluded because they lack the expected date field.
+  // This could potentially happen after a registered-only organisation becomes
+  // accredited (or vice versa): historical registered-only records have MONTH_RECEIVED_FOR_*
+  // but the accredited category looks up DATE_RECEIVED_FOR_*. See ADR 0030,
+  // Finding 3. Only wasteReceived needs this check — all other sections use
+  // the same date field name regardless of operator category.
+  const wasteReceivedRecordsExcluded = wasteRecords.filter(
+    (r) =>
+      r.type === 'received' &&
```
loveboat: is there a constant we can use for 'received' ?

src/reports/domain/aggregation/aggregate-report-detail.js
```diff
@@ -126,6 +104,19 @@ export function aggregateReportDetail(
...
+      wasteReceivedDateField &&
+      r.data[wasteReceivedDateField] == null
```
loveboat: can we use `===` instead of `==` ?

note to self: add lint rule for `eqeqeq`

src/reports/domain/aggregation/aggregate-report-detail.js
```diff
@@ -146,7 +137,51 @@ export function aggregateReportDetail(
...
+}
+
+/**
+ * Slices waste records into per-section record sets, each filtered to those
+ * whose section date field falls within [startDate, endDate].
+ *
+ * @param {import('#domain/waste-records/model.js').WasteRecord[]} wasteRecords
+ * @param {{ wasteReceived?: string, wasteExported?: string, wasteRepatriated?: string, wasteSentOn?: string }} sectionDateFields
+ * @param {string} startDate
+ * @param {string} endDate
```
loveboat: do we have a better type than string for date?

src/reports/routes/get-detail.js
```diff
@@ -49,6 +49,22 @@ export const reportsGetDetail = {
       period
     })
 
+    if (
+      'diagnostics' in report &&
```
loveboat, line 56: is the a constant for this key?

src/domain/organisations/accreditation.js
dansweeting: bd3eb993

src/reports/domain/aggregation/aggregate-report-detail.js
dansweeting: bd3eb993

src/reports/domain/aggregation/aggregate-report-detail.js
dansweeting: bd3eb993

src/reports/domain/aggregation/aggregate-report-detail.js
dansweeting: The project-wide convention is `@param {string}` with an `ISO date string (YYYY-MM-DD)` descriptor (see `aggregate-waste-exported.js`, `filter-records-by-date.js`, `generate-reporting-periods.js`). Tightened the JSDoc on `sliceRecordsByPeriod` to match.

src/reports/routes/get-detail.js
dansweeting: The `'diagnostics' in report` check was a type discriminator rather than a property lookup: `fetchOrGenerateReportForPeriod` returns `Report | AggregatedReportDetail`, and stored `Report`s don't carry diagnostics. Switched to optional chaining (`report.diagnostics?.wasteReceivedRecordsExcluded ?? 0`), which handles both shapes cleanly without a string literal.

dansweeting: Correction to my previous reply: the optional-chaining swap broke `tsc`. `fetchOrGenerateReportForPeriod` returns `Report | AggregatedReportDetail`, and the `'diagnostics' in report` check was doing double duty as a TypeScript type discriminator, narrowing to the variant that carries both `diagnostics` **and** `operatorCategory`. Without the `in` narrowing, `report.operatorCategory` on line 59 fails type-checking. Restored the check in `a57f2a87` with a comment explaining why it is load-bearing.

loveboat (approved):

PR #1045, opened 2026-04-10
PAE-1323: Per-row ORS approval status in report aggregation

GraemeF:
Ticket: [PAE-1323](https://eaflood.atlassian.net/browse/PAE-1323)

## Summary

Adds per-row overseas site approval status to the report aggregation layer, so the frontend can conditionally show an Approved column for accredited exporters.

### Expose validFrom in ORS details map

- `getOrsDetailsMap` now includes `validFrom: site?.validFrom ?? null` alongside the existing `siteName` and `country` fields
- JSDoc type annotations updated to reflect the new `validFrom: Date|null` field

### Per-row approval status in aggregation

- Extract shared `isOrsApprovedAtDate(validFrom, dateOfExport)` predicate to `src/overseas-sites/domain/approval.js` — single source of truth for the validFrom business rule, used by both waste balance classification and report aggregation
- `aggregateWasteExported` now accepts `operatorCategory` (plumbed from `aggregateReportDetail`)
- Rename `hasApprovedSite` → `isResolvedSite` (checks DB presence via siteName, not approval)
- Resolved overseas sites now carry a per-row `approved` boolean:
  - Accredited exporters: `validFrom <= DATE_OF_EXPORT`
  - Registered-only exporters: always `false`
- Sites group by `(orsId, approved)` — a single ORS ID can produce two rows when records straddle the validFrom boundary
- Unresolved sites (Table 2 / `unapprovedOverseasSites`) unchanged
- Joi schema updated with `approved: Joi.boolean().required()`
- `received-loads-for-export.js` refactored to use the shared predicate

## Breaking change

`overseasSites` entries now include `approved: boolean`. Reports persisted in MongoDB before this change won't have the field — **test environment report data needs wiping** after deploy.

[PAE-1323]: https://eaflood.atlassian.net/browse/PAE-1323

### DIFF
--- a/src/reports/domain/aggregation/aggregate-waste-exported.js
+++ b/src/reports/domain/aggregation/aggregate-waste-exported.js
@@ -98,7 +98,7 @@ function calculateTonnageReceivedNotExported(
  * @param {import('#domain/waste-records/model.js').WasteRecord[]} params.wasteReceivedRecords
  * @param {string} params.startDate - ISO date string (YYYY-MM-DD)
  * @param {string} params.endDate - ISO date string (YYYY-MM-DD)
- * @param {Map<string, { siteName: string|null, country: string|null, validFrom: string|null }>} [params.orsDetailsMap]
+ * @param {Map<string, { siteName: string|null, country: string|null, validFrom: Date|null }>} [params.orsDetailsMap]
  */
 export function aggregateWasteExported({
   wasteExportedRecords,
--- a/src/reports/domain/aggregation/aggregate-waste-exported.js
+++ b/src/reports/domain/aggregation/aggregate-waste-exported.js
@@ -7,6 +7,7 @@ import {
 import { groupAndSum, isYes } from './helpers.js'
 import { WASTE_RECORD_TYPE } from '#domain/waste-records/model.js'
 import { isDateInRange } from './filter-records-by-date.js'
+import { isOrsApprovedAtDate } from '#overseas-sites/domain/approval.js'
 
 const ORS_ID_DIGITS = 3
 const ZERO = '0'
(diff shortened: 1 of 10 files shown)


src/reports/repository/schema.js
```diff
@@ -74,7 +74,8 @@ const overseasSiteSchema = Joi.object({
   orsId: Joi.string().required(),
   siteName: Joi.string().allow(null).required(),
   country: Joi.string().allow(null).required(),
-  tonnageExported: Joi.number().min(0).required()
+  tonnageExported: Joi.number().min(0).required(),
+  approved: Joi.boolean().required()
```
GraemeF, line 78: Hooray! We can tell we broke the existing data!

src/reports/domain/aggregation/aggregate-waste-exported.js
```diff
@@ -52,7 +70,7 @@ const generateOverseasSiteSummaries = (wasteExportedRecords, orsDetailsMap) => {
 
   const unapprovedOverseasSites = summariseTonnage(
     groupAndSum(
-      recordsWithOrsId.filter((record) => !hasApprovedSite(record)),
+      recordsWithOrsId.filter((record) => !isResolvedSite(record)),
       ({ data }) => zeroPadOrsId(data.OSR_ID),
```
crosslandwa, line 74: Implementation detail that I don't expect to undertand....

...but just checking the its intentional that the `getKey` implementation here (`unapprovedOverseasSites`) returns `zeroPadOrsId(data.OSR_ID)`

whereas the `getKey` implementation  at line 55 (`overseasSites`) returns ``${zeroPadOrsId(data.OSR_ID)}:${approved}``

crosslandwa (approved):

src/reports/domain/aggregation/aggregate-waste-exported.js
GraemeF: That's because the same ORS can have 2 rows in the first table - one for approved loads, one for unapproved - but they're all unapproved in the second table.

PR #1051, opened 2026-04-12
PAE-1321: set tonnageReceivedNotExported to null for registered-only exporters

senthilinfy:
Ticket: [PAE-1321](https://eaflood.atlassian.net/browse/PAE-1321)
- Set `tonnageReceivedNotExported` to `null` for registered-only exporters as the value cannot be calculated without cross-tab correlation in the summary log
- Allow `null` in Joi schema for `exportActivitySchema.tonnageReceivedNotExported`
- Update port type to `number|null`
- Update tests to reflect `null` for registered-only exporter category

[PAE-1321]: https://eaflood.atlassian.net/browse/PAE-1321?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/reports/domain/aggregation/aggregate-waste-exported.js
+++ b/src/reports/domain/aggregation/aggregate-waste-exported.js
@@ -183,11 +183,14 @@ export function aggregateWasteExported({
     overseasSites,
     unapprovedOverseasSites,
     totalTonnageExported,
-    tonnageReceivedNotExported: calculateTonnageReceivedNotExported(
-      wasteReceivedRecords,
-      startDate,
-      endDate
-    ),
+    tonnageReceivedNotExported:
+      operatorCategory === OPERATOR_CATEGORY.EXPORTER_REGISTERED_ONLY
+        ? 0
+        : calculateTonnageReceivedNotExported(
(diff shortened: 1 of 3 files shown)


src/reports/domain/aggregation/aggregate-waste-exported.js
```diff
@@ -183,11 +183,14 @@ export function aggregateWasteExported({
...
     unapprovedOverseasSites,
     totalTonnageExported,
-    tonnageReceivedNotExported: calculateTonnageReceivedNotExported(
-      wasteReceivedRecords,
-      startDate,
-      endDate
-    ),
+    tonnageReceivedNotExported:
+      operatorCategory === OPERATOR_CATEGORY.EXPORTER_REGISTERED_ONLY
+        ? null
```
crosslandwa: As discussed, we'll pivot to returning `0` here (with a comment in code explaing) so that there is no trickle impact on the types

crosslandwa (approved):

PR #1053, opened 2026-04-13
PAE-1328: User-entered tonnageNotExported for registered-only exporters

senthilinfy:
Ticket: [PAE-1328](https://eaflood.atlassian.net/browse/PAE-1328)
## Summary

- Return `null` from aggregation for `EXPORTER_REGISTERED_ONLY` instead of calculating `tonnageReceivedNotExported` from waste records
- Accept `tonnageNotExported` in PATCH payload and map to `exportActivity.tonnageReceivedNotExported`
- Add `exportActivity` to MongoDB `performUpdateReport` so the user-entered value is persisted
- Relax `exportActivitySchema` to allow `null` on `tonnageReceivedNotExported`

[PAE-1328]: https://eaflood.atlassian.net/browse/PAE-1328?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/reports/routes/patch.js
+++ b/src/reports/routes/patch.js
@@ -5,6 +5,7 @@ import { StatusCodes } from 'http-status-codes'
 import { REPORT_STATUS } from '#reports/domain/report-status.js'
 import { fetchCurrentReport } from '#reports/application/report-service.js'
 import { maxTwoDecimalPlaces } from '#reports/repository/schema.js'
+import { WASTE_PROCESSING_TYPE } from '#domain/organisations/model.js'
 import {
   periodParamsSchema,
   standardUserAuth,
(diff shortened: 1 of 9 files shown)


src/reports/routes/patch.js
```diff
@@ -109,6 +111,10 @@ function buildUpdateFields(payload, report) {
     }
   }
 
+  if (tonnageNotExported !== undefined) {
```
crosslandwa, line 127: Should only allow this if reg-only exporter (else 400?)

crosslandwa (approved):

PR #1058, opened 2026-04-13
PAE-1194: Add dev organisation delete cascade

dansweeting:
Ticket: [PAE-1194](https://eaflood.atlassian.net/browse/PAE-1194)

## Summary

Adds `DELETE /v1/dev/organisations/{id}` to `epr-backend`. The endpoint is gated by the existing `FEATURE_FLAG_DEV_ENDPOINTS` feature flag via `plugins/router.js`, so it is only registered in environments where dev endpoints are enabled (test / perf-test) and returns `404` everywhere else.

Calling the endpoint cascades a delete across every collection that holds data linked to an organisation, downstream to root:

1. `packaging-recycling-notes`
2. `waste-balances` (joined via the org's accreditation IDs)
3. `reports`
4. `waste-records`
5. `summary-logs`
6. `overseas-sites` (joined via the org's registration-level `overseasSites` map)
7. `epr-organisations`

The response is `200 OK` with per-collection deleted counts:

```json
{
  "orgId": "...",
  "deletedCounts": {
    "packaging-recycling-notes": 3,
    "waste-balances": 2,
    "reports": 4,
    "waste-records": 17,
    "summary-logs": 1,
    "overseas-sites": 0,
    "epr-organisations": 1
  }
}
```

The endpoint is idempotent: calling it for an org that does not exist (or has already been cleaned up) returns `200` with all-zero counts.

## Why

The test and perf-test environments currently accumulate data from every journey-test run with no cleanup path, so volumes have drifted far from a prod-like shape. This PR is the first of three tasks under PAE-1194; the follow-ups will wire the endpoint into `epr-backend-journey-tests` and `epr-frontend-journey-tests` CI workflows so each run owns cleanup for the data it created, and will revert the temporary `setDefaultTimeout(30000)` workaround in the backend journey tests.

This PR ships the endpoint dormant. Nothing calls it until the follow-up PRs land.

## Implementation

The cascade lives in a dedicated dev-only module at `src/non-prod-data-reset/`. Production repository ports, adapters, and contract tests are genuinely untouched: no `deleteBy*` methods added to `organisationsRepository`, `packagingRecyclingNotesRepository`, `wasteBalancesRepository`, `reportsRepository`, `wasteRecordsRepository`, `summaryLogsRepository`, or `overseasSitesRepository`.

- `mongodb.js` owns the cascade semantics. A declarative list of collection plus filter steps (`buildCascadeSteps`) is walked by a generic `runCascade`. Collection names and filter paths live in one place.
- `mongodb.plugin.js` is the production plugin. It depends on `mongodb` and decorates `request.nonProdDataReset` with the adapter.
- `inmemory.plugin.js` is the test stub. It registers a noop that returns empty counts; tests that exercise the dev handler pass a custom stub via `createTestServer({ repositories: { nonProdDataReset } })`.
- `mongodb.test.js` is a single mongo integration test covering the cascade end to end: full cascade, isolation from unrelated orgs, non-existent org, invalid object id, idempotency, empty-accreditation and empty-overseas-site short circuits, and malformed org documents.

Both plugins are gated by `FEATURE_FLAG_DEV_ENDPOINTS`. When the flag is off the plugin is not registered at all, so `request.nonProdDataReset` does not exist on the request object in production. That is runtime defence in depth beyond the router-level route gate.

## Handler

`src/routes/v1/dev/organisations/delete-by-id.js` is a thin shell that runs Joi param validation and delegates to `request.nonProdDataReset.deleteByOrgId(id)`. All schema knowledge (accreditation IDs, overseas site IDs, filter paths) lives in the cascade module. The handler itself has no cascade logic, no repository fan-out, and no knowledge of the organisation document shape.

Jira: PAE-1194

[PAE-1194]: https://eaflood.atlassian.net/browse/PAE-1194?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/non-prod-data-reset/mongodb.js
```diff
@@ -0,0 +1,145 @@
...
+    })
+  )
+  return Object.fromEntries(entries)
+}
+
+/**
+ * Creates the non-prod data reset adapter.
+ *
+ * This module is intentionally the only place in the codebase that performs a
+ * cascade delete across organisation-scoped collections. It is registered via a
```
senthilinfy, line 133: Its probably better(defensive) to be able to enable these endpoins only in non prod env. sth like below. And log an error if its enabled in production?

```isDevEndpointsEnabled() {
    return !isProduction && config.get('featureFlags.devEndpoints')
  }
```

crosslandwa (approved):

src/non-prod-data-reset/mongodb.js
```diff
@@ -0,0 +1,145 @@
...
+ * Executes cascade steps in parallel. deleteMany is idempotent and each
+ * step's filter is independent of the others, so Promise.all is safe and
+ * meaningfully faster when journey tests batch cleanup across many orgs.
+ *
+ * @param {Db} db
+ * @param {ReturnType<typeof buildCascadeSteps>} steps
+ */
+const runCascade = async (db, steps) => {
+  const entries = await Promise.all(
+    steps.map(async ({ label, collection, filter }) => {
```
senthilinfy, line 118: Just wanted to mention ,deleting in parallel leaves data in inconsistent state. Not an issue here.

src/non-prod-data-reset/mongodb.js
dansweeting: 5cc985f69c83ca4f710a20193bab8d01d47f24bc

src/non-prod-data-reset/mongodb.js
dansweeting: Yes, for prod-facing concerns this would not be suitable

PR #1064, opened 2026-04-14
PAE-1340: Add context.traceId to SQS command message envelope

tomhalley:
## What

Adds the originating HTTP request's trace ID (`x-cdp-request-id`) to SQS command messages via an optional `context` envelope field, separating observability data from domain payload.

## Why

When the queue consumer processes validate/submit commands, it runs outside the HTTP request's `AsyncLocalStorage` context, so `getTraceId()` returns `null`. This means the entire validation and submission flow produces logs that cannot be correlated back to the originating request. This change captures the trace ID at send time so a follow-up change in the consumer can restore it.

## How

- `sendCommandMessage()` in `sqs-command-executor.js` now calls `getTraceId()` from `@defra/hapi-tracing` and includes it in an optional `context` object on the message body
- When no trace ID is available (e.g. local dev without the header, or non-HTTP producers), the `context` field is omitted — backward compatible with existing consumers

### Message format

**Before:**
```json
{"command": "validate", "summaryLogId": "abc-123"}
```

**After (with trace):**
```json
{"command": "validate", "summaryLogId": "abc-123", "context": {"traceId": "d4f8e2a1-..."}}
```

**After (without trace):**
```json
{"command": "validate", "summaryLogId": "abc-123"}
```

## Testing

- Integration tests added for both scenarios (trace present / absent)
- 100% coverage maintained on touched file
- All existing tests pass (18/18 in SQS executor suite)

## Follow-up

Consumer-side trace restoration will be done in a separate PR.

### DIFF
--- a/src/adapters/sqs-command-executor/sqs-command-executor.integration.test.js
+++ b/src/adapters/sqs-command-executor/sqs-command-executor.integration.test.js
@@ -1,7 +1,13 @@
 import { describe, expect, vi, beforeEach } from 'vitest'
 import { GetQueueUrlCommand, ReceiveMessageCommand } from '@aws-sdk/client-sqs'
+import { getTraceId } from '@defra/hapi-tracing'
 import { it } from '#vite/fixtures/sqs.js'
 import { createSqsCommandExecutor } from './sqs-command-executor.js'
+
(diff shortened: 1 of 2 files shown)


tomhalley:
Recreating with renamed branch (no slash).

PR #1070, opened 2026-04-15
PAE-1336: validate report completeness before status transition

loveboat:
Ticket: [PAE-1336](https://eaflood.atlassian.net/browse/PAE-1336)

- add completeness check on the report status-update endpoint so incomplete reports cannot transition to READY_TO_SUBMIT or SUBMITTED
- per-category required fields enforced via Joi schemas derived from the existing section schemas
- on incomplete, returns Boom.badRequest so the FE's existing generic error path handles it unchanged

[PAE-1336]: https://eaflood.atlassian.net/browse/PAE-1336?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/reports/application/is-report-complete.test.js
```diff
@@ -0,0 +1,250 @@
+import { OPERATOR_CATEGORY } from '#reports/domain/operator-category.js'
+import { describe, expect, it } from 'vitest'
```
timrutter, line 2: apparently these don't need to be imported

src/reports/routes/status.js
```diff
@@ -42,20 +44,26 @@ export const reportsStatus = {
...
+        reportsRepository,
+        organisationId,
+        registrationId,
+        year,
+        cadence,
+        period
+      )
+    ])
 
     if (!report) {
```
timrutter, line 68: registration can also be null here

crosslandwa: Interesting - the `registrationId` is used to load both the registration and the report

Assuming data integrity, a `report` being returned means the `registrationId` is valid, and thus there's an implicit assumption that a `registration` would be returned too...

timrutter: theoretically, you could have a report without a registration. Though i imagine in practice it won't happen

src/reports/application/is-report-complete.test.js
loveboat: I rather like being explicit about this sort of thing - `globals: true,` in `vitest.config.js` is it? Well, we'll see about that sort of thing! 😆

src/reports/routes/status.js
senthilinfy: organisationId, registrationId is mandatory to create report

loveboat: registration can't be null here [according to the mongo impl](https://github.com/DEFRA/epr-backend/blob/PAE-1336-validate-report-completeness/src/repositories/organisations/mongodb.js#L332-L334)

senthilinfy (approved):

timrutter (approved):

PR #1073, opened 2026-04-16
PAE-1191: Fix approved person extraction and user collation

sommerville-ee-defra:
Ticket: [PAE-1191](https://eaflood.atlassian.net/browse/PAE-1191)
## Summary
Fixes two independent bugs that caused approved person records to be wrong and at least one real approved person to be rejected at sign-in with "email not recognised". The registration extractor was reading the wrong form fields for submitter and approved person details, and a transition guard in the user collation logic prevented admin edits from propagating to the users list used by sign-in.

## What Changed
- Split submitter, approved person, and application contact form-field constants so each section of the submitted form is read independently. Previously all three collapsed onto the application contact fields.
- Dropped the transition guard in user collation so edits to approvedPersons on an already approved registration propagate to the users list. Side effect worth noting: the users list is now a pure projection of current state, so if a registration reverts from approved its approvedPersons are dropped on the next write (they used to persist).
- Extracted user collation into its own module with a simplified single-argument signature, plus a shared status module to break the resulting circular import.
- Added unit tests for the registration and accreditation paths and edge cases around missing fields, plus an integration test exercising edit-while-approved end to end.

## Why
Regulators reported that the approved person shown in the service did not match their records for some organisations, and that a real approved person was being bounced at sign-in. The extractor bug meant the service was recording whoever was entered as the application contact, not the approved person. The collation bug meant that even when an admin corrected the approved person via the Admin UI, the change never reached the list used to authorise sign-in.

[PAE-1191]: https://eaflood.atlassian.net/browse/PAE-1191?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/forms-submission-data/registration/form-field-constants.js
```diff
@@ -27,10 +27,15 @@ const ORG_NAME_FIELD = 'Org name'
...
 
-const APP_CONTACT_NAME = 'App contact name'
-const APP_CONTACT_EMAIL = 'App contact email address'
-const APP_CONTACT_TELEPHONE = 'App contact telephone number'
-const APP_CONTACT_JOB_TITLE = 'App contact job title'
+const AP_NAME = 'AP name'
+const AP_EMAIL = 'AP email address'
+const AP_TELEPHONE = 'AP telephone number'
+const AP_JOB_TITLE = 'AP role'
+
```
senthilinfy, line 39: There are three user details for registration(approved person, submitter, application contact).

Apart from above two we have to add application contact details as well

```

 {
      "shortDescription": "App contact email address",
      "title": "What is their email address?",
      "type": "EmailAddressField",
      "value": "reexserviceteam@defra.gov.uk"
    },
    {
      "shortDescription": "App contact telephone number",
      "title": "What is their telephone number?",
      "type": "TelephoneNumberField",
      "value": "020 7946 0123"
    },
    {
      "shortDescription": "App contact job title",
      "title": "What is their job title?",
      "type": "TextField",
      "value": "Director"
    },

{
      "shortDescription": "AP email address",
      "title": "What is their email address?",
      "type": "EmailAddressField",
      "value": "james.patterson@ecoretail.co.uk"
    },
    {
      "shortDescription": "AP telephone number",
      "title": "What is their telephone number?",
      "type": "TelephoneNumberField",
      "value": "020 7946 0123"
    },
    {
      "shortDescription": "AP role",
      "title": "What is their role within the organisation?",
      "type": "RadiosField",
      "value": "Chair, treasurer or secretary"
    },
    {
      "shortDescription": "Submitter name",
      "title": "What is your full name?",
      "type": "TextField",
      "value": "James Patterson"
    },
    {
      "shortDescription": "Submitter email address",
      "title": "What is your email address?",
      "type": "EmailAddressField",
      "value": "james.patterson@ecoretail.co.uk"
    },
    {
      "shortDescription": "Submitter telephone number",
      "title": "What is your telephone number?",
      "type": "TelephoneNumberField",
      "value": "020 7946 0123"
    },
    {
      "shortDescription": "Submitter job title",
      "title": "What is your job title?",
      "type": "TextField",
      "value": "Directory"
    }
```

src/repositories/organisations/collate-users.js
```diff
@@ -0,0 +1,143 @@
...
+ * @returns {SlimUser[]}
+ */
+const collateItems = (updated, collectionKey, extractAdditionalUsers) => {
+  /** @type {SlimUser[]} */
+  const users = []
+
+  for (const item of updated[collectionKey]) {
+    const itemStatus = getCurrentStatus(item)
+
+    if (itemStatus === REG_ACC_STATUS.APPROVED) {
```
senthilinfy, line 36: I don't have the background. But once a user is approved then suspended/cancelled, they should still be able to login right?

src/forms-submission-data/registration/form-field-constants.js
sommerville-ee-defra: Claude dropped these because nothing referenced them after the rewire, but it makes the code less discoverable so I'll add them back in.

src/repositories/organisations/collate-users.js
sommerville-ee-defra: I'm not sure I really have the background either, but this is Claude's take:

Good question - short answer is this PR doesn't change that behaviour, but it's worth explaining. collateItems has always only contributed users for currently-APPROVED registrations (pre-fix it was stricter still: only on the transition into approved). What keeps a suspended/cancelled reg's user signing in is the ...(updated.users ?? []) spread in collateUsers - in the Admin UI fetch-then-PUT pattern, users added during the APPROVED period persist through subsequent status transitions.

There's a related pre-existing inconsistency: getUserRolesForStatus returns [STANDARD] for non-APPROVED statuses (no INITIAL / sign-in role), but collateItems gates on APPROVED-only so that branch never runs via reg/acc. Looks like dead code or an unimplemented intent. Worth a separate ticket to clarify product intent for suspend/cancel - out of scope for PAE-1191.

Does that help?

senthilinfy: I discussed this with Will and it makes sense for users array to reflect latest state. 

Once users link an organisation after that authentication is not driven by users array but which organisation user is part of in defra id. so above is fine

senthilinfy (approved):

PR #1076, opened 2026-04-16
PAE-1230: Immediate redrive on transient errors and fix log levels

dansweeting:
Ticket: [PAE-1230](https://eaflood.atlassian.net/browse/PAE-1230)
## Summary

When an SQS message fails processing (transient error, invalid JSON, unknown command), the consumer now resets the message's visibility timeout to 1s so it is immediately available for retry, rather than waiting for the full queue visibility timeout (30s default).

This uses sqs-consumer's built-in `terminateVisibilityTimeout` option on `Consumer.create`, which handles the `ChangeMessageVisibility` call for all handler errors in one place. Earlier iterations did this manually inside `handleCommandError`, which missed unparseable messages (they throw before reaching that code path). The library option covers both paths.

Other changes:
- Use `warn` log level for non-terminal transient failures, reserving `error` for permanent failures and final transient attempts
- Demote `processing_error` event handler to `warn` (most errors are already logged with full context in `handleCommandError`; this handler catches the remainder)

### Background

The original PR (#951) was reverted (#971) because the ECS task role in CDP lacked `sqs:ChangeMessageVisibility`; that has since been granted. This PR also replaces the manual `resetVisibilityTimeout` helper with the library's built-in support, removing the `ChangeMessageVisibilityCommand` import from our code entirely.

## Test plan

- [x] Unit tests: warn log levels, final attempt error logging, `terminateVisibilityTimeout: 1` passed to `Consumer.create`
- [x] Integration test: creates a queue with 30s visibility timeout, verifies retry completes in under 10s (proving the reset works)
- [x] Manual test: sent malformed messages (invalid JSON, unknown command) to LocalStack queue, confirmed ~2s retries and DLQ delivery after 3 attempts
- [x] All existing tests passing, 100% coverage

[PAE-1230]: https://eaflood.atlassian.net/browse/PAE-1230?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/common/helpers/sqs/sqs-client.js
+++ b/src/common/helpers/sqs/sqs-client.js
@@ -1,7 +1,8 @@
 import {
   SQSClient,
   GetQueueUrlCommand,
-  GetQueueAttributesCommand
+  GetQueueAttributesCommand,
+  ChangeMessageVisibilityCommand
 } from '@aws-sdk/client-sqs'
 
 /** @typedef {import('@aws-sdk/client-sqs').SQSClient} SQSClientType */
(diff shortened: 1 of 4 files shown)


src/common/helpers/sqs/sqs-client.js
```diff
@@ -56,3 +57,26 @@ export async function getMaxReceiveCount(sqsClient, queueUrl) {
...
+export async function resetVisibilityTimeout(
+  sqsClient,
+  queueUrl,
+  receiptHandle
+) {
+  await sqsClient.send(
+    new ChangeMessageVisibilityCommand({
+      QueueUrl: queueUrl,
+      ReceiptHandle: receiptHandle,
+      VisibilityTimeout: 0
```
GraemeF: Would a little bit of time help whatever is having the transient problem recover? Or is that not what this is?

dansweeting: Fair, that is what this is. I toyed with the exponential backoff sort of thing. Wait some time, wait a bit longer on the next retry maybe. Felt as though it deserved its own discussion. 

But I'm happy to put in a nominal delay. Then we've had the discussion !

dansweeting: Changed to 1s in 794ce7de, which is the minimum non-zero value SQS accepts (whole seconds only).

GraemeF (approved):

PR #1080, opened 2026-04-17
PAE-1230: Acknowledge unparseable SQS messages instead of retrying

dansweeting:
Ticket: [PAE-1230](https://eaflood.atlassian.net/browse/PAE-1230)
## Summary

- When a message fails JSON parsing, envelope validation, or payload schema validation, the consumer now acknowledges (deletes) the message immediately instead of throwing, which would leave it for SQS to retry after the visibility timeout
- Schema-invalid messages fail identically on every attempt, so retrying them wastes processing time and eventually sends them to the DLQ for no reason
- The error is still logged at the point of failure inside `parseCommandMessage`, so observability is unchanged

## Changes

- `consumer.js`: Changed `throw new Error(...)` to `return message` when `parseCommandMessage` returns null (line ~194). Returning the message signals sqs-consumer to acknowledge/delete it.
- `consumer.test.js`: Updated 6 unit tests from `rejects.toThrow` to `expect(result).toBe(message)` for all parsing failure paths.
- `consumer.integration.test.js`: Converted 2 existing "sends to DLQ" tests to assert the message is acknowledged (gone from both main queue and DLQ). Added a new integration test for unparseable JSON.

## Test plan

- [x] Unit tests pass (47 tests)
- [x] Integration tests pass against local ElasticMQ (11 tests)
- [ ] Full test suite passes in CI

[PAE-1230]: https://eaflood.atlassian.net/browse/PAE-1230?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


GraemeF:
The original idea (before we built any of this) was that whatever is in the DLQ probably needs investigation, so I would have expected unparseable payloads from our own services to be in that category. Is the DLQ just too noisy compared to error logs?

dansweeting:
> The original idea (before we built any of this) was that whatever is in the DLQ probably needs investigation, so I would have expected unparseable payloads from our own services to be in that category. Is the DLQ just too noisy compared to error logs?

The DLQ isnt particularly noisy. It does make sense to be able to inspect anything we couldn't process in a DLQ.

I think I might have ended up at this solution because for some reason unparseable payloads were not being retried? But then I questioned that: what is the point of retrying an invalid payload - just fail fast.

Maybe the fix really is to stand back and look at the error processing here to see why that was the case in the first place. If unprocessable items show up in DLQ, maybe that should count as a `PermanentError` and not retry at all.

PR #1083, opened 2026-04-20
PAE-1367 Endpoint to get all submitted/pending report submissions

senthilinfy:
Ticket: [PAE-1367](https://eaflood.atlassian.net/browse/PAE-1367)
## PAE-1367 Endpoint to get all submitted/pending report submissions

- This endpoint will be used to generate csv download of all submitted/pending reports

[PAE-1367]: https://eaflood.atlassian.net/browse/PAE-1367?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/reports/routes/submissions.js
```diff
@@ -0,0 +1,51 @@
...
+  path: getReportSubmissionsPath,
+  options: {
+    auth: getAuthConfig([ROLES.serviceMaintainer]),
+    tags: ['api', 'admin'],
+    response: {
+      schema: Joi.object({
+        generatedAt: Joi.string().isoDate().required(),
+        reportSubmissions: Joi.array()
+          .items(
+            Joi.object({
```
sommerville-ee-defra, line 20: From CC: "Joi requires non-empty strings for submitter/approver phone/email. Any org with empty approvedPersons or missing contact fields will 500 the entire endpoint via failAction: 'error'. Add .allow('') like the other fields."

How confident are we that `approvedPersons` is always populated?

src/reports/application/report-submissions.js
```diff
@@ -0,0 +1,170 @@
...
+) {
+  const currentYear = new Date().getUTCFullYear()
+  /** @type {ReportSubmissionsRow[]} */
+  const rows = []
+
+  const allPeriodicReports = await reportsRepository.findAllPeriodicReports()
+  /** @type {Map<string, import('#reports/repository/port.js').PeriodicReport[]>} */
+  const reportsByKey = new Map(
+    allPeriodicReports.map((pr) => [
+      `${pr.organisationId}::${pr.registrationId}`,
```
sommerville-ee-defra: From CC: "silently collapses multi-year docs for the same org/reg. Quarterly registered-only operators will lose historical-year data the minute we roll into next year. Fix by aggregating into arrays."

I think the issue here is the each `periodicReport` includes a `year`, but this isn't being used in the key which might result in data loss?

src/reports/routes/submissions.js
senthilinfy: It should always be present based on registration schema. 

```
  approvedPersons: Joi.array().items(userSchema).required().min(1),

const baseUserSchema = Joi.object({
  fullName: Joi.string().required(),
  email: Joi.string().email().required()
})

export const userSchema = baseUserSchema.keys({
  phone: Joi.string().required(),
  jobTitle: Joi.string().required()
})
```

src/reports/application/report-submissions.js
senthilinfy: This should work across years as findAllPeriodicReports returns an array with one entry for each year( PeriodicReport is at year level). Added a test(last commit) to verify the behavior as well


/**
 * @typedef {Object} PeriodicReport
 * @property {string} organisationId
 * @property {string} registrationId
 * @property {number} year
 * @property {{ monthly?: PeriodicReportSlots, quarterly?: PeriodicReportSlots }} reports
 */

senthilinfy: let me take a look at this. its valid, inside report-submission, overlooked earlier

sommerville-ee-defra: Yeah, was just going to re-raise - I think the test needs to call `generateReportSubmissions`, which would surface the issue?

sommerville-ee-defra (approved):

PR #1088, opened 2026-04-21
PAE-1288: Add DLQ admin endpoints

dansweeting:
Ticket: [PAE-1288](https://eaflood.atlassian.net/browse/PAE-1288)

## What

Adds two admin-only API endpoints for Dead Letter Queue management:

- `GET /v1/admin/queues/dlq/messages` — returns `{ approximateMessageCount, messages }` where each message includes parsed command fields and the raw body
- `POST /v1/admin/queues/dlq/purge` — purges all messages from the DLQ and returns `{ purged: true }`

Both routes require the `serviceMaintainer` role.

## Why

Operations teams need to inspect DLQ message contents (summary log IDs, command types) when troubleshooting live issues, without requiring breakglass prod access. The queue management page in the admin frontend (PAE-1288) consumes these endpoints.

## Implementation

Generic SQS primitives in `sqs-client.js`:

- `resolveDlqUrl` — discovers the DLQ URL from a queue's redrive policy
- `getApproximateMessageCount` — reads the approximate message count from queue attributes
- `receiveMessages` — loops `ReceiveMessage` calls with a short visibility timeout, deduplicates by MessageId, and returns up to a configurable cap (default 100)
- `purgeQueue` — purges all messages from a queue

A `dlq-admin` plugin creates an SQS client at startup, resolves the DLQ URL, and decorates `request.dlqService` with `{ getMessages, purge }`. The messages endpoint parses each message body as JSON to extract command type and payload; if parsing fails, the `command` field is null and the raw body is still returned.

[PAE-1288]: https://eaflood.atlassian.net/browse/PAE-1288?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/common/helpers/sqs/sqs-client.js
```diff
@@ -56,3 +66,117 @@ export async function getMaxReceiveCount(sqsClient, queueUrl) {
...
+ * @param {SQSClientType} sqsClient
+ * @param {string} queueUrl
+ * @returns {Promise<void>}
+ */
+export async function purgeQueue(sqsClient, queueUrl) {
+  await sqsClient.send(new PurgeQueueCommand({ QueueUrl: queueUrl }))
+}
+
+/**
+ * Receives messages from a queue, deduplicating by MessageId.
```
dansweeting: Should be made clear here that this is a 'peek' style retrieval with the intent to not lose the messages.

dansweeting: It's also unclear in the jsdoc as to why we even need to deduplicate.

src/common/helpers/sqs/sqs-client.integration.test.js
```diff
@@ -0,0 +1,241 @@
...
+        await sqsClient.send(
+          new SendMessageCommand({ QueueUrl: dlqUrl, MessageBody: 'test' })
+        )
+
+        const count = await getApproximateMessageCount(sqsClient, dlqUrl)
+
+        expect(count).toBeGreaterThanOrEqual(1)
+      }
+    )
+  })
```
dansweeting, line 85: These functions existed beforehand and had unit tests, but given there were so many new excuses for integration tests in the newer queue functions, tests for the old functions have moved here for consistency's sake.

src/common/helpers/sqs/sqs-client.js
```diff
@@ -56,3 +66,136 @@ export async function getMaxReceiveCount(sqsClient, queueUrl) {
...
+    messageId: msg.MessageId,
+    sentTimestamp: Number.isFinite(timestamp)
+      ? new Date(timestamp).toISOString()
+      : null,
+    approximateReceiveCount:
+      Number(msg.Attributes?.ApproximateReceiveCount) || 0,
+    body: msg.Body
+  }
+}
+
```
dansweeting: Move this to the bottom of the file, you've squatted it right before the comments for the exported function

src/common/helpers/sqs/sqs-client.js
```diff
@@ -56,3 +66,136 @@ export async function getMaxReceiveCount(sqsClient, queueUrl) {
...
+      })
+    )
+
+    const unique = (result.Messages ?? []).filter((msg) => {
+      if (!msg.MessageId || msg.Body === undefined || seen.has(msg.MessageId)) {
+        return false
+      }
+      seen.add(msg.MessageId)
+      return true
+    })
```
dansweeting: Is filter a sensible idea? You've ended up having to make the filter callback cause a side effect in the `seen` set

src/common/helpers/sqs/sqs-client.js
dansweeting: The JSDoc already opens with "Peeks at messages in a queue without consuming them" and explains the short VisibilityTimeout mechanism. No change needed beyond what was already there.

dansweeting: The JSDoc explains this in the second paragraph: "Because SQS distributes messages across servers, repeated polling may return the same messages once the visibility timeout expires. Results are deduplicated by MessageId and the loop exits when a batch yields nothing new."

src/common/helpers/sqs/sqs-client.integration.test.js
dansweeting: Acknowledged.

src/common/helpers/sqs/sqs-client.js
dansweeting: Moved to the bottom of the file in 74efbeb5.

src/common/helpers/sqs/sqs-client.js
dansweeting: Replaced with a plain for loop in 74efbeb5.

GraemeF (approved):

PR #1090, opened 2026-04-21
PAE-1379: Fix cadence and category for registrations with unapproved accreditation

sommerville-ee-defra:
Ticket: [PAE-1379](https://eaflood.atlassian.net/browse/PAE-1379)
## Summary
Registrations linked to an accreditation that has not been approved were being treated as fully accredited - getting monthly report cadence and the accredited operator category. This change makes the determination consider the accreditation's status, not just the presence of a linked ID.

## What Changed
- Added a domain helper that treats a registration as accredited only when its linked accreditation is currently `approved` or `suspended`
- Wired the helper into the reports calendar route (cadence) and the operator category derivation (report sections, completeness rules)
- Extended unit and integration tests to cover the full status matrix, including the unapproved cases (`created`, `rejected`, `cancelled`) and the edge case where an accreditation ID points at nothing

## Why
An approved registration can be linked to an accreditation that is still in `created`, `rejected`, or `cancelled` state. Previously the backend only checked whether `accreditationId` was present, so such operators were asked for monthly submissions instead of quarterly and routed down the wrong validation path. The fix is backend-only and preserves the existing API contract, so no coordinated deployment with the frontend is required.

[PAE-1379]: https://eaflood.atlassian.net/browse/PAE-1379?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/reports/domain/is-registration-accredited.js
+++ b/src/reports/domain/is-registration-accredited.js
@@ -1,0 +1,18 @@
+import { REG_ACC_STATUS } from '#domain/organisations/model.js'
+
+/**
+ * Returns true when the registration is linked to an accreditation that is
+ * live (approved or suspended). Presence of accreditationId alone is not
+ * sufficient — an accreditation in 'created', 'rejected', or 'cancelled'
+ * state has never been active and must be treated as registered-only.
+ *
+ * @param {{ accreditation?: { status?: string } | null }} [registration]
+ * @returns {boolean}
+ */
+export function isRegistrationAccredited(registration) {
+  const status = registration?.accreditation?.status
+
+  return (
+    status === REG_ACC_STATUS.APPROVED || status === REG_ACC_STATUS.SUSPENDED
+  )
+}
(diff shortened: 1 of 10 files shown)


src/reports/domain/is-registration-accredited.js
```diff
@@ -0,0 +1,18 @@
...
+ * Returns true when the registration is linked to an accreditation that is
+ * live (approved or suspended). Presence of accreditationId alone is not
+ * sufficient — an accreditation in 'created', 'rejected', or 'cancelled'
+ * state has never been active and must be treated as registered-only.
+ *
+ * @param {{ accreditation?: { status?: string } | null }} [registration]
+ * @returns {boolean}
+ */
+export function isRegistrationAccredited(registration) {
+  const status = registration?.accreditation?.status
```
senthilinfy: is optional chaining needed on registration? that should never be null right

senthilinfy (approved):

PR #1091, opened 2026-04-21
PAE-1380: Derive waste balance from waste records, not rowId-keyed ledger

GraemeF:
Ticket: [PAE-1380](https://eaflood.atlassian.net/browse/PAE-1380)
## Summary

The waste-records collection is uniquely keyed by `(organisationId, registrationId, type, rowId)`, but waste-balance transactions key their entity ids by naked `String(rowId)`. When records of different types share a numeric rowId under the same accreditation (e.g. REPROCESSOR_INPUT with Received rows running past 4000 and colliding with Reprocessed rows 4000-4280), the later-processed record's target silently offsets the earlier record's credit, corrupting the closing balance.

The fix leaves the per-record transaction ledger untouched — transactions still get written exactly as before for audit purposes. What changes is how the balance totals on the waste-balance document are computed:

- `newAmount` and `newAvailableAmount` are now derived directly from the current waste records (via `sum(getTargetAmount(r))`) plus the net adjustment from PRN transactions in the historic ledger.
- PRN transactions use `prnId` as the entity id and do not suffer the collision, so their net effect is safely pulled from the existing ledger by matching on `prn:*` entity types.

Remediating the seven accreditations that already hold corrupt balances is covered by a separate ticket.

## Changes

- `src/domain/waste-balances/calculator.js` — add `sumPrnTransactionAdjustment` and use it together with a direct sum of waste-record target amounts to compute `newAmount` and `newAvailableAmount`. Per-record transaction generation (and its delta-on-`creditedAmountMap` bookkeeping) is unchanged.
- `src/domain/waste-balances/calculator.test.js` — adds scenarios covering the rowId collision, a re-upload after a PRN_ISSUED debit, a re-upload against a PRN_CREATED ringfence, and a malformed PRN transaction missing opening/closing fields.
- `src/repositories/waste-balances/contract/updateWasteBalanceTransactions.contract.js` — updates two assertions that previously relied on carry-forward of an anonymous opening balance. Under the new semantics, the balance totals are fully derived from the waste records plus PRN transactions, so stray prior amounts without a corresponding waste record or PRN transaction are not retained.
- `src/routes/v1/organisations/registrations/summary-logs/integration.waste-balance-arithmetic.test.js` — adds an end-to-end scenario that uploads records, issues a PRN (reducing total `amount`), re-uploads the same rows, and verifies the issued debit survives the re-upload.

[PAE-1380]: https://eaflood.atlassian.net/browse/PAE-1380?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/domain/waste-balances/calculator.js
+++ b/src/domain/waste-balances/calculator.js
@@ -110,23 +110,32 @@ const isPrnTransaction = (transaction) =>
   )
 
 /**
+ * @typedef {'openingAmount' | 'closingAmount' | 'openingAvailableAmount' | 'closingAvailableAmount'} WasteBalanceField
+ */
+
+/**
  * Sums the net effect of PRN transactions on a balance field. Waste-record
  * transactions are excluded because they are keyed by naked rowId and can
  * silently lose data under a rowId collision (PAE-1380); the waste-record
  * contribution is derived directly from the waste records themselves.
  *
  * @param {Array<import('#domain/waste-balances/model.js').WasteBalanceTransaction>} transactions
- * @param {'Amount' | 'AvailableAmount'} field
+ * @param {WasteBalanceField} closingField
+ * @param {WasteBalanceField} openingField
  * @returns {number}
  */
-const sumPrnTransactionAdjustment = (transactions, field) => {
+const sumPrnTransactionAdjustment = (
+  transactions,
+  closingField,
+  openingField
+) => {
   let adjustment = 0
   for (const transaction of transactions) {
     if (!isPrnTransaction(transaction)) {
       continue
     }
-    const closing = transaction[`closing${field}`] ?? 0
-    const opening = transaction[`opening${field}`] ?? 0
+    const closing = transaction[closingField] ?? 0
+    const opening = transaction[openingField] ?? 0
     adjustment = toNumber(add(adjustment, subtract(closing, opening)))
   }
   return adjustment
@@ -235,11 +244,13 @@ export const calculateWasteBalanceUpdates = ({
   // collision-free, so their net adjustment is safe to pull from the ledger.
   const prnAmountAdjustment = sumPrnTransactionAdjustment(
     historicTransactions,
-    'Amount'
+    'closingAmount',
+    'openingAmount'
   )
   const prnAvailableAmountAdjustment = sumPrnTransactionAdjustment(
     historicTransactions,
-    'AvailableAmount'
+    'closingAvailableAmount',
+    'openingAvailableAmount'
   )
 
   return {
--- a/src/domain/waste-balances/calculator.js
+++ b/src/domain/waste-balances/calculator.js
@@ -110,35 +110,39 @@ const isPrnTransaction = (transaction) =>
   )
 
 /**
- * @typedef {'openingAmount' | 'closingAmount' | 'openingAvailableAmount' | 'closingAvailableAmount'} WasteBalanceField
- */
-
-/**
- * Sums the net effect of PRN transactions on a balance field. Waste-record
- * transactions are excluded because they are keyed by naked rowId and can
- * silently lose data under a rowId collision (PAE-1380); the waste-record
- * contribution is derived directly from the waste records themselves.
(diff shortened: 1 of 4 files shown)


src/domain/waste-balances/calculator.js
```diff
@@ -93,6 +93,42 @@ const updateCreditedAmountMap = (creditedAmountMap, transaction) => {
...
+  for (const transaction of transactions) {
+    if (!isPrnTransaction(transaction)) {
+      continue
+    }
+    const closing = transaction[`closing${field}`] ?? 0
+    const opening = transaction[`opening${field}`] ?? 0
+    adjustment = toNumber(add(adjustment, subtract(closing, opening)))
+  }
+  return adjustment
+}
```
dansweeting, line 146: If you took the full property names instead you'd get `keyof` safety:

```js
/**
 * @param {WasteBalanceTransaction[]} transactions
 * @param {keyof WasteBalanceTransaction} closingField
 * @param {keyof WasteBalanceTransaction} openingField
 */
const sumPrnTransactionAdjustment = (transactions, closingField, openingField) => {
  ...
  const closing = transaction[closingField] ?? 0
  const opening = transaction[openingField] ?? 0
}
```

Called as:

```js
sumPrnTransactionAdjustment(historicTransactions, 'closingAmount', 'openingAmount')
sumPrnTransactionAdjustment(historicTransactions, 'closingAvailableAmount', 'openingAvailableAmount')
```

Slightly more verbose at the call sites but one less thing to go wrong silently.

src/domain/waste-balances/calculator.js
```diff
@@ -180,19 +219,34 @@ export const calculateWasteBalanceUpdates = ({
         type
       )
 
-      // Update State
       currentAmount = transaction.closingAmount
       currentAvailableAmount = transaction.closingAvailableAmount
```
dansweeting, line 236: Now that newAmount/newAvailableAmount are derived independently, currentAmount and currentAvailableAmount only survive to populate the openingAmount/closingAmount fields on the audit transactions. 

But they're still seeded from currentBalance.amount and carried forward through the old delta logic, so for an accreditation that already has a corrupt balance, wouldn't the audit entries inherit misleading opening/closing values? Do we care?

src/domain/waste-balances/calculator.js
dansweeting: Not a blocker, and probably a debate to be had since it doesn't rule out passing contradictory keys of the type. There's probably a more convoluted type workaround. Or just having done with it and coupling the two sets of balances in the same call - maintaining two adjustments in the same func.

GraemeF: Good call — and your follow-up nudged it further. Narrowing the params to a `'openingAmount' | 'closingAmount' | …` literal union still admits `('openingAmount', 'openingAmount')` or swapped opening/closing, so the miswiring risk doesn't really go away. Folded both adjustments into a single-pass `sumPrnTransactionAdjustments` that returns `{ amount, availableAmount }` — no field-name params at all, direct property access, `tsc` sees numeric types trivially. Pushed in c7df824.

src/domain/waste-balances/calculator.js
GraemeF: Real concern, but I think we can defer. The balance self-heals automatically: on the next submission for a corrupt accreditation, `newAmount` is derived from the waste records, so `balance.amount` becomes correct without any remediation step. The transaction ledger stays scarred — the first post-fix submission's transactions still chain opening/closing from the corrupt seed — but that's consistent with the PR's scope of leaving the ledger untouched, and the audit trail stays internally consistent (no mix of corrupt-seeded and derived-seeded transactions in the same ledger). The strategic fix is the per-event ledger rewrite in draft ADR 0031 (DEFRA/epr-re-ex-service#187), which will reconcile audit history properly.

dansweeting (approved):

PR #1117, opened 2026-04-24
PAE-1406: Skip file copy for unlinked form submissions during migration

sommerville-ee-defra:
Ticket: [PAE-1406](https://eaflood.atlassian.net/browse/PAE-1406)
## Summary
Adds a config-driven allow-list for excluding known-orphaned accreditation form submissions from form data migration. Targets the Backend errors alert firing on every pod restart when a submission references an organisation that does not exist in the backend, causing its operator-uploaded files to be retried and to fail indefinitely.

## What Changed
- New config key `skipMigratingAccreditations` (env: `MIGRATION_ACCREDITATIONS_SKIP`) accepting a JSON array of accreditation form submission mongo ids.
- Listed ids are filtered out of `pendingMigration` in the migration delta calculator, so they are never transformed, split, linked, or passed to the file copy step.
- Unit test covering the new filter behaviour.
- Mirrors the existing `TEST_ORGANISATIONS_SKIP_TRANSFORM` pattern for consistency.

## Why
Known-orphaned submissions previously stayed in `pendingMigration` on every pod start (because nothing was ever persisted for them), so the file copy retried and alerted on every deployment. A targeted skip list avoids the retry noise without changing the behaviour for legitimate unlinked submissions, where eager file copy is still desirable to protect against data loss if an operator orgId typo is later corrected. Once an orphaned id is added to the env var in CDP config, its alert stops immediately on next deploy; removing it later resumes normal migration behaviour. Setting the env var in prod to `["697a10349b947884b8407c7d"]` will silence the currently-firing alert.

[PAE-1406]: https://eaflood.atlassian.net/browse/PAE-1406?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


senthilinfy:
It will be better to  add config similar to TEST_ORGANISATIONS_SKIP_TRANSFORM to skip these accreditations in migration-delta-calculator.js

```
  const pendingAccrs = submissionIds.accreditations.difference(
    migratedIds.accreditations
  )

```

 ```2 accreditations not linked to an organisation: id=697a10349b947884b8407c7d,systemReference=68dfc5198687d2e196d972c2,orgId=500001; id=69eb80e8fafcb1bd51cd81e0,systemReference=68dfc5198687d2e196d972c2,orgId=500001```


The main reason is 

1. The registration/accreditation's this change is processing include old/already migrated one's as well. If they are older than 90 days copy will fail and alert
2. There can be valid operator submitted  registration/accreditation's that are not linked to an org due to typo. But files uploaded by them should still be copied

senthilinfy (approved):

PR #1119, opened 2026-04-24
PAE-1396: Add reconciliation system logs to /v1/me/organisations

dansweeting:
Ticket: [PAE-1396](https://eaflood.atlassian.net/browse/PAE-1396)

## What changed

When a logged-in user hits `GET /v1/me/organisations`, the handler now records a system log capturing the DefraID-to-EPR reconciliation state at the time of the call. This gives us visibility into identity mismatches and user-organisation relationships across the login flow.

**New file: `src/auditing/organisations-discovery.js`**

Exports `auditOrganisationsDiscovery`, which writes an event with:

- `event`: `identity / defra-id-reconciliation / organisations-discovered`
- `context.organisationId`: EPR org UUID, present only when a linked org exists
- `context.defraIdOrg`: current org from the token (`{ id, name }`)
- `context.defraIdRelationships`: all relationships from the token, each with `defraIdOrgId`, `defraIdOrgName`, and `isCurrent`
- `context.linked`: the linked EPR org with `status`, `linkedBy`, and `linkedAt` (ISO 8601), or `null`
- `context.unlinked`: linkable EPR orgs with `status`, or `[]`

Also exports `auditTokenValidationFailed`, which writes a `token-validation-failed` event when the DefraID token is malformed (missing current organisation or required fields). This captures the raw relationships and error message, giving support visibility into login failures caused by broken tokens.

Both functions call `safeAudit` (CDP pipeline) and `recordSystemLog` (MongoDB), matching the pattern used elsewhere in the auditing module.

**Handler change (`src/routes/v1/me/organisations/get.js`)**

`getOrgDataFromDefraIdToken` is now called at the handler level so the full relationships array is available for auditing. It was previously called inside `getCurrentDetailsFromToken`, which only returned the current org. The refactored `getCurrentDetailsFromToken` receives the parsed relationships array directly.

`getCurrentDetailsFromToken` is wrapped in a try/catch: if the token is malformed, a `token-validation-failed` system log is persisted before the 403 is re-thrown. The audit call itself has its own try/catch for resilience.

The happy-path audit call is also wrapped in a `try/catch`: if it throws, the error is logged but the endpoint still returns 200. This is intentional, as the login-path endpoint must not be broken by audit infrastructure failures.

The audit log uses separate `auditLinked` and `auditUnlinked` shapes (with `status`) rather than the HTTP response shapes, which omit status.

**Type changes**

- `SystemLogsRepository` added to the shared `HapiRequest` type in `hapi-types.js`, removing the need for union type overrides at call sites.
- `DefraIdArtifacts` type in `hapi-types.js`: `HapiRequest.auth.artifacts` is typed as `unknown` because different routes use different token payloads. The handler casts to `DefraIdArtifacts`, making the shape explicit and reusable.

**Tests**

Eight integration tests covering the Jira acceptance criteria: full linked context, user identity in `createdBy`, multiple DefraID relationships, no linked org with linkable orgs, no EPR orgs at all, audit failure resilience, malformed token audit logging, and malformed token with audit failure resilience.

[PAE-1396]: https://eaflood.atlassian.net/browse/PAE-1396?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/routes/v1/me/organisations/get.test.js
```diff
@@ -556,4 +560,293 @@ describe('GET /v1/me/organisations', () => {
...
+      expect(systemLogs).toHaveLength(1)
+      const [log] = systemLogs
+      expect(log.context.organisationId).toBeUndefined()
+      expect(log.context.linked).toBeNull()
+      expect(log.context.unlinked).toHaveLength(1)
+      expect(log.context.unlinked[0]).toMatchObject({
+        id: unlinkedOrg.id,
+        orgId: unlinkedOrg.orgId,
+        status: 'approved'
+      })
```
dansweeting: This could be a single `expect`

src/routes/v1/me/organisations/get.test.js
```diff
@@ -556,4 +560,293 @@ describe('GET /v1/me/organisations', () => {
...
+      expect(response.statusCode).toBe(StatusCodes.OK)
+
+      const { systemLogs } = await systemLogsRepository.find({
+        subCategory: 'defra-id-reconciliation',
+        limit: 10
+      })
+      expect(systemLogs).toHaveLength(1)
+      const [log] = systemLogs
+      expect(log.context.linked).toBeNull()
+      expect(log.context.unlinked).toEqual([])
```
dansweeting: single expect

src/routes/v1/me/organisations/get.test.js
dansweeting: Fixed in fbf34d54 — consolidated into a single `toMatchObject`. Also changed `organisationId` from `undefined` to `null` when no linked org exists (cleaner for serialisation).

src/routes/v1/me/organisations/get.test.js
dansweeting: Fixed in fbf34d54 — consolidated into a single `toMatchObject`.

src/routes/v1/me/organisations/get.js
```diff
@@ -104,7 +104,14 @@ export const organisationsLinkedGetAll = {
...
 
-    const current = getCurrentDetailsFromToken(auth, logger)
+    const {
+      decoded: { payload: tokenPayload }
+    } = /** @type {import('#common/hapi-types.js').DefraIdArtifacts} */ (
+      auth.artifacts
+    )
+    const defraIdRelationships = getOrgDataFromDefraIdToken(tokenPayload)
+
+    const current = getCurrentDetailsFromToken(defraIdRelationships, logger)
```
dansweeting: If `getCurrentDetailsFromToken` throws (malformed token with no current org), the request gets a 403 before reaching the audit call. This is intentional: there is no reconciliation state worth logging when the token itself is broken. The ACs cover genuine login flows, not invalid tokens.

GraemeF (approved):

src/routes/v1/me/organisations/get.js
```diff
@@ -92,7 +92,7 @@ export const organisationsLinkedGetAll = {
     tags: ['api']
   },
   /**
-   * @param {import('#common/hapi-types.js').HapiRequest} request
+   * @param {import('#common/hapi-types.js').HapiRequest & { systemLogsRepository: SystemLogsRepository }} request
```
crosslandwa: I've noticed that the `HapiRequest` type definition has `organisationRepository: OrganisationsRepository` declared on it.

I'm thinking we should update the `HapiRequest` type to also have `systemLogsRepository: SystemLogsRepository` declared (instead of doing the union type here)

crosslandwa: ^ though that could be done outside of this PR

src/routes/v1/me/organisations/get.js
crosslandwa: I think auditing when the DefraId token is missing information that is required  for reconciliation is a **must** - this has occurred in the live service, and I think this is one of the "user could not access service" scenarios that this new system log could help us triage

dansweeting: Hadn't realised malformed tokens were contributing to that. Things are fundamentally broken if the token is missing these things. Happy to add more scenarios into the ticket for this, potentially as a separately categorised log entry with a different identifier

src/routes/v1/me/organisations/get.js
dansweeting: easily fixed in this PR 👍

src/routes/v1/me/organisations/get.js
crosslandwa: Looking at this a bit more, I think the `throw Boom.forbidden` in `getCurrentDetailsFromToken` is the wrong design.

Access to this endpoint has already been _authenticated_ at the hapi auth layer (authentication relies on the request presenting a JWT that has been minted by Defra ID).

If the token is then missing org/email info, returning a 200 response with "couldn't find anything to link you to" seems reasonable (and the implementation can short circuit the lookup of `[linkedOrg, linkableOrgs]` in this scenario).

What's unclear to me is what the end-user experience would be before/after if the `/me/organisations` request started returning a 200 where previously it had returned a 403...

dansweeting: Returning 200 - that would make sense but IMO only about as much sense as the 403 already does. And would need orchestrating with the frontend before making that change.

As you say: the user is already _authenticated_ at that point, but it's still a matter of debate wether they're _authorised_ if we can't figure out what organisation they are from. I'd argue that we can't place them in any given set of users without the org id, so 403 is appropriate. The fact its throwing an error is probably the contentious point - that throw seems to be one of many ways of getting to http 403.

PR #1125, opened 2026-04-27
PAE-1384: validate emitted log shapes against cdp indexed-fields schema

loveboat:
Ticket: [PAE-1384](https://eaflood.atlassian.net/browse/PAE-1384)
test-time joi validator for log shapes emitted by epr-backend, derived from cdp's opensearch ingestion allowlist + index template. catches drift before unindexed fields silently land in the dropped-fields s3 sink (the pae-1384 root cause).

## what this adds

- `scripts/generate-log-schema/` — generator that reads upstream cdp sources (`cdp-tf-modules/opensearch_ingestion/vars.tf` + `cdp-tf-core/files/cdp-logs-index-template.json`), emits `parsed-sources.json` (78 keys mapped to opensearch types) + `sources.lock.json` (git refs + sha256 of upstream files)
- `src/common/helpers/logging/log-schema.js` — runtime wrapper that builds a nested joi schema from `parsed-sources.json` at import time. no generated js source; single source of truth
- `src/common/helpers/logging/log-schema.test-helper.js` — `expectLogToBeCdpCompliant(logObject)` mirrors the `err → error` rename + serializer transform that `@elastic/ecs-pino-format` applies, then validates against the schema
- `src/common/helpers/logging/err-serializer.js` — extracted from `logger-options.js` so the test helper applies the same transform pino does in prod
- two npm scripts: `log-schema:regen` and `log-schema:check`

## wired into

- `src/plugins/auth-failure-logger.test.js` — 401 emission validates
- `src/plugins/boom-error-logger.test.js` — 4 routes (400/404/500/enriched) parametrised
- `src/common/helpers/fail-action.test.js` — joi/boom/generic error paths parametrised

## pivots from the bead's plan

- used `cdp-logs-index-template.json` (authoritative; ships with the platform) instead of vendoring a sunny-curated tsv
- runtime-build pattern: `log-schema.js` is a 10-line wrapper that builds the joi schema from json at import time. no generated source code, no round-trip drift class to maintain
- tiny lock file (sha256 + git ref) for upstream-drift detection via `npm run log-schema:check`. ci doesn't have sibling repos so the check is local; can be wired to a scheduled action later

## known finding worth a follow-up

the err serializer adds `error.cause: { type, code }` when an error has a `.cause` chain — but `error.cause` is not in cdp's allowlist. those fields land in the dropped-fields s3 sink and are unsearchable. the validator catches this; needs either dropping the enrichment or lobbying cdp to allowlist `error/cause/type` and `error/cause/code`.

## test plan

- [ ] `npm test` green (5865 tests pass locally)
- [ ] `npm run lint` clean
- [ ] `npm run lint:types` clean
- [ ] `npm run log-schema:check` reports no drift
- [ ] `npm run log-schema:regen` is idempotent against current upstream
- [ ] manually inspect `parsed-sources.json` against `cdp-tf-modules/opensearch_ingestion/vars.tf` `include_keys`

[PAE-1384]: https://eaflood.atlassian.net/browse/PAE-1384?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


loveboat:
Superseded by the advisory/strict split — see #1127 (advisory validator) and #1128 (cleanup + flip to strict). Branch left intact for reference; can be deleted once #1127 + #1128 merge.

PR #1131, opened 2026-04-28
PAE-1412: hand-rolled cdp log-schema validator and error-cause cleanup

loveboat:
Ticket: [PAE-1412](https://eaflood.atlassian.net/browse/PAE-1412)
adds a joi-based validator for the cdp indexed-fields log shape and removes `error.cause` emission so logs stop landing in the dropped-fields sink.

- adds a hand-rolled log schema (`schema.js`) describing the cdp indexed-fields shape; rejects unknown keys at every level
- adds a strict test helper (`schema-test-helpers.js`); wired into `boom-error-logger`, `auth-failure-logger`, `fail-action`, `process-import-file`
- adds `enrich-boom` factories (`badRequest`/`conflict`/`internal`)
- adds per-domain error-code enums with lowercase snake_case wire values
- drops `error.cause` emission; classifier signal now in `event.reason` at the throw site
- reshapes mongo dup-key + fetch-json to use the new helpers

[PAE-1412]: https://eaflood.atlassian.net/browse/PAE-1412?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


loveboat:
closing in favour of a types-first approach for PAE-1412. the cleanup work here will be re-applied on top of the new typed logger to prove the types catch what they should.

PR #1133, opened 2026-04-28
PAE-1382: Add ledger read primitives

GraemeF:
## Summary

Adds the read-side primitives the upcoming flag-gated migration of summary-log row writes (PR #TBD), PRN write paths, and balance reads will depend on.

## Port additions (`LedgerRepository`)

- `findCreditedAmountsByWasteRecordIds(wasteRecordIds)` — bulk signed sum (credits positive, debits negative); ids without ledger transactions appear in the result with value `0`. Resolved as a single MongoDB aggregation; the in-memory adapter mirrors the semantics. Drives the per-row delta-reconciliation invariant in summary-log writes (`delta = targetAmount − creditedAmount`).
- `findLatestPerAccreditationByOrganisationId(organisationId)` and `findLatestPerAccreditationByRegistrationId(registrationId)` — one transaction per accreditation under the parent, latest only. Single aggregation pass against the denormalised parent field — no join through accreditation.

Each port primitive is exercised by a contract test suite that runs against both the in-memory and MongoDB adapters, so behaviour is identical across them.

## Application layer (`application/read-balances.js`)

Shapes port output for callers:

- `getCurrentBalance(repository, accreditationId)` — closing balance of the highest-numbered transaction; zeros when no transactions exist.
- `getCreditedAmountByWasteRecordId(repository, wasteRecordId)` — singular wrapper around the bulk primitive.
- `getOrganisationBalances(repository, organisationId)` and `getRegistrationBalances(repository, registrationId)` — return `Array<{ accreditationId, amount, availableAmount }>` sourced from the latest transaction per accreditation under the parent.

## Indexes

Two new compound indexes —

- `(organisationId, accreditationId, number desc)`
- `(registrationId, accreditationId, number desc)`

— so the parent-level "latest per accreditation" aggregation resolves without a collection scan: the prefix match on `organisationId`/`registrationId` is followed by an in-index sort that matches the aggregation's `$sort: { accreditationId: 1, number: -1 }` exactly.

## Why

Implements the read-side primitives that have a concrete consumer in the upcoming write/read migration work for the [waste balance transaction ledger](docs/architecture/decisions/0031-waste-balance-transaction-ledger.md). Speculative primitives without a consumer (per-summary-log, per-waste-record, per-PRN, per-row listings) are deliberately left out — they go in when their consumer lands.


src/waste-balances/repository/ledger-inmemory.js
```diff
@@ -18,67 +19,149 @@ import { validateLedgerTransactionInsert } from './ledger-validation.js'
...
+ * @returns {(wasteRecordIds: string[]) => Promise<Map<string, number>>}
+ */
+const performFindCreditedAmountsByWasteRecordIds =
+  (storage) => async (wasteRecordIds) => {
+    const result = new Map(wasteRecordIds.map((id) => [id, 0]))
+    for (const transaction of storage) {
+      const row = summaryLogRowOf(transaction)
+      const current = row ? result.get(row.wasteRecordId) : undefined
+      if (current !== undefined && row) {
+        result.set(row.wasteRecordId, current + signedAmount(transaction))
```
crosslandwa, line 132: Should this use the `add` function from `decimal-utils`?

(I can't recall if the waste balance is whole numbers or not)

GraemeF:
Closing without merging. Restructuring the ledger work so each port primitive lands alongside its first consumer rather than ahead of them. The credited-amounts bulk read will land with the summary-log row write migration; the parent-latest reads, the two compound indexes, and the balance application functions will land with the balance read migration.

PR #1135, opened 2026-04-29
PAE-1295: Add pagination and search to GET /v1/organisations

sommerville-ee-defra:
Ticket: [PAE-1295](https://eaflood.atlassian.net/browse/PAE-1295)
## Summary
The `GET /v1/organisations` endpoint previously returned the entire collection unfiltered. With over 5,000 organisations in TEST, this caused a backend journey test timeout and forced the admin UI to fetch the full dataset on every page load. This change adds server-side pagination and case-insensitive name search, while preserving the existing top-level array response when no query parameters are supplied so the backend can be deployed independently of the Admin UI.

## What Changed
- Added `?search=`, `?page=`, and `?pageSize=` query parameters to `GET /v1/organisations`
- Added `findPage` to the organisations repository, implemented in both the MongoDB and in-memory adapters
- Results are sorted alphabetically by company name; search is a case-insensitive substring match on the name field
- Preserved the legacy array response shape when no recognised query parameters are present, so the as-yet-unmigrated admin UI keeps working
- Added a contract test exercising both adapters plus route integration tests covering both response modes and validation

## Why
The endpoint is consumed only by the Admin UI, which currently fetches everything and filters client-side. Moving the work server-side fixes the journey test timeout and removes the Admin UI's bandwidth and rendering pressure once it adopts the new shape. The dual-mode response keeps backend and frontend deploys independent, matching the project's deployment policy.

[PAE-1295]: https://eaflood.atlassian.net/browse/PAE-1295?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/repositories/organisations/mongodb.js
+++ b/src/repositories/organisations/mongodb.js
@@ -3,6 +3,7 @@ import Boom from '@hapi/boom'
 import { ObjectId } from 'mongodb'
 import {
   createInitialStatusHistory,
+  escapeRegex,
   mapDocumentWithCurrentStatuses,
   performFindAllForOverseasSitesAdminList,
   performFindPageForOrsAdminList,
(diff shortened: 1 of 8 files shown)


src/repositories/organisations/mongodb.js
```diff
@@ -210,6 +211,35 @@ const performFindAll = (db) => async () => {
...
+              $options: 'i'
+            }
+          }
+
+    const collection = db.collection(COLLECTION_NAME)
+    const totalItems = await collection.countDocuments(filter)
+    const docs = await collection
+      .find(filter)
+      .sort({ 'companyDetails.name': 1 })
+      .skip((page - 1) * pageSize)
```
senthilinfy, line 233: Is keyset pagination preferable? Dont know if its really needed for our use case as update frequency is low

```
// To get the "next page" after a specific document
db.collection.find({
  _id: { $gt: last_id_from_previous_page }
})
.sort({ _id: 1 })
.limit(10)
```

sommerville-ee-defra: I think GDS pagination is numbered.(i.e. ?page=N), whereas keyset only gives you prev/next?  And I'm not sure the usage rate is really high enough to warrant it? Also, I think operators are likely to browse alphabetically?

senthilinfy (approved):

PR #1136, opened 2026-04-29
PAE-1309: Refactor safeAudit to use factory pattern

dansweeting:
Ticket: [PAE-1309](https://eaflood.atlassian.net/browse/PAE-1309)
## Summary

Refactors `safeAudit` so callers never pass full entity snapshots to the audit pipeline. Instead of accepting a flat payload, `safeAudit` now takes `({ event, user }, buildAuditContext)` where `buildAuditContext` is a factory function that builds only the audit-relevant identifiers.

- **New signature**: `safeAudit({ event, user }, () => ({ organisationId, ... }))`. The factory is a cognitive prompt forcing callers to explicitly choose what matters for the audit event.
- **`isPayloadSmallEnoughToAudit` is now private**: no longer exported, removing the pattern where callers conditionally reduced payloads before calling `safeAudit`.
- **All 13 call sites migrated**: six that previously did manual size-checking and conditional reduction now pass a simple factory returning identifiers; seven that already passed small contexts now wrap them in a factory.
- **System logs unchanged**: callers continue passing full entity snapshots to `recordSystemLog` / `systemLogsRepository.insert` as before.
- **Safety net preserved**: if a factory ever produces output that exceeds the size limit, `safeAudit` still falls back to sending event + user only, with a warning log.

Resolves PAE-1309.

[PAE-1309]: https://eaflood.atlassian.net/browse/PAE-1309?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/auditing/form-submission-lineage-migration.js
```diff
@@ -28,32 +28,18 @@ async function auditFormSubmissionLineageMigration(
...
-  // CDP audit has size limits - send reduced context if too big
-  const safeAuditingPayload = isPayloadSmallEnoughToAudit(payload)
-    ? payload
-    : {
-        ...payload,
-        context: { organisationId }
-      }
-
-  safeAudit(safeAuditingPayload)
+  safeAudit({ event: EVENT, user: SYSTEM_USER }, () => ({ organisationId }))
```
crosslandwa: This has changed such that information is now being dropped from the audit record in the happy path case

**before**

```
if payload small
context: {organisationId, previous, next}

if payload large
context: {organisationId}
```

**after**

```
if payload small
context: {organisationId}

if payload large
context: {organisationId}
```

This change should not have impacted what information is audited

crosslandwa (changes requested):

src/reports/application/audit.js
```diff
@@ -37,12 +36,26 @@ export async function auditReportStatusTransition(request, params) {
...
-      subCategory: AUDIT_SUB_CATEGORY,
-      action: 'status-transition'
-    },
+  const event = {
+    category: AUDIT_CATEGORY,
+    subCategory: AUDIT_SUB_CATEGORY,
+    action: 'status-transition'
+  }
+
+  safeAudit({ event, user }, () => ({
```
senthilinfy, line 45: It looks like safeAudit will miss useful fields now? Not sure if I am misreading it

**NOW**
```
const event = {
    category: AUDIT_CATEGORY,
    subCategory: AUDIT_SUB_CATEGORY,
    action: 'status-transition'
  }
```

**BEFORE**
```
{
        ...payload,
        context: {
          organisationId,
          registrationId,
          year,
          cadence,
          period,
          submissionNumber,
          reportId,
          previous: { status: previous.status.currentStatus },
          next: { status: next.status.currentStatus }
        }
      }
```

src/auditing/form-submission-lineage-migration.js
crosslandwa: I guess I was expecting the call site to look like


```
const payload = { ... }

safeAudit(
  payload,
  () => ({
    ...payload,
    context: { organisationId }
  })
)
```

and the signature of `safeAudit` to look like

```
function safeAudit(
  /** @type {Payload} */ payload,
  /** @type {() => Payload} */ safePayloadProvider,
) { ... }
```

dansweeting: I suspect at some point there's been some uncertainty/lack of decision about what data is required for audit, and no decision meant we threw the kitchen sink at the audit payload. But that exceeded sizes, hence this business of safely stripping data.

> This change should not have impacted what information is audited

_Except it was always impacted when size exceeds some arbitrary limit of the audit library_

The point of the callback arg is to force every client to think carefully about what data is necessary on the call to audit (which I'm still very unclear on requirements other than a general 'more is better' tendency from the code).

I'm happy to layer back in as much context as we would've had before so we 'lose' no data that we were logging before

However I'm keen to just always run the callback because the alternative as suggested would involve
- measuring the raw payload
- if too large, run the callback for a stripped object
- if THAT is too large, do some basic stripping and lose even more data

which results in a THREE tier variety of audited logs (the use of said audit logs is still not clear to me anyway)

src/reports/application/audit.js
dansweeting: Not sure if you were looking at a stale commit in the PR but lines 46-54 below already include the fields you pointed out in BEFORE

src/auditing/form-submission-lineage-migration.js
dansweeting: I'm resolving this because:
- this module is being deleted anyway, it was dead code
- the other places where we call safeAudit all retain the info as before

PR #1137, opened 2026-04-29
PAE-1382: Store ledger amounts as Decimal128

GraemeF:

## Summary

The new waste balance ledger collection (added under ADR 0031 / PAE-1382) was persisting `amount` and the opening / closing balance snapshots as BSON Double. The schema declared them as `Joi.number()` and the MongoDB adapter `insertMany`d the validated transactions unchanged.

ADR 0031 cites ~73,000 transactions per accreditation per year as the motivating scale for the migration. At that scale, cumulative IEEE 754 drift would silently move balances away from operator-entered tonnages — a regression versus the existing waste-balance code, which already runs on `decimal.js` arithmetic and is structurally compatible with Decimal128 storage. The drift would also undermine the delta-zero-skip and ledger-derived balance invariants the remaining migration work depends on.

## What changed

A new module, `src/waste-balances/repository/ledger-decimal.js`, owns the BSON encoding for amount fields. The MongoDB adapter calls it on the way in (`Decimal128.fromString(toDecimalString(value))`) and on the way out (existing `toNumber` helper, which already handles BSON Decimal128 + plain Numbers).

- Top-level `amount` and the `openingBalance` / `closingBalance` snapshots are persisted as `Decimal128`. Slot `number`, identifiers, and dates are unchanged.
- Schema and contract tests stay in JS-number shape — encoding is contained inside the MongoDB adapter where the BSON typing matters. Callers see no API change.
- The in-memory adapter is unchanged: it does no arithmetic over amounts and round-trips JS numbers via `structuredClone`.
- A raw-document `instanceof Decimal128` assertion in `ledger-mongodb.test.js` guards the BSON shape so that any future regression that drops the encoding step fails loudly.
- A high-precision round-trip fixture (`200.005`) lives in the contract suite so both adapters are checked for parity on values that would expose drift if anyone broke the storage shape.

## Notes

- No migration is needed. The `FEATURE_FLAG_WASTE_BALANCE_LEDGER` default is `false` and there are no environment overrides in `cdp-app-config`. `appendToLedger` has no production callers yet, so no rows exist to convert.
- No aggregation pipelines touch amount fields today — the only ledger ops are `insertMany`, `findOne`, and `createIndex`. Aggregation primitives land alongside the migration of the write and read paths; literal numeric operands inside those will need `$toDecimal` coercion at that point.
- Schema acceptance was deliberately narrowed to JS numbers only. No caller produces a Decimal128 at the schema boundary, so accepting it would have been dead-code accommodation. The decimal arithmetic discipline is upheld by where the encoding lives (the MongoDB adapter), not by what the schema additionally accepts.
- The corresponding "Decimal arithmetic" amendment to ADR 0031 lands in a separate epr-re-ex-service PR.

### DIFF
--- a/src/waste-balances/repository/ledger-contract/findLatestByAccreditationId.contract.js
+++ b/src/waste-balances/repository/ledger-contract/findLatestByAccreditationId.contract.js
@@ -69,5 +69,24 @@ export const testFindLatestByAccreditationIdBehaviour = (it) => {
       expect(x.number).toBe(1)
       expect(y.number).toBe(5)
     })
+
+    it('round-trips high-precision amounts exactly', async () => {
+      await repository.insertTransactions([
+        buildLedgerTransaction({
+          accreditationId: 'acc-precision',
+          number: 1,
+          amount: 200.005,
+          openingBalance: { amount: 0, availableAmount: 0 },
+          closingBalance: { amount: 200.005, availableAmount: 200.005 }
+        })
+      ])
+
+      const result =
+        await repository.findLatestByAccreditationId('acc-precision')
+
+      expect(result.amount).toBe(200.005)
+      expect(result.closingBalance.amount).toBe(200.005)
+      expect(result.closingBalance.availableAmount).toBe(200.005)
+    })
   })
 }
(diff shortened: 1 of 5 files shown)


loveboat (approved):
approved - clean encoding seam, well-tested. minor optional polish noted in review (test name framing, jsdoc @import placement)

PR #1139, opened 2026-04-29
PAE-1382: Route summary-log writes through the ledger (flag-gated)

GraemeF:
## Summary

Routes the summary-log row write path through the new waste-balance ledger when the `wasteBalanceLedger` feature flag is on. ADR 0031 specifies the long-term shape; this PR is the first call site that actually emits ledger transactions during a real user flow. Default behaviour is unchanged — every environment ships with the flag off.

## What changes when the flag is on

The summary-log submit pipeline now resolves a target amount per row, reads the running `wasteRecord.creditedAmount` per `(type, rowId)` once per record from the latest matching ledger transaction, and appends one ledger transaction per row whose delta is non-zero. Each new transaction stamps its own running `wasteRecord.creditedAmount` (the value `previousCreditedAmount + delta` — equal to `targetAmount`), so the per-row read is a single seek rather than an aggregation across history. A re-upload of identical data appends nothing; the recovery path described in ADR 0031 (crash mid-bulkWrite, summary log stuck in SUBMITTING until TTL) converges on the next submit.

The waste-record identity is the compound `(type, rowId)`, dodging the PAE-1380 collision class. The bulk-read primitive scopes by `accreditationId` because `(type, rowId)` is unique within an accreditation but not globally.

The pending-debit transaction type is recognised but does not contribute to credited totals: it ringfences balance, not credit attributable to a specific waste record. Pending debits do not carry a `source.summaryLogRow` path, so they are naturally outside the per-row reconciliation index and the read primitive needs no pending-debit filter.

## What is unchanged

When the flag is off, the existing embedded `transactions[]` write path runs untouched. No call sites were deleted; the v1 path will be removed in a later step once the flag has flipped on every environment and the back-fill is complete.

Audit emission is preserved on both paths. A new shared helper (`src/waste-balances/application/audit.js`) emits the `safeAudit` event and the `systemLogsRepository` insert, so the back-office system-logs view and the CDP audit stream see identical lifecycle events whichever path produced them.

## Notable implementation details

- `source.summaryLogRow` is `{ summaryLogId, wasteRecord: { type, rowId, versionId, creditedAmount } }`. Waste-record state nests inside `wasteRecord`, so each field name drops its redundant entity prefix (`type` not `rowType`, `versionId` not `wasteRecordVersionId`, `creditedAmount` not `closingCreditedAmount`) — the path provides the scope.
- The new `LedgerRepository#findLatestCreditedAmountsByWasteRecords(accreditationId, wasteRecords)` primitive returns a lookup function `(record) => number`. Mongo backs it with one indexed `findOne` per record (descending on `number`, `limit(1)`) against a new compound index `(accreditationId, source.summaryLogRow.wasteRecord.type, source.summaryLogRow.wasteRecord.rowId, number desc)`. Stringified-key Map convention is private to each adapter — never persisted, never exposed outside the primitive.
- `wasteRecord.creditedAmount` flows through the existing Decimal128 storage encoding. Decimal arithmetic discipline runs end-to-end on the v2 path; the zero-delta check uses `equals`, not `===`.
- `appendToLedger` uses an ordered insert under the hood, so a slot conflict on row N leaves rows 1..N-1 persisted and the audit emission for that batch never fires. ADR 0031 accepts this trade-off explicitly: the next re-upload converges the ledger and emits a single combined audit event.

## Migration

No data migration. The flag stays off until the back-fill bead lands and the rollout plan kicks in. The v1 embedded-array path remains the source of truth in production until then.


src/waste-balances/application/update-via-ledger.js
```diff
@@ -0,0 +1,198 @@
...
+    if (builder) {
+      builders.push(builder)
+    }
+  }
+
+  if (builders.length === 0) {
+    return
+  }
+
+  const organisationId = wasteRecords[0]?.organisationId
```
senthilinfy: is optional chaining needed? org/reg id can't be null right

src/waste-balances/application/update-via-ledger.js
```diff
@@ -0,0 +1,198 @@
...
+    builders
+  )
+
+  const last = newTransactions[newTransactions.length - 1]
+
+  await recordWasteBalanceUpdateAudit({
+    systemLogsRepository: dependencies.systemLogsRepository,
+    accreditationId: accreditation.id,
+    amount: last.closingBalance.amount,
+    availableAmount: last.closingBalance.availableAmount,
```
senthilinfy, line 194: nice ,just auditing latest instead of all updates

src/waste-balances/repository/helpers.js
```diff
@@ -222,11 +179,29 @@ export const performUpdateWasteBalanceTransactions = async ({
 
   const validatedAccreditationId = validateAccreditationId(accreditation.id)
 
+  if (dependencies.featureFlags?.isWasteBalanceLedgerEnabled?.()) {
```
senthilinfy: is optional chaining needed here

src/waste-balances/application/update-via-ledger.js
GraemeF: It's redundant because the transactions are the history. So if we want to see them then we should go to the transactions rather than the audit log. I think that's a good thing, but will check with @crosslandwa on Tuesday; can always change it before switching on!

senthilinfy (approved):

PR #1144, opened 2026-05-01
PAE-1381: Add admin unsubmit endpoint for accreditation reports

senthilinfy:
Ticket: [PAE-1381](https://eaflood.atlassian.net/browse/PAE-1381)
## Summary

  Three approaches were considered for unsubmitting a report:                                        
                                                                                                                                                                                                                                                                                                                                                                                        
  - Extend the existing operator status route with a role check — rejected, pollutes shared logic between operator and admin flows                    
  - Separate admin route reusing updateReportStatus with a dedicated slot (chosen) — passes slot: 'unsubmitted' which writes to status.unsubmitted without touching status.ready or status.submitted; all original timestamps are preserved and status.unsubmitted provides a clean audit record matching the pattern of status.submitted                                               
  - Dedicated unsubmitReport repository method — considered but unnecessary once the slot parameter on updateReportStatus was identified as sufficient; avoids duplicating repository logic                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                                                                                        
  The transition moves submitted → ready_to_submit so the operator can delete and recreate via the existing flow to pick up corrected figures. A 409 is returned if the report is not in submitted status. All transitions are audit-logged via auditReportStatusTransition.                                                                                                            
    

- Add `POST .../reports/{year}/{cadence}/{period}/unsubmit` restricted to `serviceMaintainer` role
- Move status slot to domain layer and pass as parameter to updateStatus 
- Gate endpoint behind `FEATURE_FLAG_REPORT_UNSUBMIT` (env var, default off) 


[PAE-1381]: https://eaflood.atlassian.net/browse/PAE-1381?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/feature-flags/feature-flags.config-schema.js
```diff
@@ -0,0 +1,50 @@
+export const featureFlagsSchema = {
```
crosslandwa: what's the motivation for extracting this from `src/config.js`?

src/plugins/router.js
```diff
@@ -34,8 +35,15 @@ const router = {
...
 
+        const { reportsUnsubmit: _unsubmit, ...coreReportsRoutes } =
+          reportsRoutes
+
         const reportsBehindFeatureFlag = featureFlags.isReportsEnabled()
-          ? Object.values(reportsRoutes)
+          ? Object.values(coreReportsRoutes)
+          : []
+
+        const unsubmitBehindFeatureFlag = featureFlags.isReportUnsubmitEnabled()
```
crosslandwa, line 41: Maybe we should proceed this PR with a smaller one to remove the `isReportsEnabled` feature flag.

The unsubmit endpoint should only _really_ be available if **both** the `isReportsEnabled` and `isReportUnsubmitEnabled` feature flags are enabled

As-is, we _could_ in theory have reports disabled, but still allow the unsubmission endpoint to be called (where the expected behaviour is ambiguous.unspecified...)

Alternatively, update this to

```
const unsubmitBehindFeatureFlag = featureFlags.isReportUnsubmitEnabled() && featureFlags.isReportsEnabled()
  ? [reportsUnsubmit]
  : []
```

src/reports/repository/mongodb.js
```diff
@@ -41,6 +43,27 @@ const MONGODB_DUPLICATE_KEY_ERROR_CODE = 11000
 const reportsCollection = (db) =>
   /** @type {Collection<Report>} */ (db.collection(REPORTS_COLLECTION))
 
+/**
+ * Resolves a failed findOneAndUpdate into a 404 (report missing) or 409 (version mismatch).
+ *
+ * @param {Db} db
+ * @param {string} reportId
+ * @param {number} version
+ * @returns {Promise<never>}
```
crosslandwa, line 49: maybe complement the `@returns` with an `@throws`

src/feature-flags/feature-flags.config-schema.js
senthilinfy: Its down to sonar complaining about file size being >500 lines. I can revert the split and add rule exclusion for the file if we think thats better

src/plugins/router.js
senthilinfy: will raise a PR first to remove reportsEnabled flag

src/feature-flags/feature-flags.config-schema.js
senthilinfy: I decided to add an exclusion, probably makes sense for config.js

src/config.js
```diff
@@ -342,6 +342,12 @@ const baseConfig = {
       default: false,
       env: 'FEATURE_FLAG_COPY_FORM_FILES_TO_S3'
     },
+    reports: {
```
crosslandwa: Presume adding this back is unintentional

crosslandwa (approved):

PR #1145, opened 2026-05-01
PAE-1421: Add regulator column to report submissions API

timrutter:
## Summary

- Adds a `regulator` field to every row of the `GET /v1/organisations/reports/submissions` response
- Sourced from `registration.submittedToRegulator` (lowercase, required) and converted via a `REGULATOR_DISPLAY` lookup map
- Joi response schema enforces `.valid('EA','NIEA','SEPA','NRW')` — typos or unexpected values from the producer are rejected at the boundary
- Unblocks the admin FE work that adds "Regulator" as the first CSV column ([epr-re-ex-admin-frontend#375](https://github.com/DEFRA/epr-re-ex-admin-frontend/pull/375))

## Changes

- `src/domain/organisations/model.js` — adds `REGULATOR_DISPLAY` constant alongside `REGULATOR`, mapping each lowercase code to its display form; adds `RegulatorDisplay` typedef
- `src/reports/application/report-submissions.js` — extends `ReportSubmissionsRow` typedef using `RegulatorDisplay`; imports `REGULATOR_DISPLAY`; populates `regulator` in `buildRow` via lookup
- `src/reports/routes/submissions.js` — adds `regulator: Joi.string().valid(...Object.values(REGULATOR_DISPLAY)).required()` as the first field of the response schema
- `src/reports/application/report-submissions.test.js` — extends integration fixture's `baseRow`; adds parameterised `it.each` covering all four regulators (`ea`/`niea`/`sepa`/`nrw`); adds `submittedToRegulator: 'ea'` to mock helper default

### DIFF
--- a/src/reports/routes/submissions.js
+++ b/src/reports/routes/submissions.js
@@ -20,7 +20,7 @@ export const getReportSubmissions = {
           .items(
             Joi.object({
               regulator: Joi.string()
-                .valid(...Object.values(REGULATOR_DISPLAY), '')
+                .valid(...Object.values(REGULATOR_DISPLAY))
                 .required(),
               organisationName: Joi.string().required(),
               submitterPhone: Joi.string().required(),
--- a/src/reports/routes/submissions.js
+++ b/src/reports/routes/submissions.js
@@ -18,6 +18,7 @@ export const getReportSubmissions = {
         reportSubmissions: Joi.array()
           .items(
             Joi.object({
+              regulator: Joi.string().allow('').required(),
               organisationName: Joi.string().required(),
               submitterPhone: Joi.string().required(),
               approvedPersonsPhone: Joi.string().required(),
--- a/src/reports/routes/submissions.js
+++ b/src/reports/routes/submissions.js
@@ -2,6 +2,7 @@ import Joi from 'joi'
 
 import { ROLES } from '#common/helpers/auth/constants.js'
 import { getAuthConfig } from '#common/helpers/auth/get-auth-config.js'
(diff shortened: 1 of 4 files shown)


src/reports/routes/submissions.js
```diff
@@ -18,6 +18,7 @@ export const getReportSubmissions = {
         reportSubmissions: Joi.array()
           .items(
             Joi.object({
+              regulator: Joi.string().allow('').required(),
```
GraemeF: Can this be a list of known constants? Can then avoid toUppercase downstream.

GraemeF: s/list/union

GraemeF: I mean, check how we do enums elsewhere, this could be like that.

src/reports/routes/submissions.js
```diff
@@ -18,6 +19,9 @@ export const getReportSubmissions = {
         reportSubmissions: Joi.array()
           .items(
             Joi.object({
+              regulator: Joi.string()
+                .valid(...Object.values(REGULATOR_DISPLAY), '')
```
GraemeF: Is there a case where it would be missing? If not can we drop the "or empty"?

src/reports/routes/submissions.js
timrutter: added an enum

src/reports/routes/submissions.js
timrutter: have removed the emptiness

GraemeF (approved):

PR #1154, opened 2026-05-07
PAE-1382: Route waste-balance reads through canonicalSource marker

GraemeF:
## Summary

When a waste-balance document carries `canonicalSource: 'ledger'`, the embedded `amount` / `availableAmount` are stale; the latest ledger transaction's `closingBalance` is the source of truth. `'embedded'` and `'migrating'` markers continue to read from the embedded fields unchanged because the embedded write path is still authoritative for them.

Substitution happens transparently inside `findByAccreditationId` and `findByAccreditationIds` so consumers (PRN lifecycle handlers, the `/v1/organisations/{id}/waste-balances` endpoint) see correct amounts without any caller-side wiring. The marker-driven routing matches the rollout design in [`waste-balance-ledger-rollout.md`](https://github.com/DEFRA/epr-re-ex-service/blob/main/docs/architecture/discovery/waste-balance-ledger-rollout.md).

Today no production accreditation has marker `'ledger'` (no sweep has run, no flag is on), so this change is a no-op until the rollout begins. The new contract tests exercise the substitution paths against both the in-memory and MongoDB adapters.

### DIFF
--- a/src/waste-balances/repository/marker-aware-read.js
+++ b/src/waste-balances/repository/marker-aware-read.js
@@ -11,24 +11,13 @@ import { WASTE_BALANCE_CANONICAL_SOURCE } from '../domain/model.js'
  * sweep populates the ledger before flipping the marker — but is treated as a
  * zero balance to keep the read path safe.
  *
- * `ledgerRepository` is optional so tests that exercise embedded/migrating
- * paths can wire the waste-balances repository without a ledger; an undefined
- * ledger combined with marker `'ledger'` is a wiring bug and trips an explicit
- * error rather than silently returning stale embedded amounts.
- *
  * @param {import('../domain/model.js').WasteBalance} balance
- * @param {import('./ledger-port.js').LedgerRepository | undefined} ledgerRepository
+ * @param {import('./ledger-port.js').LedgerRepository} ledgerRepository
  * @returns {Promise<import('../domain/model.js').WasteBalance>}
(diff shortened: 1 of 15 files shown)


src/waste-balances/repository/marker-aware-read.js
```diff
@@ -0,0 +1,36 @@
...
+export const resolveBalanceAmounts = async (balance, ledgerRepository) => {
+  if (balance.canonicalSource !== WASTE_BALANCE_CANONICAL_SOURCE.LEDGER) {
+    return balance
+  }
+
+  const latest = await ledgerRepository.findLatestByAccreditationId(
+    balance.accreditationId
+  )
+
+  if (!latest) {
```
senthilinfy, line 29: If this is not expected better to throw error? Otherwise unknown bug will remain unknown to a point

senthilinfy (approved):

src/waste-balances/repository/marker-aware-read.js
GraemeF: Agreed — `0` is a legitimate balance, so silently substituting it for the marker-with-empty-ledger case made a corruption indistinguishable from a real zero. Switched to throwing `Boom.internal` in [a1f777ed](https://github.com/DEFRA/epr-backend/pull/1154/commits/a1f777ed) so the invariant violation surfaces.

PR #1155, opened 2026-05-07
PAE-1439: Add optimistic concurrency control to PRN status updates

GraemeF:
Ticket: [PAE-1439](https://eaflood.atlassian.net/browse/PAE-1439)

## Summary

Two concurrent POSTs to the PRN status route were both passing validation, both writing the balance effect, and only one PRN write winning. The double write silently corrupted balances. This PR closes the race on two fronts: the PRN repository now CAS-gates writes on a per-document `version`, and `updatePrnStatus` runs the balance effect only after the CAS has chosen a winner — for every transition on an existing PRN.

## Repository CAS

- The PRN schema declares a `version` field (integer, defaults to 1 on insert).
- `updateStatus` CAS-gates writes by version. A missed CAS is disambiguated into 404 (document gone) or 409 (`Boom.conflict` with descriptive message and a `database / version_conflict_detected` structured log) by re-fetching just the version.
- The application layer threads the read version through to the repository so the CAS is exact.
- Documents missing the `version` field are tolerated at the adapter boundary rather than backfilled. Reads compensate via the Joi default on `version`, so a legacy document reads back as `version: 1`. The MongoDB CAS filter uses `$expr` with `$ifNull` so an expected version of 1 matches both `{ version: 1 }` and documents with no `version` field, and the update assigns `version: <expected> + 1` rather than incrementing — keeping post-state correct whether the field was present or absent. A backfill cannot guarantee that every document is touched before a fresh write lands, so per-call tolerance is the safer model.
- The in-memory adapter mirrors the same conflict semantics so both adapters share one contract test suite (`optimistic-concurrency.contract.js`) covering initialisation, increment, sequential threading, descriptive messages, structured logging, concurrent races (including the issue path where both writers also generate a `prnNumber`). Versionless-document tolerance is exercised by adapter-specific tests in `mongodb.test.js` and `inmemory.test.js`, since the seeding mechanism differs per adapter.
- The repository factory now requires a `TypedLogger` so version conflicts always produce a logged event.
- A route-level test asserts that `Boom.conflict` from the repository surfaces as 409 to the caller.

## CAS gates the balance effect

`applyWasteBalanceEffects` previously ran before the PRN CAS for every transition. For any transition on an existing PRN — issuance, deletion-from-awaiting-authorisation, cancellation-from-awaiting-cancellation — concurrent writers all passed `validateTransition`, all wrote the balance effect, and only one had its PRN write rejected. `updatePrnStatus` now performs the CAS first for these transitions and only the winner reaches the balance write, eliminating the symmetric double-debit (issuance) and double-credit (cancels) races.

Creation (`AWAITING_AUTHORISATION`) keeps its pre-flight balance check. The per-PRN CAS protects only same-document writes — it cannot help a new-PRN insert — and a post-CAS failure on creation would strand the PRN in a state with no legal user exit.

## Integration tests

`update-status-concurrency.test.js` wires the real in-memory PRN and waste-balance repositories together and fires two concurrent calls on the same PRN. Three scenarios: concurrent issuances, concurrent deletions from awaiting-authorisation, and concurrent cancellations from awaiting-cancellation. Each asserts the corresponding balance method is invoked exactly once and that the losing call returns 409. All three fail against the pre-reorder code (call count was 2) and pass after the reorder.

## Trade-off

A balance write that fails after the CAS leaves a residue: a phantom issued PRN with no debit, or a cancelled PRN whose tonnage was not credited back. Each occurrence is bounded to one PRN's worth, observable in the data (PRN history disagrees with balance transactions), and the lesser harm than the unbounded double-write race it replaces. Cross-writer balance races on different PRNs of the same accreditation are unaffected — they are addressed separately by the ledger flip in [PAE-1382](https://eaflood.atlassian.net/browse/PAE-1382).

[PAE-1439]: https://eaflood.atlassian.net/browse/PAE-1439?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/packaging-recycling-notes/application/update-status.js
+++ b/src/packaging-recycling-notes/application/update-status.js
@@ -251,16 +251,6 @@ export async function updatePrnStatus({
   const currentStatus = prn.status.currentStatus
   validateTransition(currentStatus, newStatus, actor)
 
-  await applyWasteBalanceEffects(wasteBalancesRepository, {
-    currentStatus,
-    newStatus,
-    accreditationId,
-    organisationId,
-    prnId: id,
-    tonnage: prn.tonnage,
-    userId: user.id
-  })
-
   const now = updatedAt ?? new Date()
   const updateParams = {
     id,
@@ -270,7 +260,7 @@ export async function updatePrnStatus({
     updatedAt: now
   }
 
-  // Issue with PRN number generation and collision retry
+  let updatedPrn
   if (newStatus === PRN_STATUS.AWAITING_ACCEPTANCE) {
     updateParams.operation = {
       slot: 'issued',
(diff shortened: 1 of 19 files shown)


src/packaging-recycling-notes/repository/mongodb.js
```diff
@@ -244,13 +250,51 @@ const performFindByStatus = (db, excludeOrganisationIds) => {
...
+  const objectId = ObjectId.createFromHexString(id)
+  const existing = await db
+    .collection(COLLECTION_NAME)
+    .findOne({ _id: objectId }, { projection: { version: 1 } })
+
+  if (!existing) {
+    return null
+  }
+
+  const actualVersion = existing.version ?? 1
```
senthilinfy, line 272: can version ever be null? for existing wb records as well

src/packaging-recycling-notes/application/update-status.js
```diff
@@ -251,25 +251,36 @@ export async function updatePrnStatus({
...
     prnId: id,
     tonnage: prn.tonnage,
     userId: user.id
-  })
+  }
+
+  // Issuance debits the balance AFTER the CAS so concurrent issues can't
+  // double-debit. Other transitions keep the pre-flight check so an
+  // insufficient balance never produces a phantom PRN.
+  if (newStatus !== PRN_STATUS.AWAITING_ACCEPTANCE) {
```
senthilinfy: why treat debit to availableAmount and amount differently and specifically avoid double debit only for issuance?

src/packaging-recycling-notes/application/update-status.js
```diff
@@ -251,25 +251,36 @@ export async function updatePrnStatus({
...
+    await applyWasteBalanceEffects(
+      wasteBalancesRepository,
+      balanceEffectsParams
```
senthilinfy, line 327: should optimistic locking using version be done for WB as well?  thinking if that can lead to incorrect wb amount/availableAmount update. May be thats separate issue to hanlde

GraemeF: Yep that's addressed by PAE-1382 so not done here.

src/packaging-recycling-notes/application/update-status.js
GraemeF: Changed the ordering on the cancel paths. Creation doesn't matter because that's a new document anyway.

senthilinfy (approved):

PR #1163, opened 2026-05-08
PAE-1443/1444: Waste records export — admin CSV endpoint

timrutter:
Ticket: [PAE-1443](https://eaflood.atlassian.net/browse/PAE-1443)
## Summary

Backend half of the new "Waste records export" admin tab. Adds `GET /v1/admin/waste-records/export.csv` (serviceMaintainer auth) which streams a CSV of every waste record in the system.

- Async-generator orchestrator iterating organisation → registration → waste record, yielding one row per record.
- CSV header is the union of every field across every summary-log table-schema (auto-derived at module load), prefixed with metadata columns: Regulator, Organisation Name, Material, Operator Processing Type, Accredited, Waste Record Type, Submitted At, Included in Waste Balance, Row ID.
- New helpers (all unit-tested with 100% coverage): `isIncludedInWasteBalance` (boolean version of `getTargetAmount`), `buildOverseasSitesContext` (per-registration), `loadSummaryLogMap` (per-registration), `csv-columns` (column derivation + row builder).
- Route streams via `Readable.from(..., { objectMode: false })` so Hapi can pipe binary chunks to the HTTP response.
- Hard-fail on mid-stream errors (no sentinel rows).

## Spec / plan

- Spec: `docs/superpowers/specs/2026-05-08-waste-records-export-design.md` (in claude-reex meta-repo)
- Plan: `docs/superpowers/plans/2026-05-08-waste-records-export.md`

## Notes / deviations from original plan

- **Regulator** field is read from `registration.submittedToRegulator` (not `org.regulator` — that property doesn't exist).
- **Reported Period column dropped** — derived from `(cadence, period, year)` not directly on a summary log; out of scope for v1.
- **Summary-log loader is per-registration** (uses existing `findAllByOrgReg`); the repo has no `findAll()` method.
- Some Jira ticket columns (Bill of lading reference, Waste Transfer Note ref, refused/stopped/repatriated for accredited exporters) have no source field in any schema — these will not appear in the CSV and need a follow-up with the FOI requester.

## Companion changes

The admin frontend half (page, button, streaming proxy) lands separately.

[PAE-1443]: https://eaflood.atlassian.net/browse/PAE-1443?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/waste-records-export/application/stream-csv-export.js
```diff
@@ -0,0 +1,165 @@
...
+        return t || String(a.rowId).localeCompare(String(b.rowId))
+      })
+
+      for (const record of recordsSorted) {
+        for (const key of Object.keys(record.data)) {
+          observedKeys.add(key)
+        }
+      }
+
+      buffered.push({
```
senthilinfy: Does this not mean all waste records in the system is in memory and not streamed for each registration?  This will potentially increase memory usage a lot

timrutter: moving to a prequery to ascertain unique fields and then the main work to extract data can be streamed.

dansweeting (approved):

src/waste-records-export/domain/csv-columns.js
```diff
@@ -0,0 +1,138 @@
...
+  'Operator Processing Type',
+  'Accredited',
+  'Waste Record Type',
+  'Submitted At',
+  'Included in Waste Balance',
+  'Row ID'
+])
+
+const FIELDS_NEVER_EXPORTED = new Set(['ROW_ID']) // already in metadata prefix
+
```
dansweeting, line 33: `processingType` is a key on every `record.data`, so `findDistinctDataKeys` will include it as a data-field column. But it's already rendered in the metadata prefix as "Operator Processing Type". Should it be added to `FIELDS_NEVER_EXPORTED` alongside `ROW_ID`?

src/waste-records-export/application/stream-csv-export.js
```diff
@@ -0,0 +1,206 @@
...
+ * @property {WasteRecordsRepository} wasteRecordsRepository
+ * @property {SummaryLogsRepository} summaryLogsRepository
+ * @property {OverseasSitesRepository} overseasSitesRepository
+ */
+
+const sortById = (a, b) => a.id.localeCompare(b.id)
+
+const sortRecords = (a, b) => {
+  const t = a.type.localeCompare(b.type)
+  return t || String(a.rowId).localeCompare(String(b.rowId))
```
dansweeting: Nit: `String(a.rowId).localeCompare(String(b.rowId))` gives lexicographic order, so `"9"` sorts after `"10"`. If row IDs are always numeric, `Number(a.rowId) - Number(b.rowId)` would give natural ordering. The tests happen to use IDs that sort identically either way.

src/plugins/router.js
```diff
@@ -7,6 +7,7 @@ import * as devRoutes from '#routes/v1/dev/index.js'
 import { formSubmissionsRoutes } from '#routes/v1/form-submissions/index.js'
 import * as systemLogsRoutes from '#routes/v1/system-logs/index.js'
 import * as wasteBalanceRoutes from '#waste-balances/routes/index.js'
+import * as wasteRecordsExportRoutes from '../waste-records-export/index.js'
```
dansweeting: Minor: this is the only relative import in the file (everything else uses `#routes/...` hash paths). Worth adding a `#waste-records-export` alias to the import map, or is the plan to keep this module outside the routes hierarchy deliberately?

src/waste-records-export/application/stream-csv-export.js
```diff
@@ -0,0 +1,206 @@
...
+    overseasSitesRepository.findAll(),
+    organisationsRepository.findAll(),
+    wasteRecordsRepository.findDistinctDataKeys()
+  ])
+
+  const sitesById = new Map(allSites.map((s) => [s.id, s]))
+  const dataFieldColumns = buildDataFieldColumns(observedKeys)
+  yield await encodeRow(buildHeaderRow(dataFieldColumns))
+
+  const orgsSorted = [...orgs].sort(sortById)
```
dansweeting, line 185: The AC requires filtering to accredited plastic registrations only (BR-1/BR-2), but `streamCsvExport` iterates every org, every registration, every record without filtering on material or accreditation status. `isRegistrationAccredited` and `registration.material` are read to populate column values but never gate whether a row is emitted.

dansweeting: Also worth noting: BR-2 explicitly calls out excluding registered-only reprocessors/exporters (i.e. not accredited), and BR-1 scopes inclusion to accredited only. Neither filter is present in the current implementation.

dansweeting (dismissed):
see comments on BR1/BR2 - we could have some functionality gaps with what's currently set out in JIRA

src/waste-records-export/application/stream-csv-export.js
timrutter: this filtering will be a post processing task rather than done in the code

src/waste-records-export/domain/csv-columns.js
timrutter: Good catch — `record.data.processingType` is set during sync (see create-row-transformer.js:26) and `findDistinctDataKeys` returns the union of every `data` key, so without exclusion the CSV would have both `Operator Processing Type` (metadata prefix) and `processingType` (data-field column). Added `processingType` to `FIELDS_NEVER_EXPORTED` with an explanatory comment, plus a regression test.

src/waste-records-export/application/stream-csv-export.js
timrutter: Fixed, but went with `localeCompare(..., { numeric: true })` rather than `Number(a.rowId) - Number(b.rowId)`: `rowId` is typed `string` in the model and originates from the spreadsheet `ROW_ID` column, so non-numeric values are possible (and would break a numeric subtraction with NaN). Natural-order collation gives `'9'` before `'10'` while still handling non-numeric IDs. Added a regression test.

src/plugins/router.js
timrutter: Fair point — no deliberate reason. Added `#waste-records-export/*` to the imports map and switched the import to use the alias.

dansweeting (approved):

PR #1166, opened 2026-05-09
PAE-1439: Compensate post-CAS partial failures in PRN status transitions

GraemeF:
Ticket: [PAE-1439](https://eaflood.atlassian.net/browse/PAE-1439)
## What this fixes

`updatePrnStatus` co-mutates the PRN doc and the waste balance across four transitions, with no transactional bond between the two writes. If the second write throws after the first commits, the system is left in an inconsistent state. Fixed transitions:

- `draft → awaiting_authorisation` (creation, pre-flight balance debit then PRN write)
- `awaiting_authorisation → awaiting_acceptance` (issuance, PRN write then total-balance debit) — fraud vector, customer-favourable phantom
- `awaiting_authorisation → deleted` (cancel-pending, PRN write then available-balance credit)
- `awaiting_cancellation → cancelled` (cancel-issued, PRN write then full-balance credit)

## Approach

Saga-flavoured compensation pattern around each pair:

1. **Repository rollback primitives** — `rollbackIssuance`, `rollbackPendingCancellation`, `rollbackIssuedCancellation` on `PackagingRecyclingNotesRepository`. CAS-gated on the post-forward `expectedVersion`, append a normal history entry recording the reverted-to status, unset the slot fields and (for issuance) `prnNumber`. Implemented identically across in-memory and MongoDB backends and exercised through a new contract test.

2. **Compensation in `updatePrnStatus`** — forward writes are wrapped in `try/catch`. On failure, a small `ROLLBACK_METHOD_BY_TRANSITION` table dispatches to the matching repo rollback (or, for creation, just credits the balance back). If compensation itself throws, the original forward error is rethrown so the user-visible HTTP outcome stays coherent. Two paired log entries (forward error + compensation error) share the same `event.action: compensation_failure` and `event.reference: <prnId>` so ops can find the half-applied transition for manual reconciliation.

3. **Tests** — a new application-layer test file exercises every compensation path against real in-memory repos with a single throwing-method proxy. The repository rollback primitives themselves are contract-tested across both backends. The two `it.fails` markers in `integration.waste-balance-arithmetic.test.js` are removed; those tests now drive issuance compensation through the route layer naturally and pass.

## Known residual race

A producer accepting an issued PRN can land between the forward `awaiting_acceptance` write and the rollback. The producer's CAS succeeds (version 2 → 3), the rollback's CAS (`expectedVersion: 2`) then fails, and we land in the `compensation_failure` log with the PRN now `accepted` but the total balance never debited. The window is far narrower than the original phantom and is captured in the log entry for ops to reconcile, but it isn't fully closed by this change. PAE-1382's ledger flip collapses this failure mode entirely by folding the canonical balance write into the same logical issuance.

## Out of scope

- Cross-writer balance race on `saveBalance.transactions[]` — already punted to PAE-1382.
- Pruning the unreachable `awaiting_authorisation → cancelled` branch in `applyWasteBalanceEffects`. The transition is rejected by `validateTransition` so the branch is dead. An inline comment documents this; pruning the dead OR is a separate cleanup.
- Restoring original `currentStatusAt` on rollback. After rollback, `currentStatusAt` reflects the rollback timestamp, not the original entry into the reverted-to status. Treated as the new entry-into-status time (consistent with the convention used elsewhere); the original timestamp is still recoverable from the `history` array. If a `findByStatus` consumer needs the original time, that's a separate change.

[PAE-1439]: https://eaflood.atlassian.net/browse/PAE-1439?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/packaging-recycling-notes/application/update-status.js
```diff
@@ -261,13 +283,201 @@ async function applyStatusUpdate({
...
+      category: LOGGING_EVENT_CATEGORIES.DB,
+      action: LOGGING_EVENT_ACTIONS.COMPENSATION_FAILURE,
+      reference: prnId
+    }
+  })
+}
+
+async function tryCompensate(compensator, context) {
+  try {
+    await compensator()
```
senthilinfy, line 444: It would be good to log when there's a successful compensation. Logging on compensation failure is already there

src/packaging-recycling-notes/application/update-status.js
```diff
@@ -261,13 +283,201 @@ async function applyStatusUpdate({
...
+  })
+
+  try {
+    await applyWasteBalanceEffects(
+      wasteBalancesRepository,
+      balanceEffectsParams
+    )
+  } catch (forwardError) {
+    const rollbackMethod =
+      ROLLBACK_METHOD_BY_TRANSITION[`${currentStatus}|${newStatus}`]
```
senthilinfy: (Optional) Is it worth being defensive when ROLLBACK_METHOD_BY_TRANSITION doesn't have entry for the status transition? Otherwise it will log some runtime error

src/packaging-recycling-notes/application/update-status.js
```diff
@@ -261,13 +283,201 @@ async function applyStatusUpdate({
...
+ * Runs `compensator` to undo a forward write that has already committed.
+ * If the compensator itself throws, the failure is logged with full context
+ * (the original forward error and the compensation error) but not rethrown —
+ * the caller always sees the original forward error so the user-visible
+ * outcome stays coherent. The log entry is what ops use to find PRNs left
+ * in an inconsistent state for manual reconciliation.
+ *
+ * @param {() => Promise<unknown>} compensator
+ * @param {Object} context
+ * @param {Error} context.forwardError
```
senthilinfy: This JSDoc probably belongs to tryCompensate

src/packaging-recycling-notes/application/update-status.js
senthilinfy, line 444: General question, is there any audit needed when transactions fail and are being compensated?  May be its already covered in existing calls

GraemeF: Done — added a `compensation_success` event action and a `logCompensationSuccess` call from `tryCompensate` on the recovery path. Logged at `warn` level so a forward failure is still visible to ops even when we cleaned it up, with the original `forwardError` on `err` so it indexes. Pairs with `compensation_failure` via the same `event.reference` (prnId). 1b219dc5.

src/packaging-recycling-notes/application/update-status.js
GraemeF: Good catch — the JSDoc had drifted up to `logCompensationFailure` during a refactor. Moved back to `tryCompensate`. 1b219dc5.

src/packaging-recycling-notes/application/update-status.js
GraemeF: Tightened with types rather than a runtime guard. `ROLLBACK_METHOD_BY_TRANSITION` is now typed as `Readonly<Record<PostCasRollbackTransition, RollbackMethodName>>` with literal-string keys, so tsc rejects a missing entry, an orphan entry, or a stale method name. The lookup uses a JSDoc cast that carries the runtime invariant (this catch block is only reachable when `applyWasteBalanceEffects` throws, and that only fires for these three transitions). Verified by deliberately breaking the table both ways and watching tsc fire. 1b219dc5.

Worth noting: even if a future refactor broke the invariant, the bad lookup would land in the existing `compensation_failure` log path via `tryCompensate`'s catch — so observability is bounded.

src/packaging-recycling-notes/application/update-status.js
GraemeF: Open question — would value your steer. The system_logs audit (`auditPrnStatusTransition` in `routes/status.js:139`) only fires on the success path, because on a compensated failure the transition didn't take effect end-to-end. What we do have on the failure-then-compensation path:

1. The PRN doc's own `history` array gets a rollback entry written by the repo primitive — visible on the PRN itself.
2. The `compensation_success`/`compensation_failure` log entries on the same `event.reference: <prnId>` so ops can correlate.

Is there a Defra audit standard you'd want applied to compensated transitions, or is the log + PRN history sufficient for the reconciliation use-case? Happy to extend if there's a specific requirement.

senthilinfy: There's no specific DEFRA auditing standard I was referring to, haven't seen a one so :man_shrugging: .

I asked for operational needs(system logs) especially for waste balance updates. May be there's some unknown use case(say a bug in compensating transaction), and having the information in system logs helps debug.

senthilinfy (approved):

PR #1175, opened 2026-05-13
PAE-1382: Only match registrations in approved/cancelled/suspended state in divergence diagnostic

senthilinfy:
- Filter the registration lookup in the balance divergence diagnostic to only match registrations in `approved`, `cancelled`, or `suspended` state
- Registrations in `created` or `rejected` state are excluded and surface as a rebuild failure in the diagnostic log
- Add `status` field to four existing test fixtures that were missing it to keep them passing under the new filter
- Add a new test covering the case where a registration exists but is in an excluded state

### DIFF
--- a/src/domain/organisations/registration.js
+++ b/src/domain/organisations/registration.js
@@ -63,7 +63,7 @@
  * @typedef {RegistrationBase & {
  *  registrationNumber?: string;
  *  cbduNumber?: string;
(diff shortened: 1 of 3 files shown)


dansweeting (approved):
So in the prod run, all those registration=null instances would throw now, and we'll end up with something like `...failed=33` in the logs on bootup?

PR #1181, opened 2026-05-15
PAE-1424: Add compliance reporting periods to public register CSV

senthilinfy:
Ticket: [PAE-1424](https://eaflood.atlassian.net/browse/PAE-1424)
- Add dynamic compliance period columns (Jan–Dec, Q1–Q4) to the public register CSV, sourced from `generateReportCompliance`
- Include cancelled registrations in compliance reporting alongside approved and suspended
- Encode period applicability in `submittedDates` map (key present = applicable, absent = N/A; value null = not yet submitted, date string = submitted)
- Extract compliance field building out of `transformRegAcc` into a dedicated `buildReportComplianceFields` function

[PAE-1424]: https://eaflood.atlassian.net/browse/PAE-1424?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/reports/domain/compliance-reporting-periods.js
+++ b/src/reports/domain/compliance-reporting-periods.js
@@ -1,0 +1,100 @@
+import { CADENCE } from './cadence.js'
+import { generateReportingPeriods } from './generate-reporting-periods.js'
+
+/**
+ * @typedef {{
+ *   key: string;
+ *   cadence: string;
+ *   year: number;
+ *   period: number;
+ *   label: string;
+ *   startDate: string;
+ *   endDate: string;
+ *   dueDate: string;
+ * }} CompliancePeriod
+ */
+
+const MONTH_ABBR = [
+  'Jan',
+  'Feb',
+  'Mar',
+  'Apr',
+  'May',
+  'Jun',
+  'Jul',
+  'Aug',
+  'Sep',
+  'Oct',
+  'Nov',
+  'Dec'
+]
+
+/**
+ * @param {string} cadence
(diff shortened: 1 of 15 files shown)


dansweeting (commented):
The complexity around dynamic amount of columns looks might be necessary, since the AC's seem to want to show the difference between reports that were submitted versus not submitted versus N/A.

But it looks as though:
- dynamic cols are coming in an order doesn't quite match the ticket ACs (this PR has all the months listed first then followed by quarters, the ticket shows an example where we do Jan/Feb/Mar/Q1/April etc
- therefore consumers of this csv are going to be dealing with a growing number of columns month by month (and because quarters go at the end of the column list, each month/quarter that goes by, those column indexes represent different data)

The way the columns are ordered in the ticket at least gives us a predictable order when the year advances

dansweeting (approved):

src/reports/domain/compliance-reporting-periods.js
```diff
@@ -0,0 +1,107 @@
...
+}
+
+/**
+ * Returns all ended reporting periods (both cadences) for the current calendar year,
+ * from January 1 through today, ordered ascending by end date with monthly
+ * before quarterly on the same end date.
+ *
+ * @returns {CompliancePeriod[]}
+ */
+export function generateComplianceReportingPeriods() {
```
dansweeting, line 89: 👍

PR #1179, opened 2026-05-14
PAE-1382: Add event-sourced waste balance stream

dansweeting:
Ticket: [PAE-1382](https://eaflood.atlassian.net/browse/PAE-1382)

## Summary

Replaces the per-row ledger design (ADR-0031) with an event-sourced stream for waste balance tracking. See the [design document](https://github.com/DEFRA/epr-re-ex-service/blob/PAE-1382-event-sourced-stream-design/docs/architecture/discovery/event-sourced-waste-balance-stream.md) for full context. The old ledger code is deleted entirely. All new behaviour is behind the existing `FEATURE_FLAG_WASTE_BALANCE_LEDGER` feature flag, and only activates for balances whose `canonicalSource` has been promoted to `'ledger'`. Nothing currently performs that promotion, so the stream path is inert in production until the promotion orchestrator lands separately.

### What changed

- **Stream repository layer**: new `waste-balance-events` collection with schema, port, in-memory and MongoDB adapters, contract tests, and five MongoDB indexes (slot uniqueness, two partial-unique idempotency indexes, kind+number, and PRN watermark)
- **Append primitives**: `appendToStream` reads the latest event, computes closing balances per event kind, and appends; `performUpdateViaStream` aggregates row-level target amounts into a single `creditTotal` per submission
- **Summary-log dispatch**: threads `summaryLogId` through the submission call chain; when the feature flag is on and `canonicalSource === 'ledger'`, dispatches to the stream instead of the embedded `transactions[]` array
- **PRN operations dispatch**: threads `registrationId` through all four PRN balance operations (create, issue, cancel pre-issue, cancel post-issue); dispatches to the stream when `canonicalSource === 'ledger'`
- **Balance read path**: `resolveBalanceAmounts` now reads the latest stream event's `closingBalance` instead of the latest ledger transaction, for balances marked as `'ledger'`
- **Plugin wiring**: both MongoDB and in-memory plugins create and inject `streamRepository` instead of `ledgerRepository`
- **Ledger removal**: all ledger source files, tests, and contract tests deleted; `ensureLedgerCollection` removed from MongoDB setup

### Concurrency

- Strict sequential numbering enforced at repository level (`StreamSequenceError` on gaps, `StreamSlotConflictError` on duplicate numbers)
- Summary-log idempotency via partial unique index on `(registrationId, accreditationId, kind, payload.summaryLogId)`
- PRN idempotency via partial unique index on `(registrationId, accreditationId, kind, payload.prnId)`

## Tests

[...]

PR #1187, opened 2026-05-20
PAE-1382: Add stream repository layer

dansweeting:

## Summary

- Adds the new `waste-balance-events` stream repository: schema, port, in-memory and MongoDB adapters, contract tests, and five MongoDB indexes (slot uniqueness, two partial-unique idempotency indexes, kind+number, and PRN watermark)
- Purely additive: nothing imports these files yet. This is the foundation layer for the event-sourced waste balance stream (PR 1 of 6 splitting #1179)

## Test plan

- [x] 20 schema validation tests pass
- [x] 27 in-memory adapter tests pass (including full contract suite)
- [x] 34 MongoDB adapter tests pass (including index verification and error classification)
- [x] 100% coverage, no gaps
- [x] Type check clean


src/waste-balances/repository/stream-contract/deleteAllForPartition.contract.js
```diff
@@ -0,0 +1,109 @@
...
+      await repository.appendEvent(
+        buildStreamEvent({
+          registrationId: 'reg-target',
+          accreditationId: 'acc-target',
+          number: 2,
+          payload: { summaryLogId: 'log-2', creditTotal: 200 }
+        })
+      )
+
+      await repository.deleteAllForPartition('reg-target', 'acc-target')
```
crosslandwa: What's the use case for `deleteAllForPartition`?

I'm a bit wary of having _delete_ capable production code

If this is for facilitating the "cleanup org data" routine used in tests, can this code be lifted out and co-located there?

dansweeting: The existing non-prod data reset already deletes collection items directly via `deleteMany` in `buildCascadeSteps`, so having delete capability on the repository port is unnecessary indirection. 

Removed `deleteAllForPartition` from the port, both adapters, and the contract tests. The `waste-balance-events` collection will be added to the cascade steps when the non-prod reset gets updated in a later PR in this series.

src/waste-balances/repository/stream-contract/findLatestByPartition.contract.js
```diff
@@ -0,0 +1,163 @@
...
+          accreditationId: 'acc-y',
+          number: 1,
+          payload: { summaryLogId: 'log-y', creditTotal: 500 }
+        })
+      )
+
+      const x = await repository.findLatestByPartition('reg-x', 'acc-x')
+      const y = await repository.findLatestByPartition('reg-y', 'acc-y')
+
+      expect(x.number).toBe(1)
```
crosslandwa, line 100: Probably worth adding assertions for

```
  expect(x.accreditationId).toBe('acc-x')
  expect(y.accreditationId).toBe('acc-y')
```

to confirm x and y aren't the same event

src/waste-balances/repository/stream-contract/findLatestByPartitionAndKind.contract.js
```diff
@@ -0,0 +1,78 @@
...
+      const result = await repository.findLatestByPartitionAndKind(
+        'reg-empty',
+        'acc-empty',
+        STREAM_EVENT_KIND.SUMMARY_LOG_SUBMITTED
+      )
+
+      expect(result).toBeNull()
+    })
+  })
+}
```
crosslandwa, line 154: Probably worth adding tests for

`isolates results by partition`
`treats null and non-null accreditationId as separate streams`

as per the contract tests for `findLatestByPartition`

crosslandwa (approved):

src/waste-balances/repository/stream-contract/findLatestByPartition.contract.js
dansweeting: Added accreditationId assertions to confirm x and y are distinct events.

0288a5830b50fad395e3e762d7e44621f6e6a593

src/waste-balances/repository/stream-contract/findLatestByPartitionAndKind.contract.js
dansweeting: Added both: 'isolates results by partition' and 'treats null and non-null accreditationId as separate streams', mirroring the findLatestByPartition contract suite.

0288a5830b50fad395e3e762d7e44621f6e6a593

src/waste-balances/repository/stream-schema.js
```diff
@@ -0,0 +1,118 @@
...
+  accreditationId: Joi.string().allow(null).required(),
+  organisationId: Joi.string().required(),
+  number: Joi.number().integer().min(1).required(),
+  kind: Joi.string()
+    .valid(...kindValues)
+    .required(),
+  payload: Joi.when('kind', {
+    is: STREAM_EVENT_KIND.SUMMARY_LOG_SUBMITTED,
+    then: summaryLogPayloadSchema.required(),
+    otherwise: prnPayloadSchema.required()
```
senthilinfy: Instead of otherwise clause for PRN, should it not be on PRN kind values to be defensive?

src/waste-balances/repository/stream-port.js
```diff
@@ -0,0 +1,109 @@
...
+ * slot conflict — lives above the port in `appendToStream`, not on adapters.
+ */
+
+/**
+ * Raised by `appendEvent` when a `(registrationId, accreditationId, number)`
+ * slot is already occupied. Adapters translate their native conflict signal
+ * (MongoDB `E11000`, in-memory index check) into this typed error so callers
+ * can react uniformly.
+ */
+export class StreamSlotConflictError extends Error {
```
senthilinfy, line 16: It it better to call the stream as WasteBalanceStream instead of  using stream in a generic sense in file names and function names etc....

src/waste-balances/repository/stream-mongodb.js
```diff
@@ -0,0 +1,230 @@
...
+ * matching specifications.
+ *
+ * @param {Db} db
+ * @returns {Promise<Collection>}
+ */
+export async function ensureStreamCollection(db) {
+  const collection = db.collection(WASTE_BALANCE_EVENTS_COLLECTION_NAME)
+
+  await collection.createIndex(
+    { registrationId: 1, accreditationId: 1, number: 1 },
```
senthilinfy, line 37: I think UK constraint is needed separately on registrationId and accreditationId alongwith number. Can't have multiple ledger for same registration or accreditation

src/waste-balances/repository/stream-schema.js
dansweeting: Good point. Replaced the `otherwise` with explicit `Joi.valid()` matching against the four PRN kinds. Now if a new kind is added without updating the schema, it will fail validation rather than silently getting the PRN payload shape.

src/waste-balances/repository/stream-port.js
dansweeting: Renamed the typedefs to `WasteBalanceStreamRepository` and `WasteBalanceStreamRepositoryFactory` for clarity. Kept the filenames as `stream-*.js` since they already live under `src/waste-balances/repository/` which provides the domain context.

src/waste-balances/repository/stream-mongodb.js
dansweeting: Not sure I understand, the uniqueness will always be a combination of these three fields, this is the constraint for 'slot uniqueness'. We'll definitely have many events for a given reg/accred combo but these have to be unique across the dimension of event number - must advance in numeric sequence.

So I don't think we'll be able to purely constrain just on reg/accred - by design we'll have duplicates as soon as the second event in a stream is added?

PR #1190, opened 2026-05-20
PAE-1382: Add stream application primitives

dansweeting:

## Summary

- Adds `appendToStream` primitive: reads the latest event, computes closing balances per event kind, and appends a single event with sequential numbering
- Adds `performUpdateViaStream`: aggregates row-level target amounts into a single `creditTotal` per submission
- Purely additive: nothing calls these yet. This is PR 2 of 6 splitting #1179

## Test plan

- [x] 11 append-to-stream tests pass (all five event kinds plus slot conflict)
- [x] 7 update-via-stream tests pass
- [x] 100% coverage
- [x] Type check clean

### DIFF
--- a/src/waste-balances/application/append-to-stream.js
+++ b/src/waste-balances/application/append-to-stream.js
@@ -1,0 +1,160 @@
+import { add, subtract, toNumber } from '#common/helpers/decimal-utils.js'
+
+import { STREAM_EVENT_KIND } from '../repository/stream-schema.js'
+
+const ZERO_BALANCE = Object.freeze({ amount: 0, availableAmount: 0 })
+
+/**
+ * @param {import('../repository/stream-port.js').StreamEvent | null} latest
+ */
+const openingBalanceFrom = (latest) =>
+  latest === null ? { ...ZERO_BALANCE } : { ...latest.closingBalance }
+
+/**
+ * @param {import('../repository/stream-port.js').StreamEvent | null} latest
+ */
+const nextNumberFrom = (latest) => (latest === null ? 1 : latest.number + 1)
+
+/**
+ * Compute closing balance for a summary-log-submitted event.
+ *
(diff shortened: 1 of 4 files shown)


src/waste-balances/application/append-to-stream.js
```diff
@@ -0,0 +1,160 @@
...
+
+  let closingBalance
+
+  if (kind === STREAM_EVENT_KIND.SUMMARY_LOG_SUBMITTED) {
+    const { creditTotal } =
+      /** @type {import('../repository/stream-schema.js').SummaryLogSubmittedPayload} */ (
+        payload
+      )
+
+    const previousSubmission = await repository.findLatestByPartitionAndKind(
```
senthilinfy, line 123: should `findLatestByPartitionAndKind` not get latest as of ```number``` fetched in beginning. 

I understand this fine since we assume number remains same(no further transactions) but then fail during update if number has moved up. Just wanted to check

dansweeting, line 123: The latest event overall (fetched on line 107) might be a PRN event, not a summary-log event. We need the previous `creditTotal` to compute the delta, and only `summary-log-submitted` events carry that field. So we look back for the most recent event of that kind specifically.

e.g. if the stream is `[SL(creditTotal=100), PRN_CREATED, PRN_ISSUED]` and a new SL arrives with `creditTotal=150`:
- `latest` (L107) = PRN_ISSUED (for opening balance + next slot number)
- `previousSubmission` (L123) = the first SL (for `previousCreditTotal=100`)
- `delta = 150 - 100 = 50`

dansweeting: That's right, the sequence numbering is checked and enforced at the repo level.

senthilinfy (approved):

PR #1197, opened 2026-05-21
PAE-1428 Add basic auth strategy to GET /organisations endpoint

crosslandwa:
Ticket: [PAE-1428](https://eaflood.atlassian.net/browse/PAE-1428)
## Description

This enables us to expose this API to other services on CDP (by providing them with user/pass credentials, ie shared secrets).

This strategy compliments the existing 'access-token' strategy which (for this specific endpoint) expects an Entra ID JWT token

By default strategy will always return unauthorized

To enable the strategy, configure the server with credentials via the `BASIC_AUTH_USERNAME` and `BASIC_AUTH_PASSWORD` environment variables

---

[PAE-1428]: https://eaflood.atlassian.net/browse/PAE-1428?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/plugins/auth/basic-auth-plugin.js
```diff
@@ -0,0 +1,100 @@
...
+              ) {
+                return h.unauthenticated(Boom.unauthorized(null, 'Basic'))
+              }
+
+              const [username, password, ...rest] = Buffer.from(
+                authorization.slice(6),
+                'base64'
+              )
+                .toString('utf-8')
+                .split(':')
```
timrutter: claude spots:   RFC 7617 allows colons in the password (only the userid is colon-free). Splitting on : and rejecting
  rest.length > 0 makes any password with a colon unusable. Acceptable for a controlled shared secret,
  but worth a deliberate decision — the safer parse is "split on the first colon only":

  const idx = decoded.indexOf(':')

src/routes/v1/organisations/get.js
```diff
@@ -25,6 +24,7 @@ export const organisationsGetAll = {
   path: organisationsGetAllPath,
   options: {
     auth: {
+      strategies: ['access-token', BASIC_AUTH],
```
senthilinfy, line 27: Would it not be better to 

- Add a new scope called `registrationService` or sth like that 
- Have `BASIC_AUTH_USERNAME` and password specific to registration service
- When its basic auth from registration service give read access to get end point

This way later if we move  towards congnito auth the scope bit can remain as its?

crosslandwa: I think the trick here is mapping "who is the client" to "what scopes they have"

Basing the declaration of the endpoints on `scopes` means the endpoint doesn't care _who_ the user is, only what they can do (as dictated by their scopes)

A think an improvement would to be protect this endpoint by _two_ scopes
- `admin.read` (ie the existing scope)
- `organisation.read` (ie a new scope)

then update the `basic-auth-plugin` to give any caller that passes authorisation (which we _know_ will be the registration service, based on us having shared a user/pass with them*) the (new) `organisation.read` scope

(* in the future, if we had _more than one_ client who used basic auth we give each their own credentials, and update the plugin to be aware of a list of `user/pass` pairs, rather than just one pair)

src/plugins/auth/basic-auth-plugin.js
crosslandwa: Addressed

src/routes/v1/organisations/get.js
crosslandwa: This should be addressed now

senthilinfy (approved):

PR #1204, opened 2026-05-26
PAE-1382: Add monotonic lastAppliedEventNumber watermark to PRNs

GraemeF:

## What

This installs a repository-level guard that stops a PRN's event-stream watermark from regressing, and adds the field that guard protects. It is enforcement only — there is no business-logic or API change, and nothing populates a watermark yet, so the guard is dormant until the event-first work that follows starts setting one.

- PRN documents gain an optional integer `lastAppliedEventNumber` (minimum 1). `prnInsertSchema` and `prnReadSchema` accept it and it survives Joi's `stripUnknown`; the `PackagingRecyclingNote`, `UpdateStatusParams` and `RollbackParams` typedefs carry it.
- The guarantee the guard enforces: once a PRN carries a watermark it has migrated to event-stream status tracking, and from that point every write must carry it forward — it can never be dropped or rewound. A write may leave the watermark unchanged or advance it (a write needn't move it, only refrain from moving it backwards). A PRN that has never carried a watermark may omit it: that is the pre-migration path, and it is every PRN today.
- A regression here cannot be a lost race. A genuine race bumps the document version and already surfaces as a 409 version conflict, so the only way to drop or rewind a watermark is for the calling code to feed an out-of-order or absent value into the projection — a coding error. It is therefore rejected as a 500 internal error and logged as a `watermark_regression_detected` database event, with a message that distinguishes a dropped watermark from a backwards move.
- The guard runs on every write, not just status updates. The three rollback methods take the same optional watermark, threaded from the loaded PRN, and run through the same monotonic check. Both adapters route the rejection through one shared helper so the message, log event and status code stay identical — the MongoDB adapter folds the check into the same atomic compare-and-set as the optimistic version guard, and the in-memory adapter mirrors it.
- The watermark is internal projection state and never part of the contract with API clients: it is absent from the external PRN schema and the external read mapping, and the external accept/reject handlers read nothing from the request body but the timestamp, so a client can neither see it nor set it.
- The field is optional in storage, so the embedded (non-stream) write path, which never sets a watermark, is unaffected.
- Separately, the PRN repository's schema and contract test files now type-check cleanly: deliberately malformed validation inputs are expressed type-safely — required fields are dropped by rest-destructuring rather than mutated away with `delete`, and invalid overrides are built by spreading so there is no contextual type to violate.

## Why

This is the safety rail for the event-first PRN work that follows. A PRN projection needs to record how far through the waste-balance event stream it has been built, so the read path can later fold any events that landed after the document was last written. Once a PRN is tracked that way, a write that silently dropped or rewound its watermark would corrupt the projection. Installing the guard now — ahead of the change that starts setting watermarks — means the first time anyone introduces such a regression it fails loudly as a coding error rather than quietly as a data bug, and it covers rollbacks as well as forward transitions. `schemaVersion` stays at 2 because the field is additive and optional in storage.

### DIFF
--- a/src/packaging-recycling-notes/repository/mongodb.js
+++ b/src/packaging-recycling-notes/repository/mongodb.js
@@ -351,6 +351,11 @@ const performUpdateStatus = async (
   }
 
   const versionMatches = { $eq: [{ $ifNull: ['$version', 1] }, version] }
+  // A supplied watermark must not go backwards: it has to be >= the stored one,
+  // with a missing stored watermark defaulting to the incoming value so the
+  // first write always passes. An omitted watermark is allowed only while the
+  // PRN carries none (the embedded path); omitting it over a stored watermark
+  // is itself a regression and fails the guard.
   const watermarkAdvances =
     lastAppliedEventNumber === undefined
       ? { $eq: [{ $ifNull: ['$lastAppliedEventNumber', null] }, null] }
--- a/src/packaging-recycling-notes/repository/mongodb.js
+++ b/src/packaging-recycling-notes/repository/mongodb.js
@@ -8,6 +8,7 @@ import {
 import { PRN_STATUS } from '#packaging-recycling-notes/domain/model.js'
 import { PrnNumberConflictError } from './port.js'
 import { validatePrnInsert, validatePrnRead } from './validation.js'
+import { throwWatermarkRegression } from './watermark-guard.js'
 
 /** @import { Collection, Db, Document, Filter, WithId } from 'mongodb' */
 /** @import { Organisation } from '#domain/organisations/model.js' */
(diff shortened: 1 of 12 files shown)


src/packaging-recycling-notes/repository/schema.js
```diff
@@ -88,6 +88,7 @@ const statusSchema = Joi.object({
 export const prnInsertSchema = Joi.object({
   schemaVersion: Joi.number().integer().valid(2).required(),
   version: Joi.number().integer().min(1).default(1),
+  lastAppliedEventNumber: Joi.number().integer().min(1).optional(),
```
dansweeting, line 91: While this is an optional field in the schema, we will implicitly need clients to pass the value as soon as we migrate - that's going to force the external clients we have to deploy a change to accomodate.

Once a PRN has a watermark, every write must carry one.

We have a couple of external-facing endpoints that another team use:
```
POST /v1/packaging-recycling-notes/{prnNumber}/accept
POST /v1/packaging-recycling-notes/{prnNumber}/reject
```

If clients of those APIs dont make necessary changes to pass lastAppliedEventNumber, and a PRN has been migrated to watermark tracking, those endpoints will 500. 

The current PR threads the field through performTransition in update-status.js, so internal callers are covered. But accept/reject have a separate path that doesn't thread it.

There's likely to be ways we can 'hide' the breaking change from external clients but it's going to involve looking up the existing watermark and blindly reapplying that value or incrementing it to satisfy the rule, so we'd have to weigh that up

GraemeF: The watermark is entirely internal, clients should not be aware of it. Let me check we've exercised that in tests...

GraemeF: Right, so yes this will break when we migrate balances over, but so will everything else until the rest of the changes are finished. It's still WIP. I don't think it's possible to write those tests until the next PR or 2 land, because the behaviour isn't implemented.

dansweeting: Sure, the main point to raise though is the concern that _after_ the WIP is complete will we have forced another team to update and re-release software to accommodate? That's a tradeoff that we missed from the design of this

GraemeF: No, it's internal. The routes will do the Right Thing.

src/packaging-recycling-notes/repository/mongodb.js
```diff
@@ -250,39 +251,81 @@ const performFindByStatus = (db, excludeOrganisationIds) => {
...
+ * and it must be greater than or equal to the stored value. A missing stored
+ * value defaults to the incoming one so the migrating write always passes.
+ *
+ * @param {number | undefined} lastAppliedEventNumber
+ */
+const watermarkNotRegressing = (lastAppliedEventNumber) =>
+  lastAppliedEventNumber === undefined
+    ? { $eq: [{ $ifNull: ['$lastAppliedEventNumber', null] }, null] }
+    : {
+        $gte: [
```
dansweeting, line 266: I think I've missed why its legal for the value of the watermark to stay un-incremented. 

Presumably we only enforce the CAS during a mutation of data, so are there mutations we allow that don't yield a new event with a higher number?

GraemeF: The watermark is the projection's position in the waste-balance event stream — the highest stream event folded into this PRN — not a per-write counter. It only moves when a write folds a new event.

Plenty of mutations don't produce a stream event: accept, reject and discard change PRN status but touch no waste balance (no branch for them in `applyWasteBalanceEffects`). A write like that re-persists the projection without folding anything new, so it carries the existing watermark forward unchanged. The rollback path is the same — it re-applies the watermark it loaded.

So the rule is monotonic non-decreasing, not strictly increasing. The guard's only job is to reject a write that moves the position backwards (or drops it) — which can't be a lost race (that's the 409) and can only be stale state fed into the projection. Equal is the normal case for any write that mutates the document without advancing the stream.

dansweeting (approved):

PR #1208, opened 2026-05-27
PAE-1382: Fix summary log ID namespace mismatch in stream diagnostic

dansweeting:
## Summary

The divergence diagnostic passed the summary log document `_id` to `computeRebuiltStream`, but waste record versions store `file.id` (`summaryLog.file.id`). These are two different UUIDs, so `reconstructDataAtSubmission` never matched any versions, every `summary-log-submitted` event had `creditTotal: 0`, and the stream replay showed incorrect balances for every accreditation.

- Pass `summaryLog.file.id` instead of `doc._id` in the mapping
- Extract `toStreamSummaryLog` helper so the integration test can exercise the production mapping against a real `findAllByOrgReg` result
- Add integration test that seeds a summary log through the real MongoDB repository, maps via the exported helper, and verifies `computeRebuiltStream` produces the correct `creditTotal` (fails with `amount: 0` under the old mapping)


GraemeF:
The fix itself looks right: persisted waste-record versions store `summaryLog.file.id` (set in `sync-from-summary-log.js`), which is a different identifier from the document `_id` — `submit.js` logging confirms the two diverge — so correlating on `file.id` is correct.

My concern is that the integration test doesn't actually exercise the id seam it's meant to protect. The summary-log side is genuinely real (insert → `findAllByOrgReg` → `toStreamSummaryLog`), but the waste-record side is hand-built with `summaryLog: { id: fileId }` (line 97) — the same local variable passed to the factory's `file: { id: fileId }`. The only thing correlating the waste record with the summary log is that both use the `fileId` literal; nothing derives the version's `summaryLog.id` from a real submission.

Concretely: if you change line 97 from `fileId` to `docId`, the test fails with `amount: 0`; as written it passes. So the outcome is decided by which literal is typed on the waste-record side, not by what the production write path stores. If the assumption had instead been that versions store `_id`, the same test shape would be green against the opposite (wrong) mapping. It verifies internal consistency between two hand-picked values, not that `file.id` is the correct join key — that fact currently lives only in the `// this is what sync-from-summary-log.js writes` comment.

What would close the gap: persist the waste record through the real path too — build the summary log with `file.id`, run `transformFromSummaryLog`/`syncFromSummaryLog` so it sets the version's `summaryLog.id`, persist via the real waste-records repo, read it back via `findByRegistration`, then run the diagnostic. Then a single id flows summary-log doc → version → mapping → match with no hardcoded join key on either side, and changing either writer breaks the test.

Separately, a smaller thing I noticed: `toStreamSummaryLog` dereferences `summaryLog.file.id` for every doc `findAllByOrgReg` returns, which includes failure-status logs (rejected/invalid/…), not just submitted ones. `file` is required on non-preprocessing inserts but optional on updates (`schema.js`), and `preprocessing → rejected` is a legal transition, so a file-less failure-status doc is possible — that would throw and get counted as a rebuild failure for that accreditation. Since only submitted logs contribute credits, mapping `summaryLog.file?.id` or filtering to submitted before mapping would avoid it.

GraemeF (approved):

PR #1209, opened 2026-05-27
PAE-1382: Write PRN balance events first on the ledger path

GraemeF:
When an accreditation's waste balance has migrated to the event-sourced stream (`canonicalSource === 'ledger'`), balance-affecting PRN status transitions are now written event-first: the balance event is appended to the stream *before* the PRN document is projected, and the document is stamped with the appended event number as a monotonic watermark (`lastAppliedEventNumber`). There is no compensation on this path — a partial failure where the event commits but the document write fails is left in place, to be reconciled by the read-side catch-up that folds the stream forward on the next read, rather than rolled back. This is the inverse of the embedded path, where the document is the source of truth and a failed balance write is compensated.

The write strategy is chosen up front from the accreditation's canonical-source marker, and only for transitions that actually move the balance (create, issue, delete-from-awaiting-authorisation, cancel-an-issued-PRN). Lifecycle-only transitions — and every transition on a non-migrated accreditation — carry the PRN's existing watermark forward unchanged, which keeps the repository's monotonic watermark guard satisfied without a marker read. On issuance the suspended-accreditation check is hoisted ahead of the append, so a suspended accreditation is rejected before any balance event is written.

The embedded (non-migrated) create and transition paths, including their credit-back and full-credit compensation, are unchanged. Every production accreditation is still `embedded` today, so the ledger path added here is dormant until balances are migrated; the change is exercised end-to-end by the waste-balance arithmetic integration tests under a seeded ledger-backed accreditation, confirming the PRN events fold into the resolved balance and the stamped watermark tracks the latest stream event.

Part of PAE-1382 — moving the waste balance from the embedded document onto the event-sourced stream.

### DIFF
--- a/src/packaging-recycling-notes/application/update-status-balance-effects.js
+++ b/src/packaging-recycling-notes/application/update-status-balance-effects.js
@@ -196,13 +196,21 @@ function logWasteBalanceUpdate(
  * @returns {boolean}
  */
 export function affectsWasteBalance(currentStatus, newStatus) {
+  const deductsAvailableOnCreation =
+    newStatus === PRN_STATUS.AWAITING_AUTHORISATION
+  const deductsTotalOnIssue = newStatus === PRN_STATUS.AWAITING_ACCEPTANCE
+  const creditsAvailableOnPreIssueCancellation =
+    (newStatus === PRN_STATUS.CANCELLED || newStatus === PRN_STATUS.DELETED) &&
+    currentStatus === PRN_STATUS.AWAITING_AUTHORISATION
+  const creditsFullOnIssuedCancellation =
+    newStatus === PRN_STATUS.CANCELLED &&
+    currentStatus === PRN_STATUS.AWAITING_CANCELLATION
+
   return (
-    newStatus === PRN_STATUS.AWAITING_AUTHORISATION ||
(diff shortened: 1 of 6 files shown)


src/packaging-recycling-notes/application/update-status-balance-effects.js
```diff
@@ -184,6 +184,36 @@ function logWasteBalanceUpdate(
   })
 }
 
+/**
+ * Whether a status transition fires a waste balance effect. The four cases
+ * mirror the firing branches in {@link applyWasteBalanceEffects} exactly: a
+ * caller that needs to know whether a balance write will happen (to choose the
+ * write ordering before any side effect runs) must agree with what the effects
+ * actually do.
```
dansweeting: Risk of drift between this and the linked function? If the two have to agree, can we manage the coupling in any safer way?

GraemeF: Good catch — the predicate is gone in c94511f8. `balanceEventsFor` now returns the balance events the transition would write, and the call site uses `events.length > 0` as the gate. One source of truth for "does this affect the balance?" and "what work to do", so there's nothing left to drift between. Event kinds align with `STREAM_EVENT_KIND` so the live path now speaks the same vocabulary as the stream itself.

dansweeting (approved):

PR #1211, opened 2026-05-28
PAE-1382: Unify PRN status projection across the ledger write and read paths

GraemeF:

## Summary

Makes a single pure fold the canonical projection of PRN status for accreditations whose waste balance has migrated to the event stream (`canonicalSource: 'ledger'`). The same `foldPrnFromTailEvents` drives both the write and read paths, so the two can no longer encode the event-kind→status mapping differently and silently diverge.

**Write path** — A ledger-marker status transition is written event-first: compute the balance-affecting stream events, fold them onto the PRN, then persist the resulting projection through a new `prnRepository.persistProjection` (optimistic concurrency on version). There is no compensation — a partial failure (events appended, doc not persisted) self-heals on the next read. Embedded-marker accreditations keep their existing field-level write path until they migrate.

**Read path** — `GET .../packaging-recycling-notes/{prnId}` returns the PRN folded forward over any stream events past its `lastAppliedEventNumber`, so a partial write failure surfaces as the correct current state. This sits behind a single `getProjectedPrnById` application function: routes ask for the fully-formed PRN and never touch the event stream themselves. A missing or soft-deleted document short-circuits with no stream query.

**Repository** — `persistProjection` on the PRN repository, and `getPrnCatchupEvents` / `appendStreamEvent` on the waste-balances repository, each across the MongoDB and in-memory adapters and covered by shared contract tests.

The gate is the marker, not the feature flag — once an accreditation is on the ledger, its truth lives in the events and both reads and writes go through the fold.

## Rollout

Dormant in production: `FEATURE_FLAG_WASTE_BALANCE_LEDGER` defaults false and every existing waste-balance doc has `canonicalSource: 'embedded'`, so the new code paths run for nobody until the cutover.

### DIFF
--- a/src/packaging-recycling-notes/routes/get-by-id.js
+++ b/src/packaging-recycling-notes/routes/get-by-id.js
@@ -74,8 +74,7 @@ export const packagingRecyclingNoteById = {
       ])
 
       if (
-        !prn ||
-        prn.organisation.id !== organisationId ||
+        prn?.organisation.id !== organisationId ||
         prn.accreditation.id !== accreditationId ||
         prn.status.currentStatus === PRN_STATUS.DELETED
       ) {
--- a/src/packaging-recycling-notes/routes/get-by-id.js
+++ b/src/packaging-recycling-notes/routes/get-by-id.js
@@ -77,8 +77,12 @@ export const packagingRecyclingNoteById = {
         )
       ])
 
+      if (!prn) {
+        throw Boom.notFound('PRN not found')
+      }
+
       if (
-        prn?.organisation.id !== organisationId ||
+        prn.organisation.id !== organisationId ||
         prn.accreditation.id !== accreditationId ||
         prn.status.currentStatus === PRN_STATUS.DELETED
       ) {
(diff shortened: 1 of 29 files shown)


src/packaging-recycling-notes/routes/get-by-id.js
```diff
@@ -49,33 +51,36 @@ export const packagingRecyclingNoteById = {
...
       ])
 
-      // Verify the PRN exists, belongs to the requested organisation/accreditation,
-      // and treat deleted PRNs as not found (soft delete)
       if (
         prn?.organisation.id !== organisationId ||
-        prn?.accreditation.id !== accreditationId ||
-        prn?.status.currentStatus === PRN_STATUS.DELETED
+        prn.accreditation.id !== accreditationId ||
+        prn.status.currentStatus === PRN_STATUS.DELETED
```
dansweeting, line 87: can we safely lose the `prn?.` from these lines? Line 81 keeps it making we wonder if it's still possible to be undefined.

GraemeF: Yes, it's possible it might be undefined, in which case it will `Boom.notFound`. It's not as obvious as it could be because that made Sonar unhappy.

GraemeF: image

dansweeting (approved):

PR #1215, opened 2026-05-28
PAE-1382: Add stream promotion orchestrator

dansweeting:
## Summary

- Add startup sweep that migrates each embedded accreditation to the event-sourced stream behind `FEATURE_FLAG_WASTE_BALANCE_LEDGER`
- Add `deleteByPartition` and `bulkAppendEvents` migration-only methods to the stream repository (both MongoDB and in-memory adapters, with contract tests)
- Relax `marker-aware-read` invariant so an empty stream under a `ledger` marker returns zero balances instead of throwing (correct for accreditations with no waste records)

## How it works

For each embedded accreditation the orchestrator:
1. Reconstructs the full event history via `computeRebuiltStream`
2. Flips the canonical source marker to `migrating` (version-gated)
3. Deletes any existing stream events (idempotency on restart)
4. Bulk-inserts the rebuilt events
5. Flips the marker to `ledger`

Recovery pass on startup resets stuck `migrating` accreditations back to `embedded` before the main pass begins. Runs under a distributed lock so only one pod per deploy executes the sweep.

## Tests

- [ ] Contract tests for `deleteByPartition` (3 tests: deletes partition, returns 0 on empty, isolates other partitions)
- [ ] Contract tests for `bulkAppendEvents` (5 tests: happy path, sequence errors, empty array, visibility)
- [ ] Both contracts verified against MongoDB and in-memory adapters
- [ ] `marker-aware-read` test updated for zero-balance return
- [ ] `findByAccreditationId` contract test updated for zero-balance return
- [ ] Orchestrator integration tests (13 tests: feature flag off, lock contention, recovery pass, full lifecycle, skip already-ledger, empty accreditation, error branches, summary logging)
- [ ] 100% code coverage maintained

### DIFF
--- a/src/server/run-stream-promotion.js
+++ b/src/server/run-stream-promotion.js
@@ -1,6 +1,5 @@
 import { logger } from '#common/helpers/logging/logger.js'
 import { getConfig } from '#root/config.js'
-import { resolveOverseasSites } from '#application/waste-records/resolve-overseas-sites.js'
 import { createOrganisationsRepository } from '#repositories/organisations/mongodb.js'
 import { createOverseasSitesRepository } from '#overseas-sites/repository/mongodb.js'
 import { createPackagingRecyclingNotesRepository } from '#packaging-recycling-notes/repository/mongodb.js'
@@ -9,17 +8,8 @@ import { createWasteRecordsRepository } from '#repositories/waste-records/mongod
 import { createWasteBalancesRepository } from '#waste-balances/repository/mongodb.js'
 import { createMongoStreamRepository } from '#waste-balances/repository/stream-mongodb.js'
 import { computeRebuiltStream } from '#waste-balances/application/compute-rebuilt-stream.js'
-import { toStreamSummaryLog } from '#server/run-balance-divergence-diagnostic.js'
+import { loadAccreditationSources } from '#server/load-accreditation-sources.js'
 import { WASTE_BALANCE_CANONICAL_SOURCE } from '#waste-balances/domain/model.js'
-import { REG_ACC_STATUS } from '#domain/organisations/model.js'
-import { SUMMARY_LOG_STATUS } from '#domain/summary-logs/status.js'
-
-/** @type {Set<import('#domain/organisations/registration.js').Registration['status']>} */
-const ACTIVE_REGISTRATION_STATUSES = new Set([
-  REG_ACC_STATUS.APPROVED,
-  REG_ACC_STATUS.CANCELLED,
-  REG_ACC_STATUS.SUSPENDED
-])
 
 const WASTE_BALANCES_COLLECTION = 'waste-balances'
 const LOCK_NAME = 'stream-promotion'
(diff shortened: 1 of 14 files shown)


src/server/run-stream-promotion.js
```diff
@@ -0,0 +1,348 @@
...
+          organisationId: 1,
+          registrationId: 1
+        }
+      }
+    )
+    .toArray()
+  return /** @type {{ accreditationId: string, organisationId: string, registrationId: string }[]} */ (
+    /** @type {unknown} */ (docs)
+  )
+}
```
dansweeting, line 59: These raw `db.collection()` queries bypass the waste-balances port, sidestepping ports/adapters. 

It's a match with the divergence diagnostic (`findEmbeddedWasteBalances` in `run-balance-divergence-diagnostic.js`), and the justification is the same: this is temporary migration code.

src/server/run-stream-promotion.js
```diff
@@ -0,0 +1,348 @@
...
+    return 'skipped'
+  }
+
+  await streamRepository.deleteByPartition(registration.id, row.accreditationId)
+  await streamRepository.bulkAppendEvents(events)
+
+  const afterStream = await wasteBalancesRepository.findByAccreditationId(
+    row.accreditationId
+  )
+  if (!afterStream) {
```
dansweeting: Discussion point: this delete-then-reinsert makes restarts idempotent (a crash between `bulkAppendEvents` and `flipCanonicalSourceToLedger` leaves stale events that would cause a sequence conflict on retry). The trade-off is that we unconditionally delete even on a first run where nothing exists yet.

Alternatives worth considering:
- **Skip the delete entirely** and rely on `bulkAppendEvents` sequence validation to detect a partial write, then reset to embedded and retry next boot. Avoids the delete but makes recovery more complex.
- **Check whether events already exist** before deleting (e.g. `findLatestByPartition` returns non-null). Avoids the no-op delete on first run but adds a read.

The current approach is the simplest and matches the design doc. Worth a quick sanity check that the team is comfortable with the unconditional delete.

GraemeF (commented):
Mechanics mostly right (delete-before-insert, recovery pass, pre-migration captured version). One correctness bug to fix before merge, plus two safety regressions that compound it.

**Blocking — `run-stream-promotion.js:162-166`.** Version is captured *after* `rebuildEvents` reads the sources. A submission that writes its waste records and bumps `version` entirely within the [source-read, version-capture] window lands in `captured.version` but not in `events`; both flips then match and the ledger goes authoritative missing that submission. Capture the version *before* `rebuildEvents` so any write during the rebuild bumps version past the capture and the flip safely no-ops.

**Should-fix — `marker-aware-read.js:30`.** Relaxing throw → silent zero can't distinguish a legitimately-empty accreditation from a promotion that flipped to `ledger` without writing events; the latter now surfaces as a silent zero balance instead of erroring. Consider a zero-amount genesis event so the stream is never empty under a `ledger` marker, preserving the invariant.

**Should-fix — `run-stream-promotion.js:296,302`.** Per-accreditation failures and the summary log at `info`. Error-log alerting won't see a sweep that fails to migrate N accreditations, and unlike the read-only diagnostics this mirrors, this one mutates data and gates the cutover. Log `error`/`warn` when `failed > 0`.

**Medium — `run-stream-promotion.js:330`.** LockManager is constructed with no expiry and the lock is held for the whole sweep. A pod dying before `lock.free()` leaves a permanent lock, so every later boot silently skips promotion and the rollout stalls with no error. Confirm stale-lock handling / unlock runbook.

**Medium — `run-stream-promotion.js:84-152`.** `rebuildEvents` + `ACTIVE_REGISTRATION_STATUSES` duplicate ~50 lines of run-balance-divergence-diagnostic (already imported from for `toStreamSummaryLog`). Extract a shared rebuild helper.

**Minor — `run-stream-promotion.js:113`.** Accreditations whose only registration is `created`/`rejected` throw → failed → retried-and-failed every boot, and are orphaned when the embedded path is retired. Matches the divergence diagnostic, so gate the flip on a clean divergence run.

src/server/run-stream-promotion.js
```diff
@@ -0,0 +1,349 @@
...
+  )
+
+  const { events } = computeRebuiltStream({
+    accreditation,
+    wasteRecords,
+    prns,
+    overseasSites,
+    summaryLogs: summaryLogDocs
+      .filter(
+        ({ summaryLog }) => summaryLog.status === SUMMARY_LOG_STATUS.SUBMITTED
```
senthilinfy: Its not an issue but the filtering is redundant

```
const buildChronologicalEvents = ({
  accreditation,
  wasteRecords,
  prns,
  overseasSites,
  summaryLogs
}) => {
  const events = []

  const submitted = summaryLogs
    .filter(
      /** @returns {sl is { id: string, status: string, submittedAt: string }} */
      (sl) => sl.status === SUMMARY_LOG_STATUS.SUBMITTED
    )
    .sort(
      (a, b) =>
        new Date(a.submittedAt).getTime() - new Date(b.submittedAt).getTime()
    )```

File: src/server/run-stream-promotion.js [hunk 4]
```diff
@@ -0,0 +1,349 @@
...
+    return 'skipped'
+  }
+
+  // Idempotency: if a previous boot crashed between bulkAppendEvents and
+  // flipCanonicalSourceToLedger, stale events may exist. Deleting first
+  // means a restart always replays from the authoritative sources rather
+  // than appending on top of a partial write. An alternative would be to
+  // skip the delete and let bulkAppendEvents fail on a sequence conflict,
+  // but that turns a recoverable restart into a stuck accreditation.
+  await streamRepository.deleteByPartition(registration.id, row.accreditationId)
```
senthilinfy: It might be clean not to have deleteByPartition in the port but outside it. As its temporary code unless I am mistaken and keeps the stream with append only ops.

src/server/run-stream-promotion.js
dansweeting: Cheers, I'll double check but Im pretty sure the filtering is there purely to rule out failed or superseded summary logs, for a couple of reasons
- summary log has not affected WB unless it was submitted
- we can assume summary `log.file.ids` and `submittedAt` are present on the document in this state.

src/server/run-stream-promotion.js
```diff
@@ -0,0 +1,349 @@
...
+    await wasteBalancesRepository.flipCanonicalSourceToMigrating({
+      accreditationId: row.accreditationId,
+      capturedVersion: captured.version
+    })
+
+  if (
+    migratingResult?.canonicalSource !==
+    WASTE_BALANCE_CANONICAL_SOURCE.MIGRATING
+  ) {
+    return 'skipped'
```
senthilinfy, line 144: Should this not be error? As previous line is flipping it into migrating status

src/server/run-stream-promotion.js
GraemeF: Yeah no events are generated for summary logs that are not in a SUBMITTED state (it's a doc that we use to track an upload through to submission)

src/server/run-stream-promotion.js
dansweeting: Fair point on keeping the port append-only in principle. The reason it lives on the port is that it needs the collection/partition wiring and gets contract-tested against both the MongoDB and in-memory adapters, which catches adapter drift. Moving it to a raw `db.collection()` call would lose that coverage.

Since it's only called from the migration sweep behind the feature flag, removal is already scoped into the feature flag cleanup: when the flag goes, the method, its contract tests, and the call site all get deleted together.

src/server/run-stream-promotion.js
dansweeting: The skip is intentional. The flip uses optimistic concurrency (capturedVersion), so it won't land if the version was bumped between the read and the flip. That happens when a concurrent mutation (PRN op, summary log upload) touches the accreditation while we're mid-sweep.

Skipping is safe because the accreditation stays as embedded and gets retried on the next boot with fresh data. Throwing an error would halt the sweep for a condition that's expected and self-healing.

senthilinfy (approved):

PR #1233, opened 2026-06-01
PAE-1382: Persist registrationId on promotion to ledger

dansweeting:

## Summary

When the stream promotion sweep flips an accreditation from `embedded` to `ledger`, it now writes `registrationId` to the waste-balance document. This ensures the ledger read path (`resolveBalanceAmounts`) can look up stream events by partition, since `findLatestByPartition` requires `registrationId`.

### Root cause

`saveBalance` never persisted `registrationId` to the `waste-balances` collection, so all existing balance documents lack it. When the read path casts `balance.registrationId` to string, `undefined` becomes `"undefined"`, the stream query matches nothing, and the balance falls through to zero.

### Why fix at the migration layer, not the read layer

The promotion sweep already resolves each accreditation's registration via `loadAccreditationSources`. Writing `registrationId` during the `flipCanonicalSourceToLedger` step fixes the data at source. This avoids teaching `findLatestByPartition` to tolerate `undefined` and keeps the stream query contract clean.

`saveBalance` does not need updating because once an accreditation is on `ledger`, the write path uses event appends, not `saveBalance`.

### What this PR does

- `flipCanonicalSourceToLedger` accepts an optional `registrationId` and writes it to the balance document via `$set` alongside `canonicalSource`. When `registrationId` is `undefined` (e.g. no active registration), it is omitted from the update.
- `promoteAccreditation` passes `registration.id` through to the flip.
- Port type updated to include `registrationId` on `FlipCanonicalSourceToLedgerParams`.
- Test assertions updated to verify `registrationId` is passed through in all promotion scenarios.

### DIFF
--- a/src/waste-balances/repository/port.js
+++ b/src/waste-balances/repository/port.js
@@ -76,6 +76,10 @@
 /**
  * @typedef {Object} FlipCanonicalSourceToLedgerParams
  * @property {string} accreditationId
+ * @property {string} [registrationId] - Resolved during the promotion sweep
+ *   from the organisation's registration. Persisted to the balance document
+ *   so the ledger read path can look up stream events by partition. Not
+ *   required when the caller is not the promotion sweep (e.g. tests).
  * @property {number} capturedVersion - The `version` observed on the
  *   `'migrating'` doc at the start of step 3 of a per-accreditation rebuild
  *   (after authoritative history has been replayed into the ledger). The flip
--- a/src/waste-balances/repository/port.js
+++ b/src/waste-balances/repository/port.js
@@ -76,10 +76,9 @@
 /**
  * @typedef {Object} FlipCanonicalSourceToLedgerParams
  * @property {string} accreditationId
- * @property {string} [registrationId] - Resolved during the promotion sweep
+ * @property {string} registrationId - Resolved during the promotion sweep
  *   from the organisation's registration. Persisted to the balance document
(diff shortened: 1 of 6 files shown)


src/waste-balances/repository/port.js
```diff
@@ -76,6 +76,10 @@
 /**
  * @typedef {Object} FlipCanonicalSourceToLedgerParams
  * @property {string} accreditationId
+ * @property {string} [registrationId] - Resolved during the promotion sweep
```
GraemeF: Can we make it required?

dansweeting: It's optional because of the no-registration promotion path: when an accreditation has no active registration (zero balance, zero activity), `rebuildResult` is null so `registrationId` is `undefined`. The flip still needs to run to promote the accreditation to ledger. The `$set` spread skips `undefined`, and the read path hits `ZERO_BALANCE` anyway so `registrationId` is never needed for stream lookup.

Having said that I'm not sure how we'd ever get an accreditation without an active registration?

dansweeting: Good call. Made it required in the latest push. The no-registration path now throws rather than silently promoting without a \`registrationId\`, which is cleaner: if there's no active registration, that's broken data worth surfacing rather than papering over.

GraemeF (approved):

PR #1253, opened 2026-06-03
PAE-1240: Prevent report submission when a newer summary log is uploaded

senthilinfy:
Ticket: [PAE-1240](https://eaflood.atlassian.net/browse/PAE-1240)
## Summary

When a new Summary Log (SL) is submitted for an org/registration, any in-progress or ready-to-submit reports are derived from stale data. This PR blocks those reports immediately at SL submission time.

**Options considered:**
  **- Option A (chosen)** — Proactively mark reports stale at SL submission. Blocks all report endpoints (GET/PATCH/POST) the moment the new SL lands, preventing any further edits. Matches the intended UX where the error page appears on navigation, not just at final submit.

  **- Option B (rejected)** —   Lazily evaluate staleness on each request(list all reports, get, patch, status) by comparing the report's source.summaryLogId against the current latest summary log at that point in time. This approach can gate views and edits as well as submissions. However, whilst the staleness check is currently straightforward — a simple ID comparison — it is likely to grow more complex over time, for example comparing individual waste record fields or handling closed-month adjustments, making on-the-fly evaluation increasingly expensive. It also leaves no persistent audit trail recording when staleness was first detected.
  
**Implementation approach:** 
   A new `stale` field (not a status change) avoids touching the status machine, `STATUS_TO_SLOT`, or `isValidReportTransition`. All three report endpoints throw 409 with `code: summary_log_changed` for stale reports.

image

[PAE-1240]: https://eaflood.atlassian.net/browse/PAE-1240?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/reports/application/summary-log-events.js
```diff
@@ -0,0 +1,52 @@
...
+ *   systemLogsRepository: import('#repositories/system-logs/port.js').SystemLogsRepository
+ * }} params
+ * @returns {Promise<void>}
+ */
+export async function onSummaryLogUploaded({
+  organisationId,
+  registrationId,
+  reportsRepository,
+  systemLogsRepository
+}) {
```
dansweeting, line 24: A question on the payload for this onSummaryLogUploaded callback: despite being a handler for 'summaryLogUploaded' we are not being told anything about the particular summary log in the event arguments - just orgId, regId. I can see why we need those arguments but they're tangential to summary logs.

Wonder if there's even more useful info we could pass when marking staleness. We know which reports should trigger the staleness rules, and we set a useful timestamp and reason for when that happened.

Could we also capture which summary log triggered the staleness?

Not something we would necessarily surface to the user but as an audit trail it'd be handy. Not impossible to figure out without it, given a visible timeline of SL activity though.

src/reports/application/summary-log-events.js
```diff
@@ -0,0 +1,52 @@
...
+  const reports = await reportsRepository.findReportsByStatus(
+    organisationId,
+    registrationId,
+    [REPORT_STATUS.IN_PROGRESS, REPORT_STATUS.READY_TO_SUBMIT]
+  )
+
+  const now = new Date().toISOString()
+  const stale = { at: now, reason: STALE_REASON.SUMMARY_LOG_CHANGED }
+
+  for (const report of reports) {
```
dansweeting: A lot of chained calls here and I'm not sure the knock on effects of partial failure.
I could see a single repo function collapsing the sequence and whole loop of calls into a single mongo operation though.

something a bit like 

```
markReportsStaleByStatus(
    organisationId,
    registrationId,
    [REPORT_STATUS.IN_PROGRESS, REPORT_STATUS.READY_TO_SUBMIT],
    stale
)
```

Which would look more like bulk mongo update with a where clause, which I think is going to be atomic.
Calls into question the following `audit' call too, but there's a separate comment.

src/reports/application/summary-log-events.js
senthilinfy: Good point. I will capture SL id, this was on my mind while planning in CC but forgot during the implementation

src/reports/application/summary-log-events.js
senthilinfy: I will markStaleReports  atomic using below code and use insertMany for system logs.  

Also move marking reports as stale before SL status is marked to SUBMITTED should retry partial failures or alert worst case errors.

```
const filter = {
   organisationId,
   registrationId,
   'status.currentStatus': { $in: statuses },
   'stale.summaryLogId': { $ne: stale.summaryLogId },
   'source.summaryLogId': { $ne: stale.summaryLogId }
 }

 const previousDocs = await reportsCollection(db).find(filter, { projection: { _id: 0 } }).toArray()
 if (previousDocs.length === 0) return []

 await reportsCollection(db).updateMany(filter, { $set: { stale }, $inc: { version: 1 } })

 return previousDocs.map(p => ({ previous: p, updated: { ...p, stale, version: p.version + 1 } }))
```

dansweeting (approved):

PR #1262, opened 2026-06-03
PAE-1541: Classify loads by reporting period status

dansweeting:
## Summary

- During summary log validation, classifies each waste record into its reporting period and determines whether that period is open or closed (i.e. whether a report has already been submitted for it)
- Computes `loadsByPeriodStatus` aggregates: counts and tonnages bucketed by open/closed, added/adjusted, and included/excluded
- Persists `loadsByPeriodStatus` alongside existing `loads` on the summary log document and returns it via the GET endpoint

## Approach

Each table schema now declares a `reportingDateField` property identifying which field holds the date used for period assignment. A cross-schema drift-guard test ensures every schema has this property and that it appears in `requiredHeaders`.

Period status is determined by looking up submitted reports for the organisation/registration via the reports repository. A period is closed if any report has been submitted for it (either the current report is in submitted status, or there are previous submissions).

For tonnage, added records contribute their full `transactionAmount`. Adjusted records contribute the delta (new amount minus old amount), reflecting the actual change to the waste balance rather than the absolute tonnage.

The classification runs only for successfully validated summary logs. If the reports lookup fails, it degrades gracefully with a warning log and omits `loadsByPeriodStatus` from the document.

## Changes

- **New**: `period-status.js` with `classifyByPeriodStatus` function and unit tests
- **New**: `reportingDateField` on all 13 table schemas, plus drift-guard test
- **New**: `loadsByPeriodStatusSchema` Joi schema for storage and response validation
- **Modified**: `validate.js` to accept `reportsRepository`, fetch submitted reports, build transaction amounts (with delta for adjusted records), and compute period status
- **Modified**: Queue consumer plugin and command handler to wire `reportsRepository` through
- **Modified**: Summary log update schema, GET handler, and response schema to persist and return `loadsByPeriodStatus`

[PAE-1541]: https://eaflood.atlassian.net/browse/PAE-1541?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/domain/summary-logs/table-schemas/index.js
+++ b/src/domain/summary-logs/table-schemas/index.js
@@ -57,6 +57,7 @@ export const MIN_TEMPLATE_VERSIONS = {
 /**
  * @typedef {Object} TableSchema
  * @property {string} rowIdField - Field name containing the row identifier
+ * @property {string} reportingDateField - Field name containing the date used for reporting period classification
  * @property {string} wasteRecordType - The waste record type this table maps to (e.g. 'received', 'exported')
  * @property {string} sheetName - The spreadsheet sheet name for this table (e.g. 'Received', 'Exported')
  * @property {(rowData: Record<string, any>, rowIndex: number) => {wasteRecordType: string, rowId: string, data: Record<string, any>}} rowTransformer - Function to transform a parsed row into waste record metadata
--- a/src/application/summary-logs/period-status.js
+++ b/src/application/summary-logs/period-status.js
@@ -161,8 +161,11 @@ export const classifyByPeriodStatus = ({
 
   const closedPeriods = buildClosedPeriods(submittedReports, cadence)
 
-  const wasteBalanceRowIds = new Set(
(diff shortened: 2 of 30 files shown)


src/domain/summary-logs/table-schemas/index.js
```diff
@@ -57,6 +57,7 @@ export const MIN_TEMPLATE_VERSIONS = {
 /**
  * @typedef {Object} TableSchema
  * @property {string} rowIdField - Field name containing the row identifier
+ * @property {string} reportingDateField - Field name containing the date used for reporting period classification
```
dansweeting, line 60: I think if we're going to call out this field it makes sense that the existing reporting logic throughout the system relies on the same field

This would give me confidence that 
- everything agrees about which loads belong in which period
- we are protected from any accidental drift

PR #1264, opened 2026-06-04
PAE-1382: Attribute live RPD PRN transitions to the RPD service

GraemeF:

A live, RPD-initiated PRN status transition (accept or reject) now attributes its waste-balance stream event to the RPD service identity — `createdBy = { id, name: 'RPD' }` with no email. RPD is a genuine service name, so `'RPD'` sits in the `name` slot; a service has no email, so none is carried.

These transitions reach the backend through the machine-credential strategy, which already identifies the RPD service in the `name` slot. The external-transition route now selects `id` and `name` into the actor explicitly at the write boundary, so the machine marker the credentials carry never travels downstream to the stream write. Previously the route cast the credentials to a shape that omitted that marker and passed the whole object through, leaving the event clean only because the storage layer stripped the extra field.

A new end-to-end test drives a live RPD accept through the real machine-credential auth and the in-memory stream adapter, then asserts the appended `PRN_ACCEPTED` event carries `{ id, name: 'RPD' }` and no email.

Builds on [#1258](https://github.com/DEFRA/epr-backend/pull/1258), which threads the human actor's name and email through the same PRN ledger write chain; this PR is based on that branch.

### DIFF
--- a/src/packaging-recycling-notes/routes/external-transition-handler.js
+++ b/src/packaging-recycling-notes/routes/external-transition-handler.js
@@ -16,7 +16,7 @@ import { getProjectedPrnByNumber } from '#packaging-recycling-notes/application/
 
 /**
  * @import { Request, Lifecycle } from '@hapi/hapi'
- * @import { HapiRequest } from '#common/hapi-types.js'
+ * @import { HapiRequest, MachineCredentials } from '#common/hapi-types.js'
  * @import { PrnStatus } from '#packaging-recycling-notes/domain/model.js'
  * @import { PackagingRecyclingNotesRepository } from '#packaging-recycling-notes/repository/port.js'
  * @import { WasteBalancesRepository } from '#waste-balances/repository/port.js'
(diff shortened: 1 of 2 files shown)


src/packaging-recycling-notes/routes/external-transition-handler.js
```diff
@@ -82,9 +82,14 @@ export function createExternalTransitionHandler({
           ? new Date(payload[timestampField])
           : new Date()
 
-        const user = /** @type {{ id: string; name: string }} */ (
+        // RPD reaches this route through the machine-credential strategy, which
```
GraemeF: unnecessary comment

src/packaging-recycling-notes/routes/external-transition-handler.test.js
```diff
@@ -173,6 +174,32 @@ describe('external PRN transition read-side fold', () => {
     expect(response.statusCode).toBe(StatusCodes.CONFLICT)
   })
 
+  it('attributes a live RPD accept to the RPD service with no email', async () => {
+    // RPD drives the accept through the machine-credential strategy. The stream
```
GraemeF: unnecessary comment

src/packaging-recycling-notes/routes/external-transition-handler.js
GraemeF: Removed — the destructuring and the test name already say it.

src/packaging-recycling-notes/routes/external-transition-handler.test.js
GraemeF: Removed — the destructuring and the test name already say it.

PR #1281, opened 2026-06-05
PAE-1541: Add reportingDateField to table schemas

dansweeting:

## Summary

- Introduces `reporting-date-fields.js`, a central registry mapping `(processingType, tableName)` to the date field names that determine which reporting period a row belongs to
- All 13 table schemas declare `reportingDateFields` by referencing this registry via `Object.values()`
- The reports module's `SECTION_DATE_FIELDS_BY_OPERATOR_CATEGORY` derives its per-section date field mapping from the same registry by named access
- Adds a drift-guard test ensuring every schema's `reportingDateFields` is a non-empty array whose entries all appear in `requiredHeaders`
- Updates the `TableSchema` JSDoc typedef with the new property
- Fixes pre-existing type errors in `index.test.js` (nullable narrowing on a file we touched)

## Why

To classify waste records by reporting period (open vs closed), we need to know which date fields on each table determine the period a record belongs to. Until now, this knowledge lived in two independent places: the table schemas had no concept of reporting dates, and the reports module hardcoded a parallel mapping in `SECTION_DATE_FIELDS_BY_OPERATOR_CATEGORY`. Any code needing to assign a record to a period (e.g. the upcoming period-status classification) would have had to duplicate these rules or pick one of the two sources.

The new `reporting-date-fields.js` module is the single source of truth. Both consumers import from it, so the date field mappings can't drift apart.

## Design

The registry uses named objects at the leaf level:

```js
EXPORTER: {
  RECEIVED_LOADS_FOR_EXPORT: {
    DATE_RECEIVED_FOR_EXPORT: 'DATE_RECEIVED_FOR_EXPORT',
    DATE_OF_EXPORT: 'DATE_OF_EXPORT'
  }
}
```

- **Table schemas** call `Object.values()` to produce the `string[]` array for `reportingDateFields`
- **Reports module** accesses fields by name (e.g. `RDF.EXPORTER.RECEIVED_LOADS_FOR_EXPORT.DATE_RECEIVED_FOR_EXPORT`), avoiding fragile array indexing

The property is an array because some tables have multiple date fields relevant for period classification in different report sections. For accredited exporters, `RECEIVED_LOADS_FOR_EXPORT` has `DATE_RECEIVED_FOR_EXPORT` (determines the period for the received section) and `DATE_OF_EXPORT` (determines the period for the exported section). Changing either date on a row could affect a different closed period. Similarly, the registered-only exporter's `LOADS_EXPORTED` declares both `DATE_OF_EXPORT` and `DATE_THE_REFUSED_STOPPED_WASTE_REPATRIATED`. Most schemas have a single entry.

[PAE-1541]: https://eaflood.atlassian.net/browse/PAE-1541?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/domain/summary-logs/table-schemas/exporter-registered-only/loads-exported.js
+++ b/src/domain/summary-logs/table-schemas/exporter-registered-only/loads-exported.js
@@ -15,6 +15,7 @@ import {
 import { WASTE_RECORD_TYPE } from '#domain/waste-records/model.js'
 import { createRowTransformer } from '#application/waste-records/row-transformers/create-row-transformer.js'
 import { PROCESSING_TYPES } from '#domain/summary-logs/meta-fields.js'
+import { REPORTING_DATE_FIELDS } from '#domain/summary-logs/reporting-date-fields.js'
 const ALL_FIELDS = Object.values(FIELDS)
 
 /**
@@ -26,10 +27,9 @@ const ALL_FIELDS = Object.values(FIELDS)
  */
 export const LOADS_EXPORTED = {
   rowIdField: FIELDS.ROW_ID,
-  reportingDateFields: [
-    FIELDS.DATE_OF_EXPORT,
-    FIELDS.DATE_THE_REFUSED_STOPPED_WASTE_REPATRIATED
-  ],
+  reportingDateFields: Object.values(
+    REPORTING_DATE_FIELDS.EXPORTER_REGISTERED_ONLY.LOADS_EXPORTED
+  ),
   wasteRecordType: WASTE_RECORD_TYPE.EXPORTED,
   sheetName: 'Exported (sections 2 and 3)',
   rowTransformer: createRowTransformer({
--- a/src/domain/summary-logs/table-schemas/exporter-registered-only/loads-exported.js
+++ b/src/domain/summary-logs/table-schemas/exporter-registered-only/loads-exported.js
@@ -26,6 +26,7 @@ const ALL_FIELDS = Object.values(FIELDS)
  */
 export const LOADS_EXPORTED = {
   rowIdField: FIELDS.ROW_ID,
+  reportingDateField: FIELDS.DATE_OF_EXPORT,
   wasteRecordType: WASTE_RECORD_TYPE.EXPORTED,
   sheetName: 'Exported (sections 2 and 3)',
   rowTransformer: createRowTransformer({
--- a/src/domain/summary-logs/table-schemas/exporter-registered-only/loads-exported.js
+++ b/src/domain/summary-logs/table-schemas/exporter-registered-only/loads-exported.js
@@ -26,7 +26,7 @@ const ALL_FIELDS = Object.values(FIELDS)
  */
 export const LOADS_EXPORTED = {
   rowIdField: FIELDS.ROW_ID,
-  reportingDateField: FIELDS.DATE_OF_EXPORT,
+  reportingDateFields: [FIELDS.DATE_OF_EXPORT],
   wasteRecordType: WASTE_RECORD_TYPE.EXPORTED,
   sheetName: 'Exported (sections 2 and 3)',
   rowTransformer: createRowTransformer({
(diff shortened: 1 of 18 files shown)


GraemeF (approved):

src/domain/summary-logs/table-schemas/exporter-registered-only/loads-exported.js
```diff
@@ -26,6 +26,7 @@ const ALL_FIELDS = Object.values(FIELDS)
  */
 export const LOADS_EXPORTED = {
   rowIdField: FIELDS.ROW_ID,
+  reportingDateFields: [FIELDS.DATE_OF_EXPORT],
```
senthilinfy: I think its better to extract common logic to find whether a waste record falls into a reporting period. as similar logic already exists in reporting aggregation. If you check it there are more fields involved than this

senthilinfy: A record falls into a period when `wasteRecord.data[dateField]` is within `[startDate, endDate]` (inclusive, lexicographic). The field used depends on **operator category × report section** (`fields-by-operator-category.js`):

  | Operator category | Report section | `data` field | Format |
  |---|---|---|---|
  | `EXPORTER` | wasteReceived | `DATE_RECEIVED_FOR_EXPORT` | `YYYY-MM-DD` |
  | `EXPORTER` | wasteExported | `DATE_OF_EXPORT` | `YYYY-MM-DD` |
  | `EXPORTER` | wasteRepatriated | `DATE_THE_REFUSED_STOPPED_WASTE_REPATRIATED` | `YYYY-MM-DD` |
  | `EXPORTER` | wasteSentOn | `DATE_LOAD_LEFT_SITE` | `YYYY-MM-DD` |
  | `EXPORTER_REGISTERED_ONLY` | wasteReceived | `MONTH_RECEIVED_FOR_EXPORT` | `YYYY-MM` (normalised to `YYYY-MM-01`) |
  | `EXPORTER_REGISTERED_ONLY` | wasteExported | `DATE_OF_EXPORT` | `YYYY-MM-DD` |
  | `EXPORTER_REGISTERED_ONLY` | wasteRepatriated | `DATE_THE_REFUSED_STOPPED_WASTE_REPATRIATED` | `YYYY-MM-DD` |
  | `EXPORTER_REGISTERED_ONLY` | wasteSentOn | `DATE_LOAD_LEFT_SITE` | `YYYY-MM-DD` |
  | `REPROCESSOR` | wasteReceived | `DATE_RECEIVED_FOR_REPROCESSING` | `YYYY-MM-DD` |
  | `REPROCESSOR` | wasteSentOn | `DATE_LOAD_LEFT_SITE` | `YYYY-MM-DD` |
  | `REPROCESSOR_REGISTERED_ONLY` | wasteReceived | `MONTH_RECEIVED_FOR_REPROCESSING` | `YYYY-MM` (normalised to `YYYY-MM-01`) |
  | `REPROCESSOR_REGISTERED_ONLY` | wasteSentOn | `DATE_LOAD_LEFT_SITE` | `YYYY-MM-DD` |
  
   After date filtering, `received` records are further filtered to tonnage > 0 (`TONNAGE_RECEIVED_FOR_RECYCLING` / `TONNAGE_RECEIVED_FOR_EXPORT`). Records with a missing, null, or invalid date are silently excluded.

dansweeting: Yeah my instinct was that common logic should dictate both, the analysis was that the date fields would suffice but I'll take another look

dansweeting: Agreed, addressed in 425dbb16. Extracted a shared `reporting-date-fields.js` module that both the table schemas and `SECTION_DATE_FIELDS_BY_OPERATOR_CATEGORY` now derive from.

senthilinfy (approved):

PR #1288, opened 2026-06-08
PAE-1547: unlink org from Defra ID org + relax re-linking

timrutter:
Ticket: [PAE-1547](https://eaflood.atlassian.net/browse/PAE-1547)
## Description

Backend support for unlinking an organisation from its Defra ID organisation (INC2486016), and the relaxed linking logic that lets a user re-link afterwards.

## Summary

- Adds a `DELETE /v1/organisations/{organisationId}/link` admin endpoint that clears `linkedDefraOrganisation` while leaving the organisation status unchanged, recording an audit event, system log, and metric. Gated on `admin.write` (the `service_maintainer_write` tier).
- Relaxes the linking constraint so an organisation is a linking candidate when its status is `approved` **or** `active`-but-unlinked — without this, an unlinked organisation (which stays `active`) could never be re-linked.
- Fixes a latent bug found via the MongoDB contract test: clearing an optional field via `replace` stores `linkedDefraOrganisation` as `null` (not absent), so the linkable-orgs query now matches `null` (covering both never-linked and unlinked organisations) instead of `$exists: false`.

## Changes

- `src/routes/v1/organisations/unlink.js` — new DELETE endpoint (admin.write); clears the link, preserves status, audits + logs + meters.
- `src/routes/v1/organisations/index.js` — register the new route.
- `src/auditing/organisation-linking.js` — `auditOrganisationUnlinking` (audit event + system log).
- `src/common/helpers/metrics/organisation-linking.js` — `organisationUnlinkingMetrics` (`organisation.unlinked`).
- `src/routes/v1/organisations/link.js` — accept `approved` or `active` organisations as linkable.
- `src/repositories/organisations/inmemory.js` / `mongodb.js` — `findAllLinkableForUser` includes active-but-unlinked orgs; Mongo query matches `linkedDefraOrganisation: null`.
- `src/domain/organisations/model.js` — shared `LINKABLE_ORGANISATION_STATUSES` constant.
- Tests: `unlink.test.js` (new), `link.test.js`, `find.contract.js`, `organisation-linking.test.js` (metrics) — covering the unlink flow, re-linking an active-but-unlinked org, and both never-linked (absent) and unlinked (null) candidate states across both adapters.

## Out of scope

The admin-frontend unlink journey (overview page, unlink button, confirm page) and the runbook are tracked separately; this PR is backend only.

[PAE-1547]: https://eaflood.atlassian.net/browse/PAE-1547?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/routes/v1/organisations/link.js
```diff
@@ -63,8 +66,8 @@ export const organisationsLink = {
       ...organisation
     } = await organisationsRepository.findById(organisationId)
 
-    if (organisation.status !== ORGANISATION_STATUS.APPROVED) {
-      throw Boom.conflict('Organisation is not in an approvable state')
+    if (!LINKABLE_ORGANISATION_STATUSES.includes(organisation.status)) {
```
crosslandwa, line 69: I think this check (now) _also_ needs throw a `Boom.conflict` if the underlying `epr-organisation` is linked (ie has `epr-organisation.linkedDefraIdOrganisation` populated) to prevent an incoming request to (change the) link for an "active and linked" org

_This guard was previously implicit my virtue of the `status` changing to `active` happening in lock-step with populating the linked org_

src/routes/v1/organisations/unlink.js
```diff
@@ -0,0 +1,56 @@
...
+
+    const { organisationsRepository } = request
+    const {
+      id,
+      version: currentVersion,
+      ...organisation
+    } = await organisationsRepository.findById(organisationId)
+
+    if (!organisation.linkedDefraOrganisation) {
+      throw Boom.notFound('Organisation is not linked so cannot be unlinked')
```
crosslandwa: Maybe `Boom.conflict` here rather than not found

crosslandwa (approved):

src/routes/v1/organisations/unlink.js
timrutter: done

src/routes/v1/organisations/link.js
timrutter: have added an additinoal check not to allow linking when already linked

PR #1291, opened 2026-06-08
PAE-1555: Stop reading and writing the vestigial waste-balances collection

GraemeF:
Ticket: [PAE-1555](https://eaflood.atlassian.net/browse/PAE-1555)
Since the production cutover the event-sourced waste-balance stream is the sole source of truth for every accreditation's balance, and the `waste-balances` collection has been reduced to a vestigial seed/index shell whose stored `amount`/`availableAmount` are stale and no longer read. This stops all application reads and writes of that collection so the stream is the only path the code touches.

## What changed

- The waste-balances repository is collapsed onto the stream: it reads balances by `(registrationId, accreditationId)` partition from the latest stream event and writes by appending events. The shell write (`saveBalance`) and the startup creation of the `accreditationId` unique index are gone, and the dead `amount`/`availableAmount`/`version`/`schemaVersion` shell fields drop out of the read model.
- The single-balance read replaces `findByAccreditationId`/`findByAccreditationIds` with a partition-keyed `findBalance({ registrationId, accreditationId })`. The PRN deduct/credit/append paths source `registrationId` from the caller instead of loading the shell to resolve it, dropping the shell existence-gate and the `version` dependency.
- The public `GET /v1/organisations/{organisationId}/waste-balances` endpoint resolves each accreditation's `registrationId` from the organisation document (whose registrations carry the accreditation link). The request contract — `(organisationId, accreditationIds[])` — is unchanged. An accreditation that is not registered to the path organisation, whether unknown or belonging to a different organisation, is omitted from the response map, and an unknown organisation yields an empty map; the endpoint returns 200 with whatever resolved. Cross-organisation reads remain impossible because an accreditation is only ever paired with the path organisation's own registration before the stream is queried. A registrationId-disambiguated read route is tracked separately.
- The waste-balance availability aggregation enumerates accreditations from `epr-organisations` (unwinding registrations) and looks up stream events directly, rather than driving the pipeline off the `waste-balances` collection. This also removes the fragile coupling whereby an accreditation with stream events but no shell document would be silently skipped.
- The non-production data reset no longer touches the collection either: its cascade step is dropped, so the application holds no remaining reference to `waste-balances` at all. The residual rows are inert — nothing reads them and nothing inserts, so the still-present `accreditationId` unique index cannot be violated — and they will go when the collection itself is dropped.

The `waste-balances` collection and its `accreditationId` unique index are intentionally left in MongoDB, orphaned and stale — there is no drop, migration or deletion. Removing the collection physically is a separate follow-up.

[PAE-1555]: https://eaflood.atlassian.net/browse/PAE-1555?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/non-prod-data-reset/mongodb.test.js
```diff
@@ -212,12 +196,15 @@ const seedDownstreamForOrganisation = async (
...
-  )
+  // The waste-balances collection is orphaned: nothing in the app writes to it
+  // any more. Seed a residual shell directly so the cascade's teardown of the
+  // collection stays covered.
+  await database.collection('waste-balances').insertOne({
+    _id: new ObjectId(),
+    accreditationId,
+    organisationId,
+    registrationId
+  })
```
dansweeting: If we're truly not writing to this collection I'd expect we could simply stop cleaning it up; remove it from the set of collections we watch for?

GraemeF: Good shout — done. Dropped the `waste-balances` step from the cascade (plus the constant, the empty-counts entry and the now-dead empty-accreditations branch). The residual rows are inert anyway: nothing reads them and nothing inserts, so the still-present unique index can't bite. That was the last reference to the collection anywhere in the app — the read/write paths already go through `waste-balance-events` only — so the code now has zero mentions of `waste-balances`. The collection and its index stay in Mongo, orphaned, until we drop them as a separate follow-up.

dansweeting (approved):

PR #1300, opened 2026-06-09
PAE-1541: Classify loads by reporting period status

dansweeting:

## Summary

This PR provides backend capability for the ticket's [related designs](https://www.figma.com/design/scTmpd6bQJQxLF6U26hwqA/Waste-Management?node-id=13117-45366&t=qPcFZYJ3YfEd6vji-4) with the exception of the accordion breakdown of individual rows (to follow).

Adds period-status classification to the summary log validation pipeline. Each validated waste record is classified into an open or closed reporting period bucket, then further split by whether the record was added or adjusted, and whether it moved the waste balance.

The `GET /v1/organisations/{organisationId}/registrations/{registrationId}/summary-logs/{summaryLogId}` response gains an optional `loadsByReportingPeriod` field on validated summary logs:

```json
{
  "loadsByReportingPeriod": {
    "openPeriodLoads": {
      "added": {
        "balanceAffecting": { "count": 3, "tonnageDelta": 42.5 },
        "nonBalanceAffecting": { "count": 1 }
      },
      "adjusted": {
        "balanceAffecting": { "count": 1, "tonnageDelta": -2.0 },
        "nonBalanceAffecting": { "count": 0 }
      }
    },
    "closedPeriodLoads": {
      "added": { ... },
      "adjusted": { ... }
    }
  }
}
```

| Field | Meaning |
|---|---|
| `openPeriodLoads` / `closedPeriodLoads` | Whether the record's reporting period has been submitted. A period is closed when a report has been submitted for it (current status is `submitted` or there are previous submissions). |
| `added` / `adjusted` | Whether the record was newly created or updated by this summary log. |
| `balanceAffecting` / `nonBalanceAffecting` | Whether the record actually moved the waste balance, decided per leg from its rounded net tonnage delta. A leg is `balanceAffecting` when its delta is non-zero, otherwise `nonBalanceAffecting`. This makes all balance movement (including amendment reversals) live in `balanceAffecting`, so a row that is valid but contributes nothing (eg PRN-issued, outside accreditation, or an unchanged resubmission) is `nonBalanceAffecting`. |
| `count` | Number of record legs in this bucket. |
| `tonnageDelta` | Present only on `balanceAffecting`. The net waste balance impact: for added records the transaction amount, for adjusted records the net effect across the old and new period classifications. Summing `tonnageDelta` across all four `balanceAffecting` buckets gives the total predicted impact on the waste balance. `nonBalanceAffecting` carries no `tonnageDelta` because it is always zero. |

The field is absent when validation fails (status is `INVALID`). When the reports lookup fails the whole validation fails rather than persisting a misclassified split: the error propagates, the queue consumer retries the transient failure via SQS redrive, and the log is marked `validation_failed` once retries are exhausted. All inner buckets are always present when the field exists (no null nesting).

### Business rules

- **Closed-wins rule**: when a row has multiple reporting date fields (eg exporter received-loads with both `DATE_RECEIVED_FOR_EXPORT` and `DATE_OF_EXPORT`), if any date maps to a submitted period the entire row is classified as closed.
- **Per-leg classification**: an adjusted record contributes to each period it touches. The old period gets `-oldAmount` and the new period gets `+newAmount`; when both fall in the same period these merge into a single net leg. Each leg is independently `balanceAffecting` or `nonBalanceAffecting` based on its own rounded delta, so a same-period amendment that nets to zero (eg an identical resubmission) is `nonBalanceAffecting`.
- **Cadence**: accredited registrations use monthly periods, registered-only use quarterly.
- **IGNORED and unchanged records**: skipped entirely.

### Design

The classification is a pure function (`classifyByPeriodStatus` in `period-status.js`) with no I/O. Each record produces 0-2 leg entries which are folded into the final nested structure; inclusion is derived from each leg's rounded delta at fold time. The reports fetch lives in `validate.js` alongside the existing orchestration, and a fetch failure propagates so the consumer can retry then fail the log.

`reportsRepository` is threaded through the queue-consumer plugin to the validator factory.

[PAE-1541]: https://eaflood.atlassian.net/browse/PAE-1541

### DIFF
--- a/src/application/summary-logs/classify-and-persist.js
+++ b/src/application/summary-logs/classify-and-persist.js
@@ -24,7 +24,7 @@ import {
 /** @import {Registration} from '#domain/organisations/registration.js' */
 /** @import {SubmittedSummaryLog} from './validate-issue-logging.js' */
 /** @typedef {import('./load-counts.js').Loads} Loads */
-/** @typedef {import('./period-status.js').LoadsByPeriodStatus} LoadsByPeriodStatus */
+/** @typedef {import('./period-status.js').LoadsByReportingPeriod} LoadsByReportingPeriod */
 /** @typedef {string} ProcessingType */
 /** @typedef {import('#reports/repository/port.js').ReportsRepository} ReportsRepository */
 
@@ -53,7 +53,7 @@ export const filterWasteBalanceRecords = (wasteRecords, processingType) =>
  * @param {import('#reports/repository/port.js').PeriodicReport[]} params.periodicReports
  * @param {Registration} [params.registration]
  * @param {Map<string, WasteRecord>} [params.existingRecordsMap]
- * @returns {{ loads: Loads | null, loadsByWasteRecordType: import('./load-counts.js').LoadsByWasteRecordType | null, loadsByPeriodStatus: LoadsByPeriodStatus | null }}
+ * @returns {{ loads: Loads | null, loadsByWasteRecordType: import('./load-counts.js').LoadsByWasteRecordType | null, loadsByReportingPeriod: LoadsByReportingPeriod | null }}
  */
 export const classifyLoads = ({
   processingType,
@@ -69,7 +69,7 @@ export const classifyLoads = ({
     return {
       loads: null,
       loadsByWasteRecordType: null,
-      loadsByPeriodStatus: null
+      loadsByReportingPeriod: null
     }
   }
 
@@ -91,7 +91,7 @@ export const classifyLoads = ({
   })
 
   /* v8 ignore start -- defensive guard: all three are always set when status is VALIDATED */
-  const loadsByPeriodStatus =
+  const loadsByReportingPeriod =
     registration && existingRecordsMap && tableSchemas
       ? classifyByPeriodStatus({
           wasteRecords,
@@ -110,7 +110,7 @@ export const classifyLoads = ({
       : null
   /* v8 ignore stop */
 
-  return { loads, loadsByWasteRecordType, loadsByPeriodStatus }
+  return { loads, loadsByWasteRecordType, loadsByReportingPeriod }
 }
 
 /**
--- a/src/application/summary-logs/classify-and-persist.js
+++ b/src/application/summary-logs/classify-and-persist.js
@@ -1,0 +1,154 @@
+import {
+  LOGGING_EVENT_ACTIONS,
+  LOGGING_EVENT_CATEGORIES
+} from '#common/enums/index.js'
+import { isNil } from '#common/helpers/is-nil.js'
+import { SUMMARY_LOG_STATUS } from '#domain/summary-logs/status.js'
+import {
+  findSchemaForProcessingType,
+  PROCESSING_TYPE_TABLES
+} from '#domain/summary-logs/table-schemas/index.js'
+import { isRegistrationAccredited } from '#domain/organisations/registration-utils.js'
+import { ORS_VALIDATION_DISABLED } from '#domain/summary-logs/table-schemas/shared/classification-reason.js'
+import { classifyByPeriodStatus } from './period-status.js'
+import {
+  countByValidity,
+  countByWasteBalanceInclusion,
+  countByWasteRecordType,
+  mergeLoads
+} from './load-counts.js'
+
+/** @import {TypedLogger} from '#common/helpers/logging/logger.js' */
+/** @import {ValidatedWasteRecord} from '#application/waste-records/transform-from-summary-log.js' */
+/** @import {WasteRecord} from '#domain/waste-records/model.js' */
+/** @import {Registration} from '#domain/organisations/registration.js' */
+/** @import {SubmittedSummaryLog} from './validate-issue-logging.js' */
+/** @typedef {import('./load-counts.js').Loads} Loads */
+/** @typedef {import('./period-status.js').LoadsByPeriodStatus} LoadsByPeriodStatus */
+/** @typedef {string} ProcessingType */
+/** @typedef {import('#reports/repository/port.js').ReportsRepository} ReportsRepository */
+
+/**
+ * Filters waste records to only those from tables that participate in waste balance.
+ *
+ * @param {ValidatedWasteRecord[] | null} wasteRecords
+ * @param {string} processingType
+ * @returns {ValidatedWasteRecord[]}
+ */
+export const filterWasteBalanceRecords = (wasteRecords, processingType) =>
+  wasteRecords?.filter((wr) => {
+    const schema = findSchemaForProcessingType(processingType, wr.record.type)
+    return !isNil(schema?.classifyForWasteBalance)
+  }) ?? []
+
+/**
+ * Computes all load classifications for validated summary logs.
+ *
+ * @param {Object} params
+ * @param {string} params.status - Summary log status after validation
+ * @param {ValidatedWasteRecord[] | null} params.wasteRecords - All waste records
+ * @param {ValidatedWasteRecord[]} params.wasteBalanceRecords - Waste-balance-eligible records
+ * @param {string} params.summaryLogId
+ * @param {ProcessingType} params.processingType
+ * @param {import('#reports/repository/port.js').PeriodicReport[]} params.periodicReports
+ * @param {Registration} [params.registration]
+ * @param {Map<string, WasteRecord>} [params.existingRecordsMap]
+ * @returns {{ loads: Loads | null, loadsByWasteRecordType: import('./load-counts.js').LoadsByWasteRecordType | null, loadsByPeriodStatus: LoadsByPeriodStatus | null }}
+ */
+export const classifyLoads = ({
+  processingType,
+  status,
+  summaryLogId,
+  wasteBalanceRecords,
+  wasteRecords,
+  periodicReports,
+  registration,
+  existingRecordsMap
+}) => {
+  if (status !== SUMMARY_LOG_STATUS.VALIDATED || !wasteRecords) {
+    return {
+      loads: null,
+      loadsByWasteRecordType: null,
+      loadsByPeriodStatus: null
+    }
+  }
+
+  const loads = mergeLoads(
+    countByValidity({ wasteRecords, summaryLogId }),
+    countByWasteBalanceInclusion({
+      wasteRecords: wasteBalanceRecords,
+      summaryLogId
+    })
+  )
+
+  const tableSchemas = PROCESSING_TYPE_TABLES[processingType]
+
+  const loadsByWasteRecordType = countByWasteRecordType({
+    wasteRecords,
+    wasteBalanceRecords,
+    summaryLogId,
+    tableSchemas
+  })
+
+  /* v8 ignore start -- defensive guard: all three are always set when status is VALIDATED */
+  const loadsByPeriodStatus =
+    registration && existingRecordsMap && tableSchemas
+      ? classifyByPeriodStatus({
+          wasteRecords,
+          existingRecordsMap,
+          periodicReports,
+          cadence: isRegistrationAccredited(registration)
+            ? 'monthly'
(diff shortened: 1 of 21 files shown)


src/application/summary-logs/period-status.js
```diff
@@ -0,0 +1,421 @@
...
+  summaryLogId,
+  tableSchemas,
+  classificationContext
+}) => {
+  const closedPeriods = buildClosedPeriods(periodicReports, cadence)
+
+  const classify = (
+    /** @type {Record<string, string | Date | null | undefined>} */ data,
+    /** @type {string[]} */ reportingDateFields
+  ) => classifyPeriodStatus(data, reportingDateFields, closedPeriods, cadence)
```
dansweeting: For the sake of just passing `closedPeriods` and `cadence` down at line 390 this could just be inlined?

src/application/summary-logs/classify-and-persist.js
```diff
@@ -0,0 +1,154 @@
...
+          existingRecordsMap,
+          periodicReports,
+          cadence: isRegistrationAccredited(registration)
+            ? 'monthly'
+            : 'quarterly',
+          summaryLogId,
+          tableSchemas,
+          classificationContext: {
+            accreditation: registration.accreditation ?? null,
+            overseasSites: ORS_VALIDATION_DISABLED
```
senthilinfy, line 103: `overseasSites: ORS_VALIDATION_DISABLED` is this right?

src/application/summary-logs/classify-and-persist.js
```diff
@@ -0,0 +1,154 @@
...
+
+  /* v8 ignore start -- defensive guard: all three are always set when status is VALIDATED */
+  const loadsByReportingPeriod =
+    registration && existingRecordsMap && tableSchemas
+      ? classifyByPeriodStatus({
+          wasteRecords,
+          existingRecordsMap,
+          periodicReports,
+          cadence: isRegistrationAccredited(registration)
+            ? 'monthly'
```
senthilinfy: can we use the `CADENCE` enum

src/application/summary-logs/classify-and-persist.js
```diff
@@ -0,0 +1,154 @@
...
+export const fetchPeriodicReportsSafe = async ({
+  registration,
+  status,
+  summaryLog,
+  reportsRepository,
+  loggingContext,
+  logger
+}) => {
+  try {
+    if (registration && status === SUMMARY_LOG_STATUS.VALIDATED) {
```
senthilinfy: I think this should fail the summary log upload? Otherwise when CMA is live, it will show incorrect info to operators without Closed/Open period split

src/application/summary-logs/period-status.js
```diff
@@ -0,0 +1,436 @@
...
+ */
+const buildClosedPeriods = (periodicReports, cadence) => {
+  const closed = new Set()
+  for (const periodicReport of periodicReports) {
+    const slots = periodicReport.reports[cadence]
+    if (!slots) {
+      continue
+    }
+    for (const [period, slot] of Object.entries(slots)) {
+      const hasBeenSubmitted =
```
senthilinfy, line 134: Is it worth moving all logic related to reporting into reporting module application layer(is a date belong to closed/open period, not calling reporting repository) ? 

To keep module separation and they talking to each other only via application layer.

src/application/summary-logs/period-status.js
```diff
@@ -0,0 +1,436 @@
...
+
+    hasAnyDate = true
+    const period = monthToPeriod(extractMonth(dateValue), cadence)
+    const periodKey = `${extractYear(dateValue)}:${period}`
+    if (closedPeriods.has(periodKey)) {
+      return 'closed'
+    }
+  }
+
+  return hasAnyDate ? 'open' : null
```
senthilinfy: (nitpick) Consider using enum for Open/Closed string values

src/application/summary-logs/period-status.js
```diff
@@ -0,0 +1,436 @@
...
+ * Computes the transaction amount for a record via classifyForWasteBalance.
+ * Returns 0 if the schema has no classifier or the outcome is not INCLUDED.
+ *
+ * @param {import('#domain/summary-logs/table-schemas/index.js').TableSchema | null} schema
+ * @param {Record<string, any>} data
+ * @param {ClassificationContext} context
+ * @returns {number}
+ */
+const getTransactionAmount = (schema, data, context) => {
+  const result = schema?.classifyForWasteBalance?.(data, context)
```
senthilinfy, line 204: `classifyForWasteBalance` is already called in `filterWasteBalanceRecords`, is it necessary to duplicate the call?  It seems like non trivial call in term of efficiency as it does multiple checks, but I could be wrong

src/domain/summary-logs/loads-by-period-status-schema.js
```diff
@@ -0,0 +1,43 @@
...
+})
+
+const periodStatusByChangeSchema = Joi.object({
+  added: periodStatusGroupSchema.required(),
+  adjusted: periodStatusGroupSchema.required()
+})
+
+export const loadsByReportingPeriodSchema = Joi.object({
+  openPeriodLoads: periodStatusByChangeSchema.required(),
+  closedPeriodLoads: periodStatusByChangeSchema.required()
```
senthilinfy, line 30: Just asking, feel free to ignore as you have better understanding of overall ask.

When a record is amended from `closed` (30t) to `open` (50t), I find below output confusing as `closedPeriodLoads` has `-30` with count as 0. Is it better to model using suggestion at end to demarcate cross period adjustments?  Even if not needed today could be potentially useful in future as output has complete details about records that moved from Open->Closed or other-way.
```

**Current**
{
    "openPeriodLoads": {
      "adjusted": {
        "balanceAffecting":    { "count": 1, "tonnageDelta": 50 },
        "nonBalanceAffecting": { "count": 0 }
      }
    },
    "closedPeriodLoads": {
      "adjusted": {
        "balanceAffecting":    { "count": 0, "tonnageDelta": -30 },
        "nonBalanceAffecting": { "count": 0 }
      }
    }
  }
```

**Suggestion**
```
  {
    "openPeriodLoads": {
      "adjusted": {
        "balanceAffecting":    { "count": 0, "tonnageDelta": 0 },
        "nonBalanceAffecting": { "count": 0 }
      },
      "crossPeriodAdjusted": {
        "inflows":  { "count": 1, "tonnageDelta": 50 },
        "outflows": { "count": 0, "tonnageDelta": 0  }
      }
    },
    "closedPeriodLoads": {
      "adjusted": {
        "balanceAffecting":    { "count": 0, "tonnageDelta": 0 },
        "nonBalanceAffecting": { "count": 0 }
      },
      "crossPeriodAdjusted": {
        "inflows":  { "count": 0, "tonnageDelta": 0   },
        "outflows": { "count": 1, "tonnageDelta": -30 }
      }
    }
  }
```

src/application/summary-logs/period-status.js
dansweeting: Addressed in e176be33: the classify closure is gone and `closedPeriods` and `cadence` are now passed directly. Resolving.

src/application/summary-logs/period-status.js
dansweeting: `filterWasteBalanceRecords` only checks that the classifier *exists* (`!isNil(schema?.classifyForWasteBalance)`) to decide whether a record is waste-balance-eligible; it never invokes it. This line is the only place `classifyForWasteBalance` is actually called for the period-status path, and we need its result for the `transactionAmount` (which the filter discards). So no extra work is being repeated against the filter.

src/application/summary-logs/classify-and-persist.js
dansweeting: Skipping ORS validation was OK up til now, at this stage in the pipeline, but now we are concerning ourselves with any given row's impact on waste balance. So I don't think we can get away with it: Doing so will actually ignore the impact of any loads with ORS validation, and the page will report an inaccurate delta to the operator.

If it's not too much more adding to the diff we could fix it here... or I think we could follow up quickly after this PR to wire up ORS validation, and in that case we'd see the impact on waste balance filtering through.

src/application/summary-logs/classify-and-persist.js
dansweeting: Fixed by removing the swallow so the error propagates. Because the error is not a `PermanentError`, the queue consumer treats it as transient: SQS retries via the redrive policy, and only once `maxReceiveCount` is exhausted does `onFailure` mark the log `validation_failed`. We throw before `persistValidationResult`, so no incorrect open/closed split is ever written and the log stays `validating` for clean retries. So a transient blip self-heals, and a persistent failure fails the upload as you wanted.

src/domain/summary-logs/loads-by-period-status-schema.js
dansweeting: You're right that the current `closed: { count: 0, tonnageDelta: -30 }` reads oddly. A couple of things to note though:

The deltas are already accurate and complete: `-30` (out of closed) plus `+50` (into open) sum to `+20`, which is the true net impact of the row going 30t to 50t. So no information is missing, it's only the `count` that looks off.

Rather than the `crossPeriodAdjusted` shape, I think the lighter fix is to set `count: 1` on every leg a record touches (today it only sits on the record's 'home'/new period). A cross-period amendment would then read:

```
closedPeriodLoads.adjusted.balanceAffecting: { count: 1, tonnageDelta: -30 }
openPeriodLoads.adjusted.balanceAffecting:   { count: 1, tonnageDelta: +50 }
```

The signed delta already tells you it's an outflow from closed and an inflow into open, so you get the same demarcation without a new sub-structure for the FE to learn, and the buckets stay flat.

This also matters for what's coming next: cross-period amendments will soon prompt the operator to resubmit the affected period's report. With `count: 0` the closed period that needs resubmitting looks empty; with `count: 1` it correctly shows '1 adjusted load affecting this closed period'.

WDYT @senthilinfy ?

The net effect for the operator is that this load shows up in both the Open and Closed sections of the visual breakdown.

src/application/summary-logs/period-status.js
dansweeting: Agreed, this is the right direction. One clarification first: the classifier here is a pure function and doesn't call the reports repository (`validate.js` does that and passes the data in), so the repository-coupling half is already handled.

What's left is that `period-status.js` re-derives two reports-domain facts from the raw slot shape: (1) which reporting period a date falls in, and (2) whether a period has been submitted (`current.status === submitted || previousSubmissions.length > 0`). That second rule also duplicates the slot-reading `report-service` already does on the reports side.

The clean fix is to expose a small reports-side API (a `periodForDate` and an `isPeriodSubmitted`/submitted-period-keys helper) so summary-logs talks to reports through an intentional boundary rather than its storage shape, with the 'submitted' rule living in one place. The closed-wins classification (a row with multiple date fields is closed if any date lands in a submitted period) stays here, as that's summary-log domain logic.

I'd like to keep it out of this PR since it reaches into the reports module and pulls in a de-dup of `report-service` that is better reviewed on its own. Raising it as a fast follow-up.

src/domain/summary-logs/loads-by-period-status-schema.js
dansweeting: Following up: we'll do this as a fast follow rather than in this PR, to keep this change self-contained. Raised as tracked work. The deltas are already accurate here, so it's purely the count:1-in-both improvement that lands next.

src/application/summary-logs/classify-and-persist.js
dansweeting: We'll wire up ORS validation as a fast follow rather than expanding this PR's diff. Raised as tracked work so the waste-balance delta stops ignoring ORS-validated loads. Until that lands the split is accurate for everything except loads that depend on ORS validation.

senthilinfy (approved):

PR #1303, opened 2026-06-11
PAE-1556: Record user role on system log entries

GraemeF:
Ticket: [PAE-1556](https://eaflood.atlassian.net/browse/PAE-1556)
System logs display a user's scopes under a "role" label, and the user's actual role is not recorded at all. This threads the authenticated user's role through to the system log so it can be displayed correctly (PAE-1556).

When a request authenticates, the JWT strategy now carries the resolved `role` on the credentials alongside `scope`, for both identity providers. For Entra ID (admins) the role is resolved from `ADMIN_ROLES`; for Defra ID (operators) it is currently `null` — operators have scopes but no role mapping yet. Mapping operator scopes to a role is a domain decision that needs a separate conversation, so this change deliberately plumbs it through as nullable rather than inventing a mapping.

The audit helper that builds the `createdBy` actor for a human session now records `role` next to `id`, `email` and `scope`, so the role lands on every system log entry. The system-logs search endpoint and repositories are pure pass-through, so the role surfaces on read with no further change there — covered by a round-trip test.

Some system logs are written asynchronously: ORS imports and summary log submissions hand the actor to an SQS command that a queue consumer processes and records. The consumer command schemas validate the actor, so they now accept the nullable `role` field, letting it survive the round-trip to where the async log is written.

Types: `HumanCredentials` gains a required `role: string | null` (freshly minted on every authenticated request, so always present). `SystemLogActor` gains an optional `role?: string | null`, because logs written before this change carry only `id`, `email` and `scope`.

Note: this touches `src/common/helpers/auth/get-jwt-strategy-config.js`, which is part of the in-flight role/scopes refactor series — worth checking for conflicts before merge.

The frontend change to split the "role" label from the scopes and show both is tracked separately and will follow in an admin-frontend PR.

[PAE-1556]: https://eaflood.atlassian.net/browse/PAE-1556?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/auditing/helpers.test.js
+++ b/src/auditing/helpers.test.js
@@ -1,7 +1,7 @@
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest'
 import { audit } from '@defra/cdp-auditing'
 import { logger } from '#common/helpers/logging/logger.js'
-import { safeAudit, recordSystemLogs } from './helpers.js'
+import { extractUserDetails, safeAudit, recordSystemLogs } from './helpers.js'
 
 vi.mock('@defra/cdp-auditing', () => ({
   audit: vi.fn()
@@ -104,6 +104,65 @@ describe('safeAudit', () => {
 
     expect(audit).toHaveBeenCalledWith({
       event: payload.event
+    })
+  })
+})
+
+describe('extractUserDetails', () => {
+  it('records the role alongside id, email and scope for a human actor', () => {
+    const request = {
+      auth: {
+        credentials: {
+          id: 'contact-123',
+          email: 'maintainer@example.com',
+          scope: ['admin.read', 'admin.dlq.purge'],
+          role: 'service_maintainer'
+        }
+      }
+    }
+
+    expect(extractUserDetails(request)).toEqual({
+      id: 'contact-123',
+      email: 'maintainer@example.com',
+      scope: ['admin.read', 'admin.dlq.purge'],
+      role: 'service_maintainer'
+    })
+  })
+
+  it('records a null role for a human actor with no resolved role', () => {
+    const request = {
+      auth: {
+        credentials: {
+          id: 'contact-456',
+          email: 'operator@example.com',
+          scope: ['inquirer'],
+          role: null
+        }
+      }
+    }
+
+    expect(extractUserDetails(request)).toEqual({
+      id: 'contact-456',
+      email: 'operator@example.com',
+      scope: ['inquirer'],
+      role: null
+    })
+  })
+
+  it('does not record a role for a machine actor', () => {
+    const request = {
+      auth: {
+        credentials: {
+          id: 'machine-1',
+          isMachine: true,
+          name: 'batch-job'
+        }
+      }
+    }
+
+    expect(extractUserDetails(request)).toEqual({
+      id: 'machine-1',
+      name: 'batch-job'
     })
   })
 })
--- a/src/auditing/helpers.test.js
+++ b/src/auditing/helpers.test.js
@@ -110,16 +110,17 @@ describe('safeAudit', () => {
 
 describe('extractUserDetails', () => {
   it('records the role alongside id, email and scope for a human actor', () => {
-    const request = {
-      auth: {
-        credentials: {
-          id: 'contact-123',
-          email: 'maintainer@example.com',
-          scope: ['admin.read', 'admin.dlq.purge'],
-          role: 'service_maintainer'
+    const request =
+      /** @type {import('#common/hapi-types.js').HapiRequest} */ ({
+        auth: {
+          credentials: {
+            id: 'contact-123',
+            email: 'maintainer@example.com',
+            scope: ['admin.read', 'admin.dlq.purge'],
+            role: 'service_maintainer'
+          }
         }
-      }
-    }
+      })
 
     expect(extractUserDetails(request)).toEqual({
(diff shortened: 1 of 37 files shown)


src/auditing/helpers.test.js
```diff
@@ -108,6 +108,68 @@ describe('safeAudit', () => {
...
+
+    expect(extractUserDetails(request)).toEqual({
+      id: 'contact-456',
+      email: 'operator@example.com',
+      scope: ['inquirer'],
+      role: null
+    })
+  })
+
+  it('does not record a role for a machine actor', () => {
```
crosslandwa, line 154: I was wondering if this is an opportunity to retire the `isMachine` boolean, and instead provide an appropriate `role` when the incoming request is authorised, ie. something like

- incoming requests from reg-accred service (currently the basic auth client) would get `credentials.role = 'registration-and-accreditation-service`
- incoming requests from RPD to accept/reject PRNs (currently the cognito auth client) would get `credentials.role = 'rpd-service'`
- queue processing routines would get `credentials.role = 'epr-backend-service'`

(though I'm not sure if `isMachine` is set when messages from the queue are processed)

Maybe something to consider/layer-in in a followup PR

src/common/helpers/auth/get-jwt-strategy-config.test.js
```diff
@@ -139,11 +139,52 @@ describe('#getJwtStrategyConfig', () => {
...
+          }
+        }
+      }
+
+      const result = await config.validate(artifacts)
+
+      expect(result.credentials.role).toBe('service_maintainer')
+    })
+
+    test('credential role is null when user matches no admin tier', async () => {
```
crosslandwa: The behaviour this test describes is implemented in `getEntraUserRoles` (which is mocked here) rather that in `config.validate`

Suggest one of
- removing this test
- moving this test to `get-entra-user-roles.test.js`
- removing the mocking, and allowing the real `getEntraUserRoles` to run

src/overseas-sites/imports/repository/port.js
```diff
@@ -11,7 +11,7 @@
  * @property {string} _id
  * @property {string} status - 'preprocessing' | 'processing' | 'completed' | 'failed'
  * @property {OrsImportFile[]} files
- * @property {{ id: string, email: string, scope: string[] }} [createdBy] - User who initiated the import
+ * @property {{ id: string, email: string, scope: string[], role?: string | null }} [createdBy] - User who initiated the import
```
crosslandwa: I think this would be easier to reason about as

```
 * @property {{ id: string, email: string, scope: string[], role: string | null }} [createdBy] - User who initiated the import
```

ie force any callers omitting the `role` to pass `null` - how big an impact is that?

src/repositories/system-logs/port.js
```diff
@@ -1,9 +1,11 @@
...
- * The actor on a system log. A human session carries an email and scope; a
- * machine credential carries a name instead. Both always carry an id.
+ * The actor on a system log. A human session carries an email, scope and the
+ * role it resolved to (null when none); a machine credential carries a name
+ * instead. Both always carry an id. Role is optional because logs predating
+ * role capture carry only email and scope.
  *
  * @typedef {(
- *   | { id: string, email: string, scope: string[] }
+ *   | { id: string, email: string, scope: string[], role?: string | null }
```
crosslandwa: Same as comment above - can the `?` be removed from `role` - clearer if clients _have_ to pass `null` or a value I think

src/common/helpers/auth/get-jwt-strategy-config.test.js
GraemeF: Dropped the test - it's covered elsewhere

src/overseas-sites/imports/repository/port.js
GraemeF: Done!

src/auditing/helpers.test.js
GraemeF: Ideally we'd have the identity of the actor using the other system rather than the system itself, but I'll leave it alone for this PR anyway

src/server/queue-consumer/ors-import-commands.js
```diff
@@ -18,7 +18,8 @@ import { orsImportMetrics } from '#overseas-sites/metrics/ors-imports.js'
 const userSchema = Joi.object({
   id: Joi.string().required(),
   email: Joi.string().required(),
-  scope: Joi.array().items(Joi.string()).required()
+  scope: Joi.array().items(Joi.string()).required(),
+  role: Joi.string().allow(null)
```
crosslandwa: should this have a `.required()`?

src/server/queue-consumer/summary-log-commands.js
```diff
@@ -33,7 +33,8 @@ const userSchema = Joi.object({
   id: Joi.string().required(),
   name: Joi.string(),
   email: Joi.string().required(),
-  scope: Joi.array().items(Joi.string()).required()
+  scope: Joi.array().items(Joi.string()).required(),
+  role: Joi.string().allow(null)
```
crosslandwa: same as above - should this have `.required()`?

crosslandwa (approved):

PR #1335, opened 2026-06-16
PAE-1566: Prevent duplicate accreditation links when editing organisations

senthilinfy:
Ticket: [PAE-1566](https://eaflood.atlassian.net/browse/PAE-1566)
## Summary
- Org replace now rejects any update where more than one registration links to the same accreditation (`validateAccreditationLinkUniqueness`), irrespective of status, with an error identifying the duplicate accreditation IDs and the registrations linked to them.
- Org replace now rejects any update where a registration's `accreditationId` references an accreditation not present in the organisation (`validateAccreditationLinkExists`).
- Org replace now rejects any update where a registration and its linked accreditation have mismatched identity fields — `wasteProcessingType`, `material`, site postcode (reprocessors only), `reprocessingType` if non-null on either side, and `glassRecyclingProcess` for glass (`validateAccreditationLinkMatches`).
- Incremental migration linking now tracks claimed accreditation IDs so an accreditation only becomes eligible for linking when it isn't already linked to a registration, and never links the same accreditation to two registrations.
- Fixed a related bug where glass registrations split into remelt/other siblings could incorrectly match and share a single accreditation, by including `glassRecyclingProcess` in the registration/accreditation identity key.

## Test plan
- [x] `npm run lint:fix`
- [x] `npm run format:check`
- [x] `npm run lint:types`
- [x] `npm run lint:types:tests` (no new errors introduced)
- [x] `npm run test` — 358 files / 6454 tests passing, 100% coverage (statements/branches/functions/lines)

[PAE-1566]: https://eaflood.atlassian.net/browse/PAE-1566?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/repositories/organisations/helpers.js
+++ b/src/repositories/organisations/helpers.js
@@ -8,7 +8,12 @@ import {
   assertAndHandleItemStateTransition,
   assertOrgStatusTransition
 } from '#repositories/organisations/schema/status-transition.js'
-import { validateApprovals } from './schema/helpers.js'
+import {
+  validateAccreditationLinkExists,
+  validateAccreditationLinkMatches,
+  validateAccreditationLinkUniqueness,
+  validateApprovals
+} from './schema/helpers.js'
 import { collateUsers } from './collate-users.js'
 import { getCurrentStatus } from './status.js'
 
@@ -93,6 +98,15 @@ function prepareRegAccForReplace(validated, existing) {
       validated.registrations,
       validated.accreditations
     )
+  validateAccreditationLinkUniqueness(validated.registrations)
+  validateAccreditationLinkExists(
+    validated.registrations,
+    accreditationsAfterUpdate
+  )
+  validateAccreditationLinkMatches(
+    validated.registrations,
+    accreditationsAfterUpdate
+  )
   validateApprovals(validated.registrations, accreditationsAfterUpdate)
   const registrations = updateStatusHistoryForItems(
     existing.registrations,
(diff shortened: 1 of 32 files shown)


src/repositories/organisations/helpers.js
```diff
@@ -93,6 +98,15 @@ function prepareRegAccForReplace(validated, existing) {
       validated.registrations,
       validated.accreditations
     )
+  validateAccreditationLinkUniqueness(validated.registrations)
```
dansweeting, line 101: This validates the full `validated.registrations` set on every replace, not just the entities being changed. The same path is used by the incremental migration's update flow (`organisation-persistence.js` calls the validated `replace`, not `replaceRaw`), so it applies to both admin edits and migration updates.

My concern is legacy data: orgs already persisted with two glass siblings linked to one accreditation (the behaviour this PR fixes) will fail `validateAccreditationLinkUniqueness` the next time they go through replace. The incremental migration only re-links orgs that receive new submissions, so an org with no new activity keeps its old links, and an admin editing it would get a 422.

Is a full re-migration or backfill planned alongside this deploy, or could edits start failing on orgs that still hold pre-existing duplicate links?

senthilinfy: Good question. https://eaflood.atlassian.net/browse/PAE-1565 already removed duplicate links in production. I am going to run another set of diagnostics using https://github.com/DEFRA/epr-backend/pull/1346. 

Expecting there should not be too invalid links. If there's any that gets flagged, I will be running migration(too many) or get Femi to fix(if just couple of registrations to fix)

dansweeting (approved):

PR #1350, opened 2026-06-18
PAE-1556 remove org linking logic from auth layer

crosslandwa:
Ticket: [PAE-1556](https://eaflood.atlassian.net/browse/PAE-1556)
## Description

Currently there is code in `getDefraUserRoles` (auth layer) that looks if the incoming path is for `POST /v1/organisations/{organisationId}/link` - if so logic runs to determine if a specific scope (`linker`) is granted - the underlying handler is guarded by this scope. This pattern couples the auth-layer, which should be generic, to a specific handler (which makes teasing apart roles/scopes more complex)

This PR removes that logic from the auth layer - instead the linking endpoint is guarded by the `inquirer` scope (always assigned for caller with Defra ID access token), and put checks for validity of linking attempt (user exists on org as `initial_user`, org not linked, org in right status) into the handler

Functional changes in

1. `src/routes/v1/organisations/link.js`
  - co-locate all linking logic
  - plugs hole where an org could be linked to empty id/name values (if the users Defra ID token was missing that data)
2. `src/common/helpers/auth/get-defra-user-roles.js`
  - removes linking logic from auth layer

The remainder of the PR is deleting (now) unused code

Requires update to journey tests: https://github.com/DEFRA/epr-backend-journey-tests/pull/538

---

[PAE-1556]: https://eaflood.atlassian.net/browse/PAE-1556?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/routes/v1/organisations/link.js
```diff
@@ -48,49 +45,61 @@ export const organisationsLink = {
...
+    const linkableOrganisations =
+      await organisationsRepository.findAllLinkableForUser(tokenPayload.email)
+
+    const orgToLink = linkableOrganisations.find(
+      (linkableOrganisation) => linkableOrganisation.id === organisation.id
+    )
+
+    if (!orgToLink) {
+      // strictly this should 403 if the org is linkable but the user is not an initial user...
+      // ...but that differentiation is embedded within findAllLinkableForUser
```
timrutter, line 85: we lose something here don't we?  we no longer know whether the problem is the org is not in a linkable state or whether the user is the initial user.
Could we not have a performFindAllLinkable which returns a list of orgs, and then  have function isInitialUser that looks in the org.users for the email.

timrutter (approved):
happy to approve this as is, my comment is optional

src/routes/v1/organisations/link.js
crosslandwa: Agree we've lost something here, and it would be nice if a 403 could be returned when the user is not an `initial_user`.

In practical terms though, hitting this endpoint as a non-initial user can't happen in the (happy path) user flow - this endpoint is hit when the user submits the `account/linking` page in the frontend - the linkable accounts shown on this page are driven by a call to `/v1/organisations/me`, which returns a list of linkable accounts

The linkable accounts are filtered by the following _rules_ (which are implemented in _both_ the `inmemory` and `mongo` implementations of the organisations repository `performFindAllLinkable`)
- is org in right status (`approved` or `active`)
- org is not already linked (`defraIdOrganisation === null`)
- user is an initial user

For me, that filtering logic belongs in the application layer, rather than in the repository layer. I'll draw up a followup PR that lifts it up...

(...but am aware that in lifting it up, the duplication will removed from the two repository implementations, but will exist in both the `/v1/me/organisations` and `/v1/organisations/{organisationId}/link` handlers, unless some sort of helper is drawn out)

PR #1353, opened 2026-06-19
PAE-1476: Add submissionDeclaredBy to report status transition

senthilinfy:
Ticket: [PAE-1476](https://eaflood.atlassian.net/browse/PAE-1476)
## Summary

- Adds `submissionDeclaredBy` field to the `updateReportStatus` repository method — required when transitioning to SUBMITTED, forbidden otherwise (via `Joi.when()`)
- Validates the field in both the route payload schema (`status.js`) and repository schema (`schema.js`): min 2 chars, max 255 chars, no `@#$%&<>` characters
- Stores `submissionDeclaredBy` on the `status.submitted` MongoDB slot alongside `at` and `by`
- Adds unit tests for all validation cases and updates existing SUBMITTED transition tests

[PAE-1476]: https://eaflood.atlassian.net/browse/PAE-1476?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/reports/repository/mongodb.js
+++ b/src/reports/repository/mongodb.js
@@ -191,7 +191,10 @@ const performUpdateReportStatus = async (db, params) => {
     validateUpdateReportStatus(statusParams)
 
   const now = new Date().toISOString()
-  const slotValue = { at: now, by: changedBy }
+  const slotValue =
+    submissionDeclaredBy !== undefined
+      ? { at: now, by: changedBy, declaredBy: submissionDeclaredBy }
+      : { at: now, by: changedBy }
 
   const doc = await reportsCollection(db).findOneAndUpdate(
     { id: reportId, version },
@@ -199,8 +202,7 @@ const performUpdateReportStatus = async (db, params) => {
       $set: {
         'status.currentStatus': status,
         'status.currentStatusAt': now,
-        [`status.${slot}`]: slotValue,
-        ...(submissionDeclaredBy !== undefined && { submissionDeclaredBy })
+        [`status.${slot}`]: slotValue
       },
       $push: /** @type {any} */ ({
         'status.history': { status, at: now, by: changedBy }
--- a/src/reports/repository/mongodb.js
+++ b/src/reports/repository/mongodb.js
@@ -192,9 +192,9 @@ const performUpdateReportStatus = async (db, params) => {
 
   const now = new Date().toISOString()
   const slotValue =
-    submissionDeclaredBy !== undefined
-      ? { at: now, by: changedBy, declaredBy: submissionDeclaredBy }
-      : { at: now, by: changedBy }
+    submissionDeclaredBy === undefined
+      ? { at: now, by: changedBy }
+      : { at: now, by: changedBy, declaredBy: submissionDeclaredBy }
 
   const doc = await reportsCollection(db).findOneAndUpdate(
     { id: reportId, version },
--- a/src/reports/repository/mongodb.js
+++ b/src/reports/repository/mongodb.js
@@ -187,10 +187,11 @@ const performUpdateReport = async (db, params) => {
  */
 const performUpdateReportStatus = async (db, params) => {
   const { slot, ...statusParams } = params
-  const { reportId, version, status, changedBy } =
+  const { reportId, version, status, changedBy, submissionDeclaredBy } =
     validateUpdateReportStatus(statusParams)
 
   const now = new Date().toISOString()
+  const slotValue = { at: now, by: changedBy }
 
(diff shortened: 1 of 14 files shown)


src/reports/routes/status.js
```diff
@@ -26,7 +26,12 @@ const payloadSchema = Joi.object({
       REPORT_STATUS.SUBMITTED
     )
     .required(),
-  version: Joi.number().integer().min(1).required()
+  version: Joi.number().integer().min(1).required(),
+  submissionDeclaredBy: Joi.string().when('status', {
+    is: REPORT_STATUS.SUBMITTED,
+    then: Joi.required(),
+    otherwise: Joi.forbidden()
+  })
```
dansweeting: Do we have another frontend/backend ordering problem to solve if we make this required here? 

How might we deploy one, then the other, without having to deploy both in quick succession?

Perhaps if this schema item stays as optional, until we are satisfied the FE is capturing it (with the inevitable change on the port-schema too)? Once we are happy the FE is providing this, we can tighten up the schema so it's required, which would be accurate from that point on - at which point we can add back the test for 422 error

src/reports/routes/status.js
```diff
@@ -26,7 +26,12 @@ const payloadSchema = Joi.object({
       REPORT_STATUS.SUBMITTED
     )
     .required(),
-  version: Joi.number().integer().min(1).required()
+  version: Joi.number().integer().min(1).required(),
```
dansweeting, line 33: Small thing on the validation: the PR description mentions min 2 / max 255 chars and blocking `@#$%&<>`, but this is a bare `Joi.string()`, so none of those are enforced. In particular there's no upper bound, so an arbitrarily long value would be accepted and stored. Was the intent to add `.max()` / `.pattern()` here, or should the description be updated?

src/reports/repository/mongodb.js
```diff
@@ -187,18 +187,20 @@ const performUpdateReport = async (db, params) => {
...
 
   const doc = await reportsCollection(db).findOneAndUpdate(
     { id: reportId, version },
     {
       $set: {
         'status.currentStatus': status,
         'status.currentStatusAt': now,
-        [`status.${slot}`]: { at: now, by: changedBy }
+        [`status.${slot}`]: slotValue,
+        ...(submissionDeclaredBy !== undefined && { submissionDeclaredBy })
```
dansweeting: This writes `submissionDeclaredBy` at the document root rather than into the `status.submitted` slot (which stays `{ at, by }`), and the history `$push` doesn't include it either. A couple of knock-on effects worth a sanity check:

- It's overwritten on each submit and isn't cleared on unsubmit (the field is `forbidden` on that transition, so the guard just skips it), so a report that's been unsubmitted still carries the previous declarant's name at root.
- The declaration isn't captured in the immutable history.

For what's effectively a legal attestation, would it be better tied to the submitted slot or the history entry? Also flagging that the PR description says it's stored "on the `status.submitted` slot alongside `at` and `by`", which doesn't quite match what's here.

src/reports/repository/schema.js
```diff
@@ -203,7 +203,12 @@ export const updateReportStatusSchema = Joi.object({
   status: Joi.string()
     .valid(...Object.values(REPORT_STATUS))
     .required(),
-  changedBy: userSummarySchema.required()
+  changedBy: userSummarySchema.required(),
+  submissionDeclaredBy: Joi.string().when('status', {
```
dansweeting: Minor: `Joi.string()` rejects empty strings by default but accepts a single character, so the "min 2 chars" rule from the description isn't applied here. If a minimum is wanted, a `.min(2)` would cover it.

src/reports/repository/mongodb.js
senthilinfy: Sorry I made some refactor's post PR creation, and forgot to update the description.

Good question. I did have the declaration at status.submitted level initially, but changed it as it didn't feel like the right place and changed it.  Since you mention unsubmit may be its worth going back to my approach original approach. Let me think about it. Also once CMA is live storing it in `status.submitted` will make it immutable, right unsubmit/submit makes it mutable

src/reports/routes/status.js
senthilinfy: Will make it as optional in BE and mandatory in FE. The plan is

-  Deploy BE(can store new declaredBy field optionally)
-  Deploy FE
-  Deploy BE to make declaredBy mandatory ob submission

dansweeting (approved):

PR #1354, opened 2026-06-19
PAE-1590: Establish committed row-state foundation behind an off-by-default feature flag

GraemeF:
Ticket: [PAE-1590](https://eaflood.atlassian.net/browse/PAE-1590)
## What this brings

Establishes the ADR-0037 Stage 1 waste-record-state foundation on `main` as production-track work, gated behind a feature flag that is **off by default in every environment**. With the flag off, summary-log submission behaviour is byte-identical to today — no new-format documents are written — so the foundation can land ahead of the legacy backfill without ever leaving the collection holding partial new-format data.

### The foundation (lifted from [#1339](https://github.com/DEFRA/epr-backend/pull/1339), never merged)

- Waste-record-state collection, owned by the waste-records module and homed vertically alongside its read model (`waste-records/repository/`): port, schema, validation, in-memory and MongoDB adapters, and cross-implementation contracts for `upsertRowStates` / `findBySummaryLogId` / `findRowHistory`. The balance stream is the producer — it writes into the collection through an injected repository the record side owns and registers, so the balance flow no longer constructs a foreign collection.
- Row classification: `classifyWasteRecord` pairs each waste record with its waste-balance classification (outcome, reasons, contribution), and `getTargetAmount` now derives the credit from that classification — a behaviour-preserving refactor of the previous per-record computation. The classification is co-stored on each waste-record-state document, and the stored schema validates each reason `code` against the `CLASSIFICATION_REASON` set the classifier emits — strict, exactly as `outcome` and `wasteRecordType` are — so an out-of-set code is rejected on write.
- Read layer: `wasteRecordStatesForRegistration` (waste-records application) resolves a registration's waste record state at its current head and projects it to a clean `WasteRecordState` domain type — dropping the `summaryLogIds` membership, the storage `id` and the redundant partition at the storage↔domain seam, so the later read cutovers bind to domain content rather than storage shape. The head itself comes from the balance stream via `latestCommittedSummaryLogId`, which stays balance-owned; the storage port still returns the raw `RowState`. No consumer reads it yet.
- Write hook in `update-via-stream.js`: persists each row's committed state ahead of the stream event so a failed append leaves no committed balance change.

### The feature flag

- `wasteRecordStates` (`FEATURE_FLAG_WASTE_RECORD_STATES`, accessor `isWasteRecordStatesEnabled()`), default `false` in convict, added across the existing feature-flag adapter trio (port / config / in-memory).
- Gates **only** the `upsertRowStates` call in the submission write hook. Classification, credit-total arithmetic and the stream event append are unchanged, so the flag-off path is exactly today's behaviour. The flag threads from the waste-balances plugins (sourced from `server.featureFlags`) through the repository and helpers into the write hook.
- Both paths are covered by tests: flag off ⇒ no row-state writes and an identical event and closing balance; flag on ⇒ waste record states written.

### Journey tests both ways

- The journey suite runs twice in CI (`waste-record-states: [false, true]`), proving submission behaviour is identical with the flag off and on. Each leg segregates its artefacts by suffix and a gate job aggregates both results. The companion action input lands in [epr-backend-journey-tests#525](https://github.com/DEFRA/epr-backend-journey-tests/pull/525).

## Intentionally not in this first PR

PAE-1590 also covers the work that makes the new record safe to switch on; it lands in later PRs under this ticket, before the flag is enabled:

- Monitoring that compares the new waste-record-state record against the existing committed state across the estate.
- A backfill that reconstructs every historical submission's committed rows from per-row version history.
- An optimistic-concurrency guard on the `upsertRowStates` write path — today's dedup is a non-atomic find-then-insert with no unique index, so concurrent submissions for the same row could write duplicates; a unique-index / insert-contention guard (the same family as the waste-balance event stream's slot guard) closes that.
- Switching the flag on, once monitoring is clean over the backfilled estate.


src/waste-balances/repository/mongodb.test.js
```diff
@@ -68,6 +85,9 @@ describe('MongoDB waste balances repository', () => {
       await database
         .collection(WASTE_BALANCE_EVENTS_COLLECTION_NAME)
         .deleteMany({})
+      await database
+        .collection(WASTE_BALANCE_ROW_STATES_COLLECTION_NAME)
+        .deleteMany({})
```
dansweeting, line 90: We have this test tidying up the new collection 👍 

We also need to include the same collection in the non-prod data cleanup endpoint we have, I think under the /dev only route?

dansweeting (approved):

PR #1364, opened 2026-06-22
PAE-1590: Reconstruct row states from history to support the reconciliation diagnostic

GraemeF:
## What this adds

The reconstruction machinery for the waste record state collection (ADR-0037): `reconstructSubmissionRowStates` rebuilds a submission's row states in memory from the sparse `versions[]` history, and `backfillRegistrationRowStates` / `backfillEstateRowStates` walk a registration and the whole estate through that reconstruction.

## It writes nothing

This code is not wired into startup and performs no database writes. It lands as library support for the read-only, in-memory reconciliation diagnostic that follows — which reconstructs row states and compares them against the legacy committed baseline without persisting anything. The persisting migration runs later as a deliberate step, once the diagnostic has been reviewed, mirroring the waste-balance ledger cutover.

Linked to [PAE-1590](https://eaflood.atlassian.net/browse/PAE-1590).


src/waste-records/backfill/backfill-estate-rowstates.js
```diff
@@ -0,0 +1,217 @@
...
+  organisationsRepository,
+  wasteRecordsRepository,
+  summaryLogsRepository,
+  overseasSitesRepository,
+  rowStateRepository
+}) => {
+  const { accreditationId } = registration
+  if (!accreditationId) {
+    return null
+  }
```
dansweeting, line 95: This guard skips registered-only streams, same as the live path which gates the row-state write + event append on `if (accreditationId)` at `sync-from-summary-log.js:331`.

But doesn't ADR-0037 pull registered-only into scope? It says registered-only submissions "must also append events" on the `(registrationId, null)` partition, and lists "registered-only submissions join the ledger" as a win. The schema already allows `accreditationId: null` for exactly this, the `versions[]` source data exists for them, and PAE-1590's AC just says "all historical submissions".

My worry is the stage-2 readers that serve registered-only today (CSV/FOI export, report detail) will have nothing to read once the old path retires.

src/waste-records/backfill/reconstruct-submission-rowstates.test.js
```diff
@@ -0,0 +1,189 @@
...
+
+    const upserts = reconstructSubmissionRowStates({
+      wasteRecords,
+      summaryLogs,
+      accreditation,
+      overseasSites
+    })
+
+    expect(upserts[0].entries.map((e) => e.rowId)).toEqual(['row-1'])
+    expect(upserts[1].entries.map((e) => e.rowId)).toEqual(['row-1'])
```
dansweeting, line 137: I had to look twice here to appreciate that `upserts[0]` represents `s1-1` and `[1]` represents `sl-2`, might be clearer that we assert the summary logs in here too for this test?

src/waste-records/backfill/backfill-estate-rowstates.js
```diff
@@ -0,0 +1,217 @@
...
+
+  let accreditation = null
+  try {
+    accreditation = await organisationsRepository.findAccreditationById(
+      organisation.id,
+      accreditationId
+    )
+  } catch {
+    accreditation = null
+  }
```
dansweeting: This treats any throw from `findAccreditationById` as an orphaned accreditation. That puts a transient DB blip in the same bucket as "accreditation genuinely gone", so a stream could get silently recorded as orphaned and skipped with no retry signal, on a run where completeness is the point. Worth narrowing to the not-found case and letting unexpected errors propagate, so orphans and failures stay distinct.

src/waste-records/backfill/backfill-estate-rowstates.js
GraemeF: Good catch, and agreed — registered-only is in ADR-0037 scope. This PR is deliberately just the accredited reconstruction engine; the registered-only / no-accreditation backfill is tracked separately under PAE-1590 and is wired to land **before** the write flag flips, so the full population is reconstructed before any reads cut over.

On the stage-2 worry: the registered-only read-head question — the `(registrationId, null)` partition forms no `summary-log-submitted` head today, so the current read path returns nothing for it — is also tracked separately and will be settled before registered-only reads move onto the row-state collection. A real decision, but not a blocker for this engine landing dark behind the off flag. Happy to walk through the sequencing if useful.

src/waste-records/backfill/backfill-estate-rowstates.js
GraemeF: Narrowed in 31a8b08f: the orphaned-accreditation skip is now gated on Boom 404 (`isBoom && statusCode === NOT_FOUND`) only — a genuinely missing org/accreditation. Any other lookup error (e.g. a consistency-timeout 500) now propagates and fails the run, so a transient blip can no longer be silently recorded as orphaned. Added a TDD test that asserts an unexpected (non-404) throw propagates.

src/waste-records/backfill/reconstruct-submission-rowstates.test.js
GraemeF: Addressed in 31a8b08f: the test now asserts the summary log id on each upsert explicitly — `upserts[0].summaryLogId` is `sl-1` and `upserts[1].summaryLogId` is `sl-2` — so the ordering is self-evident rather than something you have to infer from the index.

dansweeting (approved):

PR #1367, opened 2026-06-24
PAE-1654: Emit numeric CSV export columns as numbers

GraemeF:
Ticket: [PAE-1654](https://eaflood.atlassian.net/browse/PAE-1654)

Makes the backend's numeric CSV and JSON output genuinely numeric: the waste-records export emits real numbers for numeric columns, and the report-submissions endpoint returns numbers rather than String-coerced text. Identifier columns stay text. The waste-records export also gains the formula-injection sanitisation it previously lacked.

**Waste-records export** (`src/waste-records-export/`)
- `domain/csv-columns.js`: data cells pass their stored value through, so genuine numbers stay numeric; identifier metadata (Row ID, registration and accreditation numbers) stays text.
- `sanitiseFormulaInjection` prefixes any string cell beginning with `=`, `+`, `-` or `@` with an apostrophe so spreadsheet software treats it as literal text; numeric cells are untouched.
- `application/stream-csv-export.js`: dropped `quoteColumns: true`, so fast-csv's default selective quoting applies — numbers serialise bare, text containing the delimiter, a quote or a newline is still quoted.

**Report submissions** (`src/reports/`)
- `application/report-submissions.js`: `formatTonnage` and `sumSentOn` return real numbers, or `''` when there is no value.
- `routes/submissions.js`: the response schema's sixteen numeric tonnage fields now accept a number or `''`; the free-text note stays a string. The only consumer of this endpoint is the admin CSV download (DEFRA/epr-re-ex-admin-frontend#445).

[PAE-1654]: https://eaflood.atlassian.net/browse/PAE-1654?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/reports/routes/submissions.js
```diff
@@ -41,22 +46,22 @@ export const getReportSubmissions = {
...
+              tonnageReceivedForRecycling: tonnageValue,
+              tonnageRecycled: tonnageValue,
+              tonnageExportedForRecycling: tonnageValue,
+              tonnageSentOnTotal: tonnageValue,
+              tonnageSentOnToReprocessor: tonnageValue,
+              tonnageSentOnToExporter: tonnageValue,
+              tonnageSentOnToOtherFacilities: tonnageValue,
+              tonnagePrnsPernsIssued: tonnageValue,
+              freeTonnagePrnsPerns: tonnageValue,
+              totalRevenuePrnsPerns: tonnageValue,
```
crosslandwa: I don't think `totalRevenuePrnsPerns` and `averagePrnPernPricePerTonne` are a `tonnageValue` - they're monetary amounts I believe - perhaps extract both

```
const tonnageValue = Joi.alternatives()
  .try(Joi.number(), Joi.valid(''))
  .required()

const monetaryValue = Joi.alternatives()
  .try(Joi.number(), Joi.valid(''))
  .required()
```

GraemeF: ooopsie

crosslandwa (approved):

PR #1372, opened 2026-06-25
PAE-1647: Clear type errors in received-loads-for-export test

dansweeting:
## What

Clears all 66 pre-existing test-typecheck errors in `received-loads-for-export.test.js`, bringing the file to zero under `jsconfig.typecheck.tests.json`.

The changes are test-only. No production code is touched and no real test behaviour changes.

- **Joi error access**: guarded `error.details[0].message` reads with optional chaining (`error?.details[0]`), since the destructured `error` from `validate()` is typed as possibly undefined.
- **Accreditation fixtures**: the test's accreditation objects are partial mocks. Imported the `Accreditation` type and cast each fixture to it, so they satisfy the `classifyForWasteBalance` parameter type. The empty-`statusHistory` fixture gets an explicit element type.
- **INCLUDED result access**: cast the classification result to the INCLUDED-branch shape when reading `transactionAmount`, which only exists on that member of the union return type.
- **Required-field delete**: typed the cloned row as `Record<string, any>` so deleting a required field is permitted.
- **Removed an unreachable test**: the "accreditation is undefined" case could only be satisfied with an `any` cast. Every production caller types `accreditation` as `Accreditation | null`, so `undefined` cannot reach this function, and the adjacent "null" test already covers the same branch. Removed it rather than widening the production type to accept an impossible value (see inline comment for the full trace).

## Why

These errors surfaced while touching this file for PAE-1647 and were scoped out of that bug fix to keep it minimal. The team rule is to clear type errors in any file we touch, so this file is marched to zero in its own change.

### DIFF
--- a/src/domain/summary-logs/table-schemas/exporter/received-loads-for-export.test.js
+++ b/src/domain/summary-logs/table-schemas/exporter/received-loads-for-export.test.js
@@ -885,7 +885,10 @@ describe('RECEIVED_LOADS_FOR_EXPORT', () => {
         })
         expect(result.outcome).toBe(ROW_OUTCOME.INCLUDED)
         expect(result.reasons).toEqual([])
-        expect(result.transactionAmount).toBe(60.5)
+        expect(
+          /** @type {{ transactionAmount: number }} */ (result)
+            .transactionAmount
+        ).toBe(60.5)
       })
 
       it('rounds transaction amount to two decimal places', () => {
@@ -897,7 +900,10 @@ describe('RECEIVED_LOADS_FOR_EXPORT', () => {
           accreditation,
           overseasSites: ORS_VALIDATION_DISABLED
         })
-        expect(result.transactionAmount).toBe(60.56)
+        expect(
+          /** @type {{ transactionAmount: number }} */ (result)
+            .transactionAmount
+        ).toBe(60.56)
       })
     })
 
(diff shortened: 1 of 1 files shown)


src/domain/summary-logs/table-schemas/exporter/received-loads-for-export.test.js
```diff
@@ -1101,16 +1112,7 @@ describe('RECEIVED_LOADS_FOR_EXPORT', () => {
       })
     })
 
-    describe('INCLUDED outcome - undefined or null accreditation', () => {
-      it('returns INCLUDED when accreditation is undefined (accreditation check passes)', () => {
```
dansweeting, line 1105: Removed this test rather than satisfying its type error with a cast.

Tracing every production caller of `classifyForWasteBalance` — `isIncludedInWasteBalance`, the `ClassificationContext` in `period-status.js`, `validate.js` (via `registration.accreditation`), and the shared `TableSchema` contract in `table-schemas/index.js` — `accreditation` is always typed `Accreditation | null`. `undefined` can never reach this function in production.

The adjacent "accreditation is null" test already exercises the same `if (!accreditation)` branch, so this case was both impossible and redundant. The type checker flagging it was correct: the right fix is to delete the unreachable test, not to widen the production type to accept a value that cannot occur.

GraemeF (approved):

PR #1386, opened 2026-06-26
PAE-1590: Emit summary-log submitted events for registered-only registrations

GraemeF:
Ticket: [PAE-1590](https://eaflood.atlassian.net/browse/PAE-1590)
## What

Registered-only registrations (those with no accreditation) submit summary logs, but no `SUMMARY_LOG_SUBMITTED` event was being written to their waste-balance stream. The read model resolves which submissions belong to a registration from those events, so registered-only row states were being written but could not be read back.

This change writes a balance-neutral zero-delta summary-log submitted event (`creditTotal: 0`) for these registrations so their row states become readable.

## How

Behind a dedicated `FEATURE_FLAG_REGISTERED_ONLY_SUBMITTED_EVENTS` flag (default off), independent of the row-state write flag:

- **Forward path** — when the flag is on, a registered-only submission writes its summary-log submitted event as part of committing its state.
- **Historical migration** — the backfill sweep writes one summary-log submitted event per existing registered-only submission. It is idempotent: it never writes a second event for a submission that already has one, so the sweep is safe to re-run.

A registered-only registration has no accreditation and therefore no waste balance, so the event carries `creditTotal: 0` and leaves all balances unchanged. Balance calculations only consider accredited registrations, so this never surfaces a balance for a registered-only registration.

## Notes

- The sweep is wired at `onPostStart` under the server lock (the standard startup-sweep pattern), so it runs automatically on every deploy — a read-only dry-run when the flag is off, executing the historical backfill when it is on. It is idempotent, so it is safe on every restart.
- The flag defaults off, but off is not inert: on each startup the dry-run logs, per registration, the event it would emit, giving the reviewable production evidence the pre-flip review gates on. No stream events are written in any environment until the flag is flipped on — only dry-run log lines appear before then.
- Related row-state rollout work is tracked separately under PAE-1590.


src/application/waste-records/sync-from-summary-log.js
```diff
@@ -298,6 +304,23 @@ const commitStateAndBalance = async ({
       overseasSites,
       summaryLogId: summaryLog.file.id
     })
+  } else if (featureFlags?.isRegisteredOnlySubmittedEventsEnabled()) {
+    await wasteBalanceService.submitSummaryLog(
```
dansweeting, line 308: Forward-path idempotency: the backfill guards every write with `existingSubmittedEventSummaryLogIds`, but this forward path calls `appendRegisteredOnlySubmittedEvent` unconditionally with no existing-event check. On an at-least-once SQS redelivery, if the process dies between the append and the `SUBMITTING` status transition, we replay `sync` and write a second zero-delta `SUMMARY_LOG_SUBMITTED` into the null partition. This is probably fine: the backfill doesn't go through queue delivery and can only ever emit zero deltas anyway, so the blast radius is a duplicate zero-tonnage event rather than any balance corruption. Still worth confirming whether the forward path wants the same guard, or noting that the accredited path already tolerates this same crash window so reviewers don't assume the two paths differ in safety.

src/application/waste-records/sync-from-summary-log.js
```diff
@@ -298,6 +304,23 @@ const commitStateAndBalance = async ({
...
+        organisationId: summaryLog.organisationId
+      },
+      { summaryLogId: summaryLog.file.id, creditTotal: 0 },
+      {
+        id: user.id,
+        ...(user.name && { name: user.name }),
+        email: user.email
+      }
+    )
+  } else {
```
dansweeting, line 321: This `else` is empty apart from a comment and is a likely SonarCloud S108 hit. Suggest dropping it and moving the note to a comment above the `if`.

src/waste-records/backfill/reconstruct-submission-rowstates.js
```diff
@@ -97,6 +104,11 @@ export const reconstructSubmissionRowStates = ({
         projectRowState({ ...record, data }, accreditation, overseasSites)
       ]
     })
-    return { summaryLogId: log.id, entries }
+    return {
+      summaryLogId: log.id,
+      entries,
+      submittedAt: log.submittedAt,
```
dansweeting, line 111: (Comment relates to the `.sort` by `submittedAt` just above, at line ~89, which is outside the diff.) Sorting by `submittedAt` alone leaves ties in undefined order, which shifts `seenSummaryLogIds` attribution between runs. Pre-existing, so not a blocker, but worth adding `id` as a stable secondary key unless `submittedAt` is guaranteed unique per registration.

src/application/waste-records/sync-from-summary-log.js
GraemeF: Good spot — confirmed the two paths don't differ in safety. The accredited path has the same crash window: `updateWasteBalanceTransactions` → `performUpdateViaStream` ends in the same unguarded ledger append. Both are duplicate-tolerant by construction because the fold is delta-based — a submission's `creditTotal` is applied as a delta against the ledger's running credit total, so a redelivered submission with unchanged records appends a zero-delta event: audit noise rather than balance movement (and the registered-only event is zero-delta by definition). Guarding only this new branch would make the two paths diverge in behaviour, so we're keeping the parity as-is.

src/application/waste-records/sync-from-summary-log.js
GraemeF: The empty else is deliberate — dropping it trips SonarCloud S126 (if…else if must end with else); S108 doesn't fire because the block carries the comment. We learnt this the hard way: a merge cleanup dropped it and the quality gate failed, hence the restore commit. It's also temporary — once the flag is permanently enabled the whole branch structure collapses to the unconditional append, and the flag removal is tracked separately, so we'd rather not reshape it in the meantime.

src/waste-records/backfill/reconstruct-submission-rowstates.js
GraemeF: Agreed — the sort is pre-existing on main (this file only gained the provenance fields in this PR), but the hardening is cheap so we'll add `id` as a stable secondary key on this branch.

dansweeting (approved):

PR #1387, opened 2026-06-26
PAE-1669: Wire optimistic concurrency through the waste balance stream

GraemeF:
Ticket: [PAE-1669](https://eaflood.atlassian.net/browse/PAE-1669)

The waste balance stream enforces optimistic concurrency with a unique slot index on `(registrationId, accreditationId, number)`: the append primitive rejects any event whose `number` isn't `head + 1`. That guard could never fire, because `appendToStream` derived the slot number from a head it read itself immediately before writing — so the number was `head + 1` by construction on every call and a competing write could never collide. The concurrency control existed in the primitive but every caller bypassed it.

This wires the caller's read position through to the append. `findBalance` now exposes `eventNumber` (the latest event's position), the PRN and summary-log write paths thread that position back as `expectedHead`, and `appendToStream` derives the slot as `expectedHead + 1`. A write whose head moved since the caller read it now targets an occupied slot, so the append fails with a `StreamSlotConflictError` and the conflict surfaces to the caller instead of silently committing from a stale position.

The conflict propagates rather than being retried in process, because re-applying a decision the race has already invalidated — a stale sufficiency check, or a frozen summary-log `creditTotal` — is exactly the double-count this guard exists to prevent. Re-evaluation belongs at the natural boundary: a summary-log submission runs in the worker job, so the conflict fails the job and the queue redelivers against fresh state; a PRN transition surfaces the conflict to the client.

On the PRN status path the stream effect is now applied before the document write, so a losing racer fails at the slot and never persists a partial document change.

New concurrency tests drive two writes against a single partition through the application path and assert exactly one wins, the loser surfaces a `StreamSlotConflictError`, and the stream head, persisted balance, and PRN document all agree on the single committed transition.

A follow-up will move the balance business logic out of `WasteBalancesRepository` into the application layer — a pure `decide` over a folded aggregate (functional core, imperative shell) — so the concurrency guarantee holds by construction rather than resting on the slot index to catch a divergent read, and the mis-layered repository methods are retired. That refactor is tracked separately and deliberately kept out of this PR to keep the change reviewable.

[PAE-1669]: https://eaflood.atlassian.net/browse/PAE-1669

### DIFF
--- a/src/packaging-recycling-notes/application/update-status-balance-effects.js
+++ b/src/packaging-recycling-notes/application/update-status-balance-effects.js
@@ -69,7 +69,8 @@ export async function deductWasteBalanceIfNeeded(
       organisationId,
       prnId,
       tonnage,
-      createdBy
+      createdBy,
+      expectedHead: balance.eventNumber
     })
   } else {
     throw Boom.badRequest(
@@ -112,7 +113,8 @@ export async function deductTotalBalanceIfNeeded(
       organisationId,
       prnId,
       tonnage,
-      createdBy
+      createdBy,
+      expectedHead: balance.eventNumber
     })
   } else {
     throw Boom.badRequest(
@@ -152,7 +154,8 @@ export async function creditWasteBalanceIfNeeded(
       organisationId,
       prnId,
       tonnage,
-      createdBy
+      createdBy,
+      expectedHead: balance.eventNumber
     })
   } else {
     throw Boom.badRequest(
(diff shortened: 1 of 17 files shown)


timrutter (approved):
Claude and i both happy.  Claude says its diverged from the ticket slightly but in a better way:
`One deliberate deviation worth your attention (this is the headline review point, not a defect): the ticket's Proposed fix says "On rejection, re-read the latest balance and retry." The PR consciously does not retry in-process — it propagates the conflict (summary-log conflict fails the worker job → queue redelivers against fresh state; PRN conflict surfaces to the client). The PR body argues, correctly in my view, that an in-process retry would re-apply a decision the race already invalidated (a stale sufficiency check or a frozen creditTotal) — exactly the double-count the guard exists to prevent. So the implementation diverges from the ticket's suggested mechanism but better serves its intent. If the ticket author specifically wanted in-process retry, that's a conversation to have, but I'd back the PR's choice.
`

PR #1388, opened 2026-06-26
PAE-1663: derive report period status as a first-class api field

loveboat:
Ticket: [PAE-1663](https://eaflood.atlassian.net/browse/PAE-1663)

derive report due/overdue status in the backend instead of the frontend (ADR-0038, first slice).

- backend: additive `periodStatus` on the reports calendar (due/overdue/in_progress/ready_to_submit/submitted), derived on read; response schema enforces it non-null
- frontend: consume `periodStatus`, drop the local due/overdue date arithmetic and its defensive null path

[PAE-1663]: https://eaflood.atlassian.net/browse/PAE-1663?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/reports/routes/get.js
```diff
@@ -89,6 +94,7 @@ export const reportsGet = {
 
     const reportingPeriods = merged.map((period) => ({
       ...period,
+      periodStatus: derivePeriodStatus(period),
```
dansweeting, line 98: Flag, not a blocker: `derivePeriodStatus` can return `null`, but the schema makes `periodStatus` required non-null. Safe today and under the CPA follow-up, since every period in the calendar is either ended or carries a report, so `null` never reaches the schema.

The catch is that invariant ("no not-ended period without a report") spans the merge output, `generateReportingPeriods`, and the schema, with nothing local saying so. It'd only break if we later surface the current open period before its end date. No change needed now; a one-line comment at this call would close it.

dansweeting (approved):

src/reports/routes/get.js
loveboat: comment added in commit: https://github.com/DEFRA/epr-backend/pull/1388/changes/34bed78f5a7751f42a7df61421d21e08444db612

PR #1396, opened 2026-06-29
PAE-1659: emit one row per submitted report in report-submissions feed (+ submission number)

dansweeting:
Ticket: [PAE-1659](https://eaflood.atlassian.net/browse/PAE-1659)

## What

The regulator report-submissions feed (`GET /v1/organisations/reports/submissions`) emits, per reporting period, a submitted date, a submitter and a full tonnage/PRN breakdown. It was built on `mergeReportingPeriods`, which exposed only the period's `current` report, and produced exactly one row per reporting period.

This change makes the feed emit **one row per submitted report**. A period submitted once still has one row; a resubmitted period fans out into one row per submission, each carrying its own submitted date, submitter, tonnage/PRN figures and a new `SubmissionNumber` column. A period with nothing submitted keeps a single blank row, so outstanding periods stay visible.

## Why

Two problems, both surfacing once report resubmission for closed periods is active:

1. **In-flight drafts masked submitted figures.** Because the feed read `current`, an in-progress submission 2 draft (no submitted date) sitting above an already-submitted submission 1 blanked the row's submitted date, submitter and every tonnage/PRN figure for the duration of the correction. The feed now sources every field from a submitted report, so an in-flight draft never masks the last submitted figures.
2. **Superseded submissions were dropped.** One row per period showed only the latest submitted report, so once a correction was submitted the original figures vanished with no trace that the period had been resubmitted. Emitting one row per submitted report keeps the full submission history visible, and the `SubmissionNumber` column identifies each (a period with rows up to N was submitted N times, i.e. N-1 resubmissions). Agreed with the regulator on the ticket.

Resubmission is gated by the `FEATURE_FLAG_CLOSED_PERIOD_ADJUSTMENTS` feature flag: enabled in the test environment, disabled in production. A period therefore cannot yet hold more than one submission in prod (the fan-out is a no-op there today), but the behaviour is exercisable end-to-end in the test environment ahead of the production rollout.

## Changes

- `merge-reporting-periods.js`: add a `selectSubmittedReports` helper returning every submitted report for a period, ascending by `submissionNumber`, and expose it as `submittedReports` on each merged period alongside the unchanged `report` (the current report, which may be an in-flight draft). Inclusion keys on a retained `submittedAt` rather than current status, so a period that was submitted and later unsubmitted (status reverts to ready_to_submit, submitted instant retained) still appears rather than blanking; an in-flight draft (no `submittedAt`) is excluded. Selection is by `submissionNumber`, independent of slot ordering. A JSDoc note warns that `report` may be an unsubmitted draft and steers regulator output to `submittedReports`, and records why this stays distinct from `build-calendar-periods`' status-based `latestSubmitted` (resubmission flagging).
- `report-submissions.js`: fan out each period to one row per `submittedReports` entry (a single blank row when none), and source the submitted date, submitter, `submissionNumber` and all tonnage fields from that one report so each row is internally consistent.
- `routes/submissions.js`: allow `submissionNumber` (a number, or blank) in the per-row response schema. The schema is per-row, so the fan-out needs no other change.
- `routes/get.js` (reports calendar): the calendar response spreads the whole merged period, which would otherwise leak the feed-only `submittedReports` field and fail the strict reporting-period response schema, so it strips that field before spreading the rest. The calendar still curates the current report via `toReportListItem` and is otherwise unchanged.
- Tests: the resubmission scenarios are seeded through real repository adapters (a `seedInFlightResubmission` helper for the draft-over-submitted case, and a `buildUnsubmittedReport` helper that submits then unsubmits to pin retain-on-unsubmit). Feed tests assert the fan-out (a resubmitted period yields one row per submission, ascending), that an in-flight draft adds no row, and no regression for single-submission and outstanding periods. Domain tests pin the ascending submitted-report list, order-independence and retention after unsubmit.

## Scope

The public-register half (PAE-1660) shared the original backend change but has been split into its own ticket: its requirements have grown (it may need both an original and a resubmission date per period) and it is pending a regulator decision, so it is not included here. It will add back whatever single-latest projection it needs when a caller exists.

Documented in ADR-0039, section "Impact on reports and consumers".

[PAE-1659]: https://eaflood.atlassian.net/browse/PAE-1659?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/reports/application/report-submissions.js
+++ b/src/reports/application/report-submissions.js
@@ -51,6 +51,7 @@ import {
  * @property {string} dueDate
  * @property {string} submittedDate
  * @property {string} submittedBy
+ * @property {number | ''} submissionNumber
  */
 
 /** @typedef {SubmissionBaseFields & TonnageFields} ReportSubmissionsRow */
@@ -132,8 +133,7 @@ function buildTonnageFields(report) {
  * @param {string} cadence
  * @param {object} mergedPeriod
  * @param {string} accreditationNumber
- * @param {string} submittedDate
- * @param {string} submittedBy
+ * @param {import('#reports/repository/port.js').ReportSummary | null} report
  * @returns {ReportSubmissionsRow}
  */
 function buildRow(
@@ -142,8 +142,7 @@ function buildRow(
   cadence,
   mergedPeriod,
   accreditationNumber,
-  submittedDate,
-  submittedBy
+  report
 ) {
   return {
     regulator: REGULATOR_DISPLAY[registration.submittedToRegulator],
@@ -169,9 +168,13 @@ function buildRow(
       mergedPeriod.year
     ),
     dueDate: mergedPeriod.dueDate,
-    submittedDate,
-    submittedBy,
-    ...buildTonnageFields(mergedPeriod.report)
+    // Every report-derived field describes this one submitted report, so the
+    // row stays internally consistent. An in-flight draft is never a `report`
+    // here, so it produces no row until it is itself submitted.
+    submittedDate: report?.submittedAt?.slice(0, 10) ?? '',
+    submittedBy: report?.submittedBy?.name ?? '',
+    submissionNumber: report?.submissionNumber ?? '',
+    ...buildTonnageFields(report)
   }
 }
 
@@ -208,16 +211,22 @@ async function buildSubmissionRows(
       cadence
     )
 
-    return merged.map((mergedPeriod) => {
-      const report = mergedPeriod.report
-      return buildRow(
-        org,
-        registration,
-        cadence,
-        mergedPeriod,
-        accreditationNumber,
-        report?.submittedAt?.slice(0, 10) ?? '',
-        report?.submittedBy?.name ?? ''
+    return merged.flatMap((mergedPeriod) => {
+      // One row per submitted report, so a resubmitted period fans out. A period
+      // with nothing submitted still gets a single blank row (`null`), keeping
+      // outstanding periods visible.
+      const reports = mergedPeriod.submittedReports.length
+        ? mergedPeriod.submittedReports
+        : [null]
+      return reports.map((report) =>
+        buildRow(
+          org,
+          registration,
+          cadence,
+          mergedPeriod,
+          accreditationNumber,
+          report
+        )
       )
     })
   })
--- a/src/reports/application/report-submissions.js
+++ b/src/reports/application/report-submissions.js
@@ -3,7 +3,10 @@
 
 import { CADENCE } from '#reports/domain/cadence.js'
 import { generateReportingPeriods } from '#reports/domain/generate-reporting-periods.js'
-import { mergeReportingPeriods } from '#reports/domain/merge-reporting-periods.js'
+import {
+  mergeReportingPeriods,
+  selectSubmittedReports
+} from '#reports/domain/merge-reporting-periods.js'
 import { formatMaterial, capitalize } from '#common/helpers/formatters.js'
 import { formatPeriodLabel } from '#reports/domain/period-labels.js'
 import { REGULATOR_DISPLAY } from '#domain/organisations/model.js'
@@ -214,10 +217,14 @@ async function buildSubmissionRows(
     return merged.flatMap((mergedPeriod) => {
       // One row per submitted report, so a resubmitted period fans out. A period
       // with nothing submitted still gets a single blank row (`null`), keeping
-      // outstanding periods visible.
-      const reports = mergedPeriod.submittedReports.length
-        ? mergedPeriod.submittedReports
-        : [null]
+      // outstanding periods visible. The feed is the only consumer of this
+      // projection, so it derives it here rather than the merge stamping it onto
+      // every period for the calendar path to strip straight back off.
+      const submitted = selectSubmittedReports({
+        current: mergedPeriod.report,
+        previousSubmissions: mergedPeriod.previousSubmissions
+      })
+      const reports = submitted.length ? submitted : [null]
       return reports.map((report) =>
         buildRow(
           org,
--- a/src/reports/domain/merge-reporting-periods.js
+++ b/src/reports/domain/merge-reporting-periods.js
@@ -13,8 +13,11 @@ const hasBeenSubmitted = (report) => Boolean(report?.submittedAt)
 /**
  * Selects every report holding a submission for the period: those that have
  * been submitted at least once, identified by a retained `submittedAt`, ordered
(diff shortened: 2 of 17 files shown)


src/reports/application/report-submissions.js
```diff
@@ -169,9 +168,13 @@ function buildRow(
       mergedPeriod.year
     ),
     dueDate: mergedPeriod.dueDate,
-    submittedDate,
-    submittedBy,
-    ...buildTonnageFields(mergedPeriod.report)
+    // Every report-derived field describes this one submitted report, so the
+    // row stays internally consistent. An in-flight draft is never a `report`
+    // here, so it produces no row until it is itself submitted.
+    submittedDate: report?.submittedAt?.slice(0, 10) ?? '',
```
dansweeting, line 177: Not new logic — this slice is pre-existing, just relocated. `submittedAt` is persisted as a full ISO timestamp (`2026-04-17T10:00:00.000Z`); the CSV "Submitted date" column wants a plain date, so `.slice(0, 10)` takes the `YYYY-MM-DD` prefix.

On `main` the identical slice already lives in `buildSubmissionRows` (computed there and passed as the `submittedDate` arg to `buildRow`). The fan-out refactor changed `buildRow` to take a single per-row `report` and derive every report-derived field from it, so the slice moved in here — which is why it shows as an added line even though the truncation is unchanged.

String-slicing rather than going via `Date` is deliberate: it extracts the calendar date without constructing a `Date`, avoiding any timezone shift (a `Date` conversion could roll an early-hours UTC instant back a day in a negative-offset locale).

src/reports/domain/merge-reporting-periods.js
```diff
@@ -66,6 +109,7 @@ export function mergeReportingPeriods(
   for (const cp of computedPeriods) {
     const key = `${cp.year}:${cp.period}`
     const slot = persistedSlots.get(key)
+    const submittedReports = slot ? selectSubmittedReports(slot) : []
```
loveboat: **Suggestion: derive `submittedReports` in the feed instead of on `MergedPeriod`?**

`submittedReports` is stamped onto every merged period, but only the submissions feed consumes it — the calendar path immediately destructures it back off in *both* `buildCalendarPeriods` and `buildAllSubmissionPeriods`, plus the `CalendarPeriod` `Omit` and the `get.test.js` `not.toHaveProperty` guard, all to undo a value that path never wanted.

Since `selectSubmittedReports` only needs `{ current, previousSubmissions }` — both already on the period — the feed can derive it at the one call site:

```js
// report-submissions.js
const submitted = selectSubmittedReports({
  current: mergedPeriod.report,
  previousSubmissions: mergedPeriod.previousSubmissions ?? []
})
const reports = submitted.length ? submitted : [null]
```

That would drop the field from `MergedPeriod`, both merge loops, both builder strips, and the `CalendarPeriod` `Omit` — the calendar path pays and strips nothing, and the projection lives with its only consumer. `selectSubmittedReports` just needs exporting from the domain module.

The one thing we'd lose is the in-your-face JSDoc on `MergedPeriod` steering regulator output away from `report`. If we want to keep that signpost, it could move onto the exported `selectSubmittedReports` / a shorter note on `report`. Happy either way — mostly flagging that we're currently computing-then-discarding across two sites. WDYT?

dansweeting: Done in 05f0cd93. Exported `selectSubmittedReports` and derive it at the feed's one call site: dropped the field from `MergedPeriod`, both merge loops, both builder strips, the `CalendarPeriod` `Omit` and the `get.test.js` guard. The signpost steering regulator output away from `report` now lives on `selectSubmittedReports`, with a one-line pointer left on `report`.

One deviation from your snippet: the `previousSubmissions` default moved inside `selectSubmittedReports` rather than the call site. Merge always populates `previousSubmissions`, so the call-site `?? []` was unreachable and failed the 100% branch gate; defaulting inside the function keeps it coverable via a direct unit test.

src/reports/domain/merge-reporting-periods.js
```diff
@@ -2,6 +2,44 @@
  * @import { PeriodicReport, ReportSummary } from '../repository/port.js'
  */
 
+/**
+ * A report has been submitted when it carries a submission instant. `submittedAt`
+ * is an ISO timestamp or null, so truthiness distinguishes the two.
+ * @param {ReportSummary | null | undefined} report
+ * @returns {report is ReportSummary}
+ */
+const hasBeenSubmitted = (report) => Boolean(report?.submittedAt)
```
senthilinfy, line 11: Better to check status, as that actually indicates status of report. plus dont know what unsubmit from adimn UI does

src/reports/routes/submissions.js
```diff
@@ -51,6 +51,9 @@ export const getReportSubmissions = {
               dueDate: Joi.string().required(),
               submittedDate: Joi.string().allow('').required(),
               submittedBy: Joi.string().allow('').required(),
+              submissionNumber: Joi.alternatives()
+                .try(Joi.number().integer(), Joi.valid(''))
```
senthilinfy: (Optional) why not use` Joi.number().integer().allow('')` ? Just wanted to understand why use try instead

src/reports/domain/merge-reporting-periods.js
dansweeting: `submittedAt` is deliberate here rather than `status`. When a report is unsubmitted from the admin UI its status reverts to `ready_to_submit` but the `submittedAt` instant is retained, so keying on `submittedAt` keeps the feed showing that report's last submitted figures rather than blanking them. That is exactly the unsubmit case: it is covered by the "keeps the last submitted figures after a submitted period is unsubmitted" test in `report-submissions.test.js`.

Checking `status === submitted` instead would drop those rows the moment a period is unsubmitted, which is the regression we are avoiding. `build-calendar-periods.js` does key on `status` (its `latestSubmitted`), but for a different job: current-state resubmission flagging. The two selectors are intentionally distinct, as the `selectSubmittedReports` JSDoc notes.

Longer term, once CPA is live there should be no need to unsubmit anything again. At that point we could measure how many unsubmitted items actually exist and, if negligible, revisit this and treat status-based checking as a correctness cleanup.

src/reports/routes/submissions.js
dansweeting: Done in 7d3c30df.

senthilinfy (approved):

loveboat (approved):

PR #1410, opened 2026-07-01
PAE-1669: Retire the WasteBalancesRepository facade

GraemeF:
Ticket: [PAE-1669](https://eaflood.atlassian.net/browse/PAE-1669)
The waste-balance application service is now the sole entry point over the ledger. The `WasteBalancesRepository` facade and its per-command contract suite are removed, and the plugin registers the service (built from the stream repository) directly alongside `streamRepository`.

`currentWasteBalance` takes a full `WasteBalanceLedgerId` (`{ organisationId, registrationId, accreditationId }`) and returns the caller's `organisationId` rather than recovering it from the latest event, so the ledger id threads through unchanged from request to result.

Test helpers, mocks and the affected call sites build the service from the stream repository and pass the full ledger id.

[PAE-1669]: https://eaflood.atlassian.net/browse/PAE-1669?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/waste-balances/repository/mongodb.plugin.js
```diff
@@ -1,22 +1,19 @@
...
     const streamRepository = streamFactory()
 
-    const factory = createWasteBalancesRepository({
-      streamRepository
-    })
-    const repository = factory()
-
-    registerRepository(server, 'wasteBalancesRepository', () => repository)
     registerRepository(server, 'streamRepository', () => streamRepository)
+    registerRepository(server, 'wasteBalanceService', () =>
```
crosslandwa, line 15: There's a bit of a smell here - `registerRepository` being used to add a "service" to `request.wasteBalanceService` and `server.app.wasteBalanceService` - suggests `registerRepository` isn't the right name (anymore).

Also interesting that this mongo version of the plugin registers `streamRepository` whereas the inmemory version does not

Finally, noticed that the existence of `wasteBalanceService` on `request` isn't advertised in the `HapiRequest` type definition in `hapi-types.js`, and probably should be

src/test/mock-repositories.js
```diff
@@ -133,21 +133,23 @@ export const createMockPackagingRecyclingNotesRepository = (
...
  * vi.fn(). Pass `overrides` to stub specific methods for a test.
  *
- * @param {Partial<WasteBalancesRepository>} [overrides]
- * @returns {WasteBalancesRepository}
+ * @param {Partial<WasteBalanceService>} [overrides]
+ * @returns {WasteBalanceService}
  */
-export const createMockWasteBalancesRepository = (overrides = {}) => ({
-  findBalance: vi.fn(),
+export const createMockWasteBalanceService = (overrides = {}) => ({
```
crosslandwa, line 142: Mocking a service in `mock-repositories` is a bit of a smell.

Does this mean we're mocking/stubbing at an internal application boundary, rather than at port?

GraemeF: Yeah we don't need to do that any more - we can run the real service against the in-memory repository instead. The logic is no longer behind the port.

src/routes/v1/organisations/registrations/summary-logs/integration-test-helpers.js
```diff
@@ -29,19 +29,17 @@ import assert from 'node:assert/strict'
...
  * @param {string} accreditationId
  * @param {string} registrationId
  */
-export const getWasteBalance = async (
-  wasteBalancesRepository,
-  accreditationId,
-  registrationId
-) => {
-  const balance = await wasteBalancesRepository.findBalance({
+export const getWasteBalance = async (env, accreditationId, registrationId) => {
```
crosslandwa, line 40: Looks like all the `getWasteBalance` calls already have `accreditationId` and `registrationId` on the `env` object - perhaps the signature for this can be reduced to just one arg (`env`)?

src/waste-balances/application/mark-excluded-records.js
```diff
@@ -0,0 +1,40 @@
...
+ * @returns {boolean} Whether the record passes validation
+ */
+const isRecordValid = (record) => {
+  const schema = findSchemaForProcessingType(
+    record.data?.processingType,
+    record.type
+  )
+
+  if (!schema) {
+    return true
```
crosslandwa, line 20: Worth logging here if there's no matching schema?

crosslandwa (approved):

src/waste-balances/application/mark-excluded-records.js
GraemeF: That's expected for any tables we have not defined schemas for (we still store their contents)

PR #1413, opened 2026-07-01
PAE-1686: Guard summary log submit against period closure

dansweeting:
Ticket: [PAE-1686](https://eaflood.atlassian.net/browse/PAE-1686)
## What

Adds a submit-time guard that rejects a summary log submission when a periodic report for the same registration has been submitted since the summary log was created.

At validate time, the summary log's open/closed period classification and the check-page preview the operator confirms are computed against the periodic reports as they stood then. Those are then acted on later at submit. A period closes when an operator submits its periodic report, and that route does not touch summary logs, so the existing submit-time staleness guard (keyed on the latest submitted summary log) never fires for it. A period can therefore transition open to submitted in the validate-to-submit window, and the log would submit against a stale snapshot: the confirmed preview no longer matches what gets written, and the restated period's submitted report is silently left out of date.

## Changes

- **Reports repository:** new `hasReportSubmittedSince(organisationId, registrationId, since)` query on the port, with implementations for the in-memory and MongoDB adapters and contract-test coverage on both. It returns true when any report for the org/registration has a `SUBMITTED` status-history entry stamped strictly after the given timestamp.
- **Submit worker:** before syncing waste records, `submitSummaryLog` calls the new query using the summary log's immutable `createdAt` as the reference point. If a report has been submitted since, it throws a `PermanentError`. The existing failure path (`markAsSubmissionFailed`) then moves the log to `SUBMISSION_FAILED` and releases the lock, so the operator sees the generic error page and re-uploads against the current state. `submitSummaryLog` is split into focused helpers to keep each function within the length limit.
- **Tests:** a new application-level integration test drives the real submit worker against real in-memory repositories, covering the rejection path (no waste records written, log left untouched) and the pass-through paths (no submitted report, and a report last submitted before the log was created). The consumer test is updated to supply the reports repository the worker now depends on.

## Why this approach

The guard is deliberately blunt and fail-safe: it fires on any report submission for the registration within the window. Each false positive costs an unnecessary re-upload; none is a bad write. No schema change, no new persisted field, and no frontend work are required. The predicate can be tightened to touched-periods-only later without changing the worker wiring.

Jira: PAE-1686

[PAE-1686]: https://eaflood.atlassian.net/browse/PAE-1686?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/application/summary-logs/submit.js
+++ b/src/application/summary-logs/submit.js
@@ -28,18 +28,17 @@ import { summaryLogMetrics } from '#common/helpers/metrics/summary-logs.js'
  */
 
 /**
- * Rejects submission when a report closed a period during the validate-to-submit
- * window. The operator confirmed a preview built from the periodic reports as
- * they stood at validate; a report submitted since (keyed on the log's immutable
- * createdAt) means that confirmed preview no longer matches what would be
- * written, so the log is failed and the operator re-uploads. See PAE-1686.
+ * Fails submission if any report for this registration was submitted since the
+ * log's createdAt, closing the validate-to-submit race where a period closes
+ * after the operator confirmed the preview. Deliberately blunt: it fires on any
+ * submission, not only periods this log touches. See PAE-1686.
  *
  * @param {object} summaryLog
  * @param {string} summaryLogId
  * @param {import('#reports/repository/port.js').ReportsRepository} reportsRepository
  * @returns {Promise<void>}
  */
-const assertNoPeriodClosedSinceCreation = async (
+const assertNoReportSubmittedSinceCreation = async (
   summaryLog,
   summaryLogId,
   reportsRepository
@@ -169,7 +168,7 @@ export const submitSummaryLog = async (summaryLogId, deps) => {
     )
   }
 
-  await assertNoPeriodClosedSinceCreation(
+  await assertNoReportSubmittedSinceCreation(
     summaryLog,
     summaryLogId,
     reportsRepository
--- a/src/application/summary-logs/submit.js
+++ b/src/application/summary-logs/submit.js
@@ -20,6 +20,7 @@ import { summaryLogMetrics } from '#common/helpers/metrics/summary-logs.js'
  * @property {import('#waste-records/repository/port.js').RowStateRepository} wasteRecordStatesRepository
  * @property {ReturnType<typeof import('#waste-balances/application/waste-balance-service.js').createWasteBalanceService>} wasteBalanceService
  * @property {import('#feature-flags/feature-flags.port.js').FeatureFlags} featureFlags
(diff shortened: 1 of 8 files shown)


src/reports/repository/mongodb.js
```diff
@@ -428,6 +428,38 @@ const performMarkActiveReportsStale = async (
...
+) => {
+  // Both `at` and `since` are canonical `new Date().toISOString()` values, so the
+  // `$gt` string compare tracks chronological order (the ISO invariant this
+  // codebase relies on throughout).
+  const match = await reportsCollection(db).findOne(
+    {
+      organisationId,
+      registrationId,
+      'status.history': {
+        $elemMatch: { status: REPORT_STATUS.SUBMITTED, at: { $gt: since } }
```
senthilinfy: You can use `status.submitted `for this filtering 

Image

src/application/summary-logs/submit.js
```diff
@@ -20,20 +20,54 @@ import { summaryLogMetrics } from '#common/helpers/metrics/summary-logs.js'
...
+ * @param {import('#reports/repository/port.js').ReportsRepository} reportsRepository
+ * @returns {Promise<void>}
+ */
+const assertNoReportSubmittedSinceCreation = async (
+  summaryLog,
+  summaryLogId,
+  reportsRepository
+) => {
+  const reportSubmittedSinceCreation =
+    await reportsRepository.hasReportSubmittedSince(
```
senthilinfy, line 47: could we make this visible via a application layer in reports module? Instead of SL directly using reports repository

dansweeting: Worth flagging that summary-logs already reads from the reports repository directly for its validate-time classification: `reportsRepository.findPeriodicReports` in `classify-and-persist.js` and `validate.js`. So this submit-time guard is following the pattern that is already in place rather than introducing new coupling.

If we wrapped just this one call behind a reports application-layer facade, the existing `findPeriodicReports` calls would still go direct, which would leave things inconsistent. Doing it properly means moving those already-merged calls too, so it is a broader refactor of existing code rather than something this PR should carry.

I have raised a follow-up task to introduce the facade across all the direct call sites, so we can keep this change minimal and land the boundary work on its own. Happy to pull it forward if you feel strongly.

src/reports/repository/mongodb.js
dansweeting: Done in b16b0c481dabaa8ef199b3d9ae462a84bf4a23e9

senthilinfy (approved):

PR #1414, opened 2026-07-01
PAE-1649: repository capability for report resubmission (1/3)

loveboat:

part 1 of 3 — splits #1403 for review size only. the resubmission feature is flag-gated end-to-end and does nothing until all three backend prs + fe #908 land. this is **not** incremental rollout.

adds the write-storage repository capability:
- `markSubmittedReportsRequiringResubmission` op + `resubmissionRequired` field
- exercised at the contract layer, wired into nothing live yet

`findAllPeriodicReports.contract.js` lives here (not with the read path) because it asserts the `resubmissionRequired` projection this pr introduces.

**stacked prs** (merge order: 1 -> {2 and 3, independent} -> fe #908):
- 1/3 repository capability — this pr
- 2/3 detection wiring — #1415
- 3/3 reports calendar — #1416

## how to merge this stack

#1415 and #1416 target this pr's branch (not `main`) so each shows only its own incremental diff. they are kept **draft** so they can't be merged as-is. do **not** merge them into this branch.

1. **merge this pr (#1414) into `main`** — squash + delete branch as normal. github then auto-retargets #1415 and #1416 base from this branch to `main`.
2. because we squash-merge, main's squash commit has a different sha than the pr1 commit in #1415/#1416 history, so they'll show pr1's files as noise until rebased. for each, once this pr is on main:
   ```
   git fetch origin
   git rebase --onto origin/main PAE-1649-pr1-repository-capability <branch>
   git push --force-with-lease
   ```
   this drops the duplicated pr1 commit, leaving a clean diff vs `main`.
3. mark #1415 and #1416 ready and merge each into `main` (independent of each other).


src/reports/repository/mongodb.js
```diff
@@ -384,48 +451,93 @@ const performMarkActiveReportsStale = async (
...
-    stale: /** @type {import('./port.js').ReportStale} */ (d.stale)
-  }))
+  const flaggedIds = latestSubmissionPerPeriod(submitted).map((doc) => doc.id)
+
+  if (flaggedIds.length === 0) {
+    return []
+  }
+
+  const scope = { id: { $in: flaggedIds } }
+  const flagged = await flagReportsBySummaryLog(db, {
```
dansweeting, line 532: Similar code where we do this for `stale` passes the helper a predicate as `selectFilter`, so the single `updateMany` selects and mutates in one self-selecting write. This path instead reads candidates, computes latest-per-period ids in JS, then flags a frozen `{ id: { $in: flaggedIds } }` list, so selection and mutation are two separate round trips. It works, but it diverges from the stale pattern and reintroduces a read-then-write step that the stale path deliberately avoids.

Worth aligning the two for symmetry if possible? Maybe it's a more fiddly prospect here.

src/reports/repository/helpers.js
```diff
@@ -2,8 +2,31 @@ import {
...
+ * from a flat list of report-like documents. Shared by the in-memory and
+ * mongodb adapters so the "latest per period" rule lives in one place.
+ *
+ * @template {{ year: number, cadence: string, period: number, submissionNumber: number }} T
+ * @param {T[]} reports
+ * @returns {T[]}
+ */
+export const latestSubmissionPerPeriod = (reports) => [
+  ...[...reports]
+    .sort((a, b) => b.submissionNumber - a.submissionNumber)
```
dansweeting, line 24: "Latest" is defined by `submissionNumber` alone, ties broken by input order (Map iteration in-memory, `find` order in mongo). That's only safe if `submissionNumber` is unique per period among submitted reports. This helper is now the single source of that rule, so state the invariant in the docstring or assert it, rather than leaving the tie-break load-bearing by accident.

src/reports/repository/contract/markSubmittedReportsRequiringResubmission.contract.js
```diff
@@ -0,0 +1,166 @@
...
+      await createAndSubmitReport(repository)
+
+      const first = await callMark()
+      expect(first).toHaveLength(1)
+
+      const second = await callMark()
+      expect(second).toEqual([])
+    })
+
+    it('skips reports whose source.summaryLogId matches the given summaryLogId', async () => {
```
dansweeting: This doesn't pin the behaviour it names: `createAndSubmitReport` already defaults `source.summaryLogId` to `'sl-1'`, and the override repeats that value, so the test passes even if the override is ignored. The uncovered case is a period with two submissions where only the latest is source-matched: both adapters should skip the whole period. Add that and assert `flagged === []`.

src/reports/repository/port.js
```diff
@@ -108,6 +117,31 @@
...
+ * {@link ReportsRepository.markSubmittedReportsRequiringResubmission}.
+ * Contains the fields needed to audit the resubmission-required transition.
+ * @typedef {{
+ *   reportId: string,
+ *   year: number,
+ *   cadence: string,
+ *   period: number,
+ *   submissionNumber: number,
+ *   resubmissionRequired: ReportResubmissionRequired
+ * }} MarkReportRequiringResubmissionResult
```
dansweeting: Purely cosmetic, take it or leave it: the typedef drops "Submitted" and singularises "Report" while the op is `markSubmittedReportsRequiringResubmission` and returns a plural array. It mirrors `MarkReportStaleResult` so there's precedent, but happy to see it renamed if you want the port to read consistently.

src/reports/repository/mongodb.js
loveboat: intrinsic here rather than an oversight: stale flags *all* active reports, so the predicate is fully set-based and one self-selecting updateMany does it. resubmission flags the *latest submitted per period*, which needs max(submissionNumber) per period resolved before we can name the target ids -- mongo cant express "latest per group" as a plain update filter. options were a read-then-write (this) or an aggregation pipeline to derive the ids server-side; the latter is more machinery, not less, for the same two logical steps. left as-is, added an invariant docstring on the shared helper so the "latest" rule is explicit.

src/reports/repository/helpers.js
loveboat: done -- added an invariant note to the `latestSubmissionPerPeriod` docstring: submissionNumber is unique per period among submitted reports (each resubmission increments it), so the highest value names a single report and the input-order tie-break is never load-bearing.

src/reports/repository/contract/markSubmittedReportsRequiringResubmission.contract.js
loveboat: good catch -- rewrote it into the case that actually bites: two submissions in one period where only the *latest* (submissionNumber 2) is source-matched to the marking log and the earlier one is from a different log. asserts `flagged === []`, so it pins that we skip the whole period on the latest-submission match rather than falling back to the earlier one.

src/reports/repository/port.js
loveboat: renamed to `MarkSubmittedReportRequiringResubmissionResult` for port symmetry with the op name -- kept singular Report since each entry is per-report.

dansweeting (approved):

PR #1415, opened 2026-07-01
PAE-1649: detection wiring for report resubmission (2/3)

loveboat:

part 2 of 3 — splits #1403 for review size only. flag-gated end-to-end; nothing renders until all three backend prs + fe #908 land. **not** incremental rollout.

flags submitted reports as requiring resubmission at submit time, gated behind the `closedPeriodAdjustments` bare config helper (`isClosedPeriodAdjustmentsEnabled()`, default false):
- validate-time: computes and persists `closedPeriods` unconditionally — inert derived data while the flag is off, so it survives to submit time if the flag flips on between validate and submit
- submit-time: early-return before `markSubmittedReportsRequiringResubmission` when off

the flag is a bare config helper, not the FeatureFlags port — `src/feature-flags/*` is untouched.

merge order: this and 3 are independent -> fe #908


src/application/summary-logs/classify-and-persist.js
```diff
@@ -115,6 +116,20 @@ export const classifyLoads = ({
...
+ * When the feature is off, returns the loads with closedPeriods emptied so
+ * nothing is persisted (or surfaced on the summary-log response) while it is
+ * disabled; the open/closed load breakdown is untouched.
+ *
+ * @param {LoadsByReportingPeriod | null} loadsByReportingPeriod
+ * @returns {LoadsByReportingPeriod | null}
+ */
+export const gateClosedPeriods = (loadsByReportingPeriod) =>
+  loadsByReportingPeriod && !isClosedPeriodAdjustmentsEnabled()
+    ? { ...loadsByReportingPeriod, closedPeriods: [] }
```
senthilinfy: Just asking to think about this scenario. What happens in below scenario

- CPA is disabled. User uploads SL and validation is complete
- CPA is enabled. User finishes submission of validated SL upload from above

src/application/summary-logs/classify-and-persist.js
```diff
@@ -115,6 +116,20 @@ export const classifyLoads = ({
...
+/**
+ * Gates the closed-period refs behind the closed-period-adjustments feature.
+ * When the feature is off, returns the loads with closedPeriods emptied so
+ * nothing is persisted (or surfaced on the summary-log response) while it is
+ * disabled; the open/closed load breakdown is untouched.
+ *
+ * @param {LoadsByReportingPeriod | null} loadsByReportingPeriod
+ * @returns {LoadsByReportingPeriod | null}
+ */
+export const gateClosedPeriods = (loadsByReportingPeriod) =>
```
dansweeting: Maybe it's ok to drop the gating of `closedPeriods` here? `classifyByPeriodStatus` computes it as pure derived data, so populating and persisting it is harmless.

The only thing the emptying protects against is `closedPeriods` surfacing on the summary-log response while the feature is dark (the persisted document is what the GET returns). I don't think that's a problem: it's derived data the frontend doesn't consume while the feature is off, and leaving it present does no harm.

So I'd suggest we remove `gateClosedPeriods` and rely solely on the submit-time flag check in `createOnSummaryLogUploaded`. That keeps population and persistence unconditional and gives us a single gate at the point where the behaviour actually happens.

src/application/summary-logs/period-status.js
```diff
@@ -103,43 +111,97 @@ const emptyResult = () => ({
...
-
-    hasAnyDate = true
-    if (isDateInSubmittedPeriod(submittedPeriods, dateValue, cadence)) {
-      return PERIOD_STATUS.CLOSED
-    }
+  const dates = reportingDateFields.map((field) => data[field])
+
+  if (
+    dates.some(
+      (date) => date && isDateInSubmittedPeriod(submittedPeriods, date, cadence)
```
senthilinfy: Superficially looks like classifyPeriodStatus and closedPeriodRefsFor has some common bits(identifying list of closed periods) that can be extracted out

```
const dates = reportingDateFields.map((field) => data[field])

  if (
    dates.some(
      (date) => date && isDateInSubmittedPeriod(submittedPeriods, date, cadence)
```

src/application/summary-logs/record-change.js
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Classifies how a record changed in this upload from its latest version.
+ *
+ * @param {ValidatedWasteRecord['record']} record
+ * @param {string} summaryLogId
+ * @returns {RecordChange}
+ */
+export const determineRecordStatus = (record, summaryLogId) => {
+  const lastVersion = record.versions.at(-1)
```
dansweeting, line 31: Worth a conscious call on this: the old `load-counts.js` copy indexed `versions[length - 1]`, so an empty `versions` array threw a `TypeError` (the old test asserted `.toThrow()` and called it "an invalid state that shouldn't occur in practice"). With `.at(-1)` plus the optional chaining, an empty array now returns `UNCHANGED` and the record is silently swept into `unchanged.included` instead.

Unifying the two copies onto one helper is the right move, and `period-status.js` already used `.at(-1)`, so the inconsistency was real. My only question is which behaviour we want to keep. An empty `versions` array means a record with no version history at all, which I'd expect to be a "can't happen" data-integrity violation rather than a benign state.

If it genuinely can't happen, silently bucketing it as unchanged hides the upstream bug that produced it, so I'd lean towards keeping it loud (throw, or at least log / emit a metric). If it can legitimately occur, could we note why in the test rather than "handles it gracefully", which reads as papering over the old crash?

src/application/summary-logs/load-counts.test.js
```diff
@@ -471,14 +471,12 @@ describe('countByWasteBalanceInclusion', () => {
...
-          wasteRecords,
-          summaryLogId: CURRENT_SUMMARY_LOG_ID
-        })
-      ).toThrow()
+      const result = countByWasteBalanceInclusion({
+        wasteRecords,
+        summaryLogId: CURRENT_SUMMARY_LOG_ID
+      })
+
+      expect(result.unchanged.included.rowIds).toEqual(['row-1'])
```
dansweeting: This test enshrines the behaviour change discussed in https://github.com/DEFRA/epr-backend/pull/1415#discussion_r3512498804: an empty `versions` array used to throw here and now classifies as `unchanged.included`. Same question applies, whether an empty array is a "can't happen" state we should keep failing loudly on.

src/config.js
```diff
@@ -521,5 +527,13 @@ function getConfig(overrides = {}) {
 
 const isProductionEnvironment = () => config.get('cdpEnvironment') === 'prod'
 const isLocalEnvironment = () => config.get('cdpEnvironment') === 'local'
+const isClosedPeriodAdjustmentsEnabled = () =>
```
senthilinfy, line 530: **nitpick**: Usually we seem to be placing these functions on feature-flags.config.js

src/application/summary-logs/period-status.js
loveboat: Yeah I thought the same and did toy with that. I did change the above functions imperative mode to be more fp-like so now they're even more alike - mostly I did this because seeing a `let` anywhere brings me out in a rash 😆

src/application/summary-logs/classify-and-persist.js
loveboat: Yes in that context gating does seem a bit daft. Better for the data to be there and unused than not there at all.

src/config.js
loveboat: Yeah `feature-flags.config.js` is the FeatureFlags port impl (injected + mocked via the inmemory adapter). I did have this going via the port/adapter originally but it involved touching about 10+ files for a single call to the flag. Seemed more optional to just read it directly in the code. I can move it for consistency but the trade-off didn't seem worth it at the time - esp if we'll be deleting it soon anyway (all being well! 🤞 )

src/application/summary-logs/period-status.js
loveboat: Ok, made a little tweak to this to introduce a shared predicate: https://github.com/DEFRA/epr-backend/pull/1415/changes/78159fdb87fbb0871631620266c213bc9a677d26

src/application/summary-logs/record-change.js
loveboat: Restored the throw: https://github.com/DEFRA/epr-backend/pull/1415/changes/66d7c4921a6615c11308cc3b7171d521f053fa0e

src/application/summary-logs/classify-and-persist.js
loveboat: Dropped the gate: https://github.com/DEFRA/epr-backend/pull/1415/changes/fb150b6b71f84960e8eaa98049294de43f24aae4

src/application/summary-logs/classify-and-persist.js
loveboat: Gating removed: https://github.com/DEFRA/epr-backend/pull/1415/changes/fb150b6b71f84960e8eaa98049294de43f24aae4

senthilinfy (approved):

PR #1416, opened 2026-07-01
PAE-1649: reports calendar for report resubmission (3/3)

loveboat:

part 3 of 3 — splits #1403 for review size only. flag-gated end-to-end; nothing writes the data this renders until all three backend prs + fe #908 land. **not** incremental rollout.

**stacked on #1414 (1/3)** — review/merge that first. independent of #1415 (2/3).

surfaces `requires_resubmission` + the skeleton calendar item on the reports calendar (the contract the fe consumes). the calendar becomes submission-grained so the original submitted report stays visible alongside the resubmission skeleton.

merge order: 1 (#1414) -> {2 (#1415) and 3 (this), independent} -> fe #908


senthilinfy (approved):

src/reports/domain/build-calendar-periods.test.js
```diff
@@ -0,0 +1,74 @@
...
+      submissionNumber: 1
+    })
+    expect(result[1]).toMatchObject({
+      submissionNumber: 2,
+      periodStatus: 'requires_resubmission',
+      report: null,
+      dueDate: '2026-02-20'
+    })
+  })
+})
```
dansweeting, line 93: Every case here passes a single-element array, so nothing exercises the `flatMap` over a mixed list. How about adding one case with three periods (one flagged, one submitted-unflagged, one with no report) asserting the four-element result in order, pinning expansion as local to the flagged period with the skeleton immediately after its origin? The single-element tests already hit full coverage, so this is purely a regression guard for a future edit to the expansion logic, but it is the one property the current suite cannot see. Note `period()` is hardcoded to period 1, so distinct periods need building inline.

loveboat: addressed in: https://github.com/DEFRA/epr-backend/pull/1416/changes/49581e2a90d1a142254004f503e5eba6c9aad174

PR #1425, opened 2026-07-02
PAE-1696: Add explicit reason codes for non-included waste balance rows

senthilinfy:
Ticket: [PAE-1696](https://eaflood.atlassian.net/browse/PAE-1696)

The waste-record CSV export's "Included in Waste Balance" column now distinguishes rows that are genuinely excluded (`false`, with a reason where one exists) from rows where inclusion simply can't be computed — no accreditation, a registered-only processing type, or no waste-balance classifier for the section — which render as `NA` instead of a fabricated reason code.

`getWasteBalanceClassification` returns `null` for the non-computable cases (checked before, and taking precedence over, the existing `excludedFromWasteBalance` flag), fully localised to the export module with no changes to `classifyForWasteBalance`, `period-status.js`, or the frontend.

Full discussion: https://defra-digital-team.slack.com/archives/C09N62D709G/p1783025989070669

[PAE-1696]: https://eaflood.atlassian.net/browse/PAE-1696?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/waste-records-export/domain/is-included-in-waste-balance.js
```diff
@@ -20,28 +22,48 @@ import { ROW_OUTCOME } from '#domain/summary-logs/table-schemas/validation-pipel
...
 
+  if (!accreditation) {
+    return {
+      included: false,
+      reasons: [{ code: CLASSIFICATION_REASON.NOT_ACCREDITED }],
+      tonnage: null
+    }
+  }
+
+  if (REGISTERED_ONLY_PROCESSING_TYPES.has(record.data?.processingType)) {
```
dansweeting: The PR description says `SUBMITTED_ON_REGISTERED_ONLY_TEMPLATE` means the record was submitted "before the operator became accredited", but this branch fires for every registered-only row regardless of date: it never looks at the record date or the accreditation window. A registered-only row dated after accreditation gets the same code. Could we either rename it to something date-agnostic (e.g. `PROCESSING_TYPE_REGISTERED_ONLY`) or drop the temporal claim from the description and docstring?

src/waste-records-export/domain/is-included-in-waste-balance.js
```diff
@@ -20,28 +22,48 @@ import { ROW_OUTCOME } from '#domain/summary-logs/table-schemas/validation-pipel
...
 export const getWasteBalanceClassification = (
   record,
   accreditation,
   overseasSites
 ) => {
-  if (accreditation === null) {
-    return null
-  }
-
   if (record.excludedFromWasteBalance) {
```
dansweeting: This `excludedFromWasteBalance` check runs before the `!accreditation` check below, and `excludedFromWasteBalance` is validity-based (independent of accreditation). So an unaccredited *invalid* row gets a blank reason rather than `NOT_ACCREDITED`. Which reason should win that slot for an unaccredited operator? Either way, could we add a test to pin the ordering, since it is currently incidental and unasserted?

src/waste-records-export/domain/is-included-in-waste-balance.test.js
```diff
@@ -68,14 +68,50 @@ const prnIssuedExporterRecord = buildWasteRecord({
...
       getWasteBalanceClassification(record, null, ORS_VALIDATION_DISABLED)
-    ).toBeNull()
+    ).toEqual({
+      included: false,
+      reasons: [{ code: 'NOT_ACCREDITED' }],
+      tonnage: null
+    })
+  })
+
+  it('returns SUBMITTED_ON_REGISTERED_ONLY_TEMPLATE for a registered-only-processing-type row once the operator is accredited', () => {
```
dansweeting: This test uses a date inside the accreditation window and asserts `SUBMITTED_ON_REGISTERED_ONLY_TEMPLATE`, but since the code ignores the date it would pass with any date. Given the "before accreditation" framing, could we add a case with a registered-only record dated *outside* the accreditation period asserting the same code? That makes the real (date-agnostic) contract explicit.

dansweeting (approved):

PR #1429, opened 2026-07-03
PAE-1650: Support resubmission drafts for restated closed periods

dansweeting:
Ticket: [PAE-1650](https://eaflood.atlassian.net/browse/PAE-1650)
When a later summary log restates an already-submitted (closed) reporting period, that period's latest submitted report is flagged as requiring resubmission. This change lets an operator act on that flag: create a correcting resubmission draft, and see the period stay in its "requires resubmission" state throughout.

## Create a resubmission draft

The create-report route previously rejected any submission beyond the first outright, so a resubmission could never be started. The route now accepts submission numbers greater than one, gated by a service-layer rule: a resubmission may be created only when the closed-period-adjustments feature flag is on and the previous submission is a submitted report flagged as requiring resubmission.

The guards run in order: an existing report for the same submission number is rejected first (so a duplicate returns `report_already_exists` regardless of the flag), then the feature flag, then the previous-submission block rule.

A disallowed attempt returns 409 with a distinct reason in the body (`resubmission_feature_disabled` when the flag is off, `resubmission_not_permitted` when the previous submission is missing, not submitted, or not flagged) so the frontend can tell it apart from the existing duplicate-period conflict:

```json
{
  "statusCode": 409,
  "error": "Conflict",
  "message": "Resubmission 2 not permitted for monthly period 1 of 2026",
  "reason": "resubmission_feature_disabled"
}
```

## Keep the calendar state once a draft exists

Previously, as soon as a resubmission draft existed the reports calendar collapsed the period to a single item carrying the draft's own status (for example `ready_to_submit`) and dropped the original submitted report from the response.

The calendar now, for a period whose latest submitted report is flagged, always emits two items:

- the original submitted item, so it stays visible in the Submitted table, and
- a `requires_resubmission` slot at the next submission number that carries the in-flight draft when one exists, or `null` before any draft is started.

The period therefore stays flagged as requiring resubmission regardless of the draft's own status, and the frontend derives the call to action from the carried report. This holds across the whole draft lifecycle: pre-draft, in progress, ready to submit, and reverting to the pre-draft state when the draft is deleted.

For a flagged period with a ready-to-submit draft (dates elided):

```json
{
  "cadence": "monthly",
  "reportingPeriods": [
    {
      "year": 2026, "period": 1, "submissionNumber": 1,
      "periodStatus": "submitted",
      "report": {
        "id": "663...a1", "status": "submitted", "submissionNumber": 1,
        "submittedAt": "2026-02-10T09:00:00.000Z", "submittedBy": { "id": "user-1" }
      }
    },
    {
      "year": 2026, "period": 1, "submissionNumber": 2,
      "periodStatus": "requires_resubmission",
      "report": {
        "id": "663...b2", "status": "ready_to_submit", "submissionNumber": 2,
        "submittedAt": null, "submittedBy": null
      }
    }
  ]
}
```

Once the resubmission is itself submitted it becomes the latest submitted report, so the period collapses back to a single submitted item and the requires-resubmission slot disappears. The original flag is never explicitly cleared: the calendar keys on the latest submitted report, so a superseded submission's stale flag is simply ignored (per ADR-0039).

Internally, the period merge now surfaces previous submissions so the calendar builder can locate the flagged submitted report whether it is the current slot or an earlier one. That detail is stripped from the API response.

[PAE-1650]: https://eaflood.atlassian.net/browse/PAE-1650?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/reports/domain/build-calendar-periods.js
```diff
@@ -1,42 +1,80 @@
...
-      return [item]
+    if (!flaggedSubmitted?.resubmissionRequired) {
+      return [{ ...period, periodStatus: derivePeriodStatus(period) }]
     }
 
+    // `current` is the highest-submissionNumber report for the period. When it
+    // is the flagged submitted report itself there is no resubmission draft yet
+    // (the pre-draft skeleton); once the operator starts one, `current` is that
+    // later draft and the flagged report drops to previousSubmissions.
+    const draft = period.report === flaggedSubmitted ? null : period.report
```
loveboat: Should this be checking the `submissionNumber`?

```js
const draft =
  period.report &&
  period.report.submissionNumber > flaggedSubmitted.submissionNumber
    ? period.report
    : null;
```

dansweeting: Done in cd805894. Switched to comparing `submissionNumber` so the draft check no longer depends on `period.report` and the flagged report being the same object instance.

loveboat (approved):

PR #1454, opened 2026-07-08
PAE-1657: calendar-expand-submissions: Expand reports calendar with an admin all-submissions view

dansweeting:
Ticket: [PAE-1657](https://eaflood.atlassian.net/browse/PAE-1657)
## What

Adds an optional `?expand=submissions` query argument to the existing
`GET /v1/organisations/{organisationId}/registrations/{registrationId}/reports/calendar`
endpoint. With the argument, the calendar lists every submission per reporting
period: each superseded submission is surfaced as its own item
(`periodStatus: submitted`), alongside the existing `requires_resubmission`
skeleton. Without the argument, the response is byte-for-byte unchanged.

## Why

Callers that need the full submission history for a registration (for example an
operations overview) can now request it from the calendar itself, rather than
only seeing the collapsed, current-state view. The expanded history is the same
data already available on the report-detail view, so it is exposed to any
authorised caller of the route with no further scope gating. The default,
collapsed calendar continues to serve the operator frontend and public register
exactly as before.

## Changes

- `reports/routes/get.js` — validates `expand` against
  `Joi.string().valid('submissions')` and selects the period builder per request:
  the expanded builder when the argument is set, otherwise today's collapsing
  builder. The two builders share a `PeriodBuilder` type.
- `reports/domain/build-all-submission-periods.js` — new domain builder that
  expands merged periods into one item per submission. It reuses
  `buildCalendarPeriods` for the current-state items, then adds the period's
  previous submissions those items do not already surface. Deduping is keyed on
  report id (not submission number) so the synthetic resubmission skeleton, which
  carries no real report, can never mask a genuine submission. Only submitted
  previous submissions are emitted, enforcing the invariant that a submission
  number is only allocated once the prior submission is submitted.
- Integration and unit tests covering the expansion, the collapsed default when
  the argument is absent, dedup of the flagged report when it is a previous
  submission, exclusion of a non-submitted report, and rejection of an unknown
  `expand` value.

Jira: PAE-1657

[PAE-1657]: https://eaflood.atlassian.net/browse/PAE-1657?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/reports/domain/build-all-submission-periods.js
+++ b/src/reports/domain/build-all-submission-periods.js
@@ -1,0 +1,46 @@
+import { buildCalendarPeriods } from './build-calendar-periods.js'
+import { PERIOD_STATUS } from './period-status.js'
+
+/**
+ * @import { MergedPeriod } from './merge-reporting-periods.js'
+ * @import { CalendarPeriod } from './build-calendar-periods.js'
+ */
+
+/**
+ * Expands merged reporting periods into one calendar item per submission, for
+ * the admin (all-submissions) view. Superseded submissions are deliberately
+ * hidden from the collapsed calendar (ADR-0038); the regulator's overview needs
+ * the full audit trail instead.
+ *
+ * Reuses buildCalendarPeriods for each period's current-state items (the current
+ * report plus, when a resubmission is pending, the requires_resubmission
+ * skeleton) then adds any previous submissions not already surfaced, deduped on
+ * submissionNumber so the flagged submitted report the skeleton derives from is
+ * not emitted twice. Historical submissions carry periodStatus 'submitted' (each
+ * was submitted once); no new status value is introduced. Items within a period
+ * are ordered by submissionNumber ascending.
+ *
+ * @param {MergedPeriod[]} mergedPeriods
+ * @returns {CalendarPeriod[]}
+ */
+export const buildAllSubmissionPeriods = (mergedPeriods) =>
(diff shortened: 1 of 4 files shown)


src/reports/domain/build-all-submission-periods.js
```diff
@@ -0,0 +1,64 @@
...
+    const { previousSubmissions = [], ...period } = mergedPeriod
+    const currentStateItems = buildCalendarPeriods([mergedPeriod])
+    const surfacedReportIds = new Set(
+      currentStateItems.flatMap((item) => (item.report ? [item.report.id] : []))
+    )
+
+    const historicalItems = previousSubmissions
+      .filter(
+        (submission) =>
+          submission.status === REPORT_STATUS.SUBMITTED &&
```
senthilinfy, line 51: previousSubmissions can never be in non submitted status right?

senthilinfy (approved):

src/reports/domain/build-all-submission-periods.js
dansweeting: That's right, you can't create a new submission unless
- the previous submission was submitted,
- and it was flagged with 'requires resubmission' as result of a closed-period adjustment

And this is what populates previousSubmissions

PR #1462, opened 2026-07-09
PAE-1733: Emit exclusion reason for non-waste-balance sections

dansweeting:
Ticket: [PAE-1733](https://eaflood.atlassian.net/browse/PAE-1733)
## What

Template sections that never contribute to the waste balance by design (tables with no `classifyForWasteBalance`) now emit an explicit `exclusionReasons` code — `TEMPLATE_SECTION_DOES_NOT_CONTRIBUTE_TO_WASTE_BALANCE` — on their `nonBalanceAffecting` rows in `loadsByReportingPeriod`, instead of an empty array.

This covers the four by-design sheet/template combinations in the ticket, each with an accredited integration test: REPROCESSOR_INPUT Processed, REPROCESSOR_OUTPUT Received, REPROCESSOR_OUTPUT Sent on, and EXPORTER Sent on. Registered-only tables also carry no classifier, so their rows pick up the same code as an incidental consequence.

The change is additive to the existing `exclusionReasons: string[]` payload, so the response schema shape is unchanged.

## Why

An empty `exclusionReasons` array was overloaded: it was produced by by-design non-contributing rows, by included zero-tonnage rows, and by adjusted rows with no net tonnage change. A by-design load could not be told apart from one that was evaluated and excluded for a data reason, making by-design loads look like they had a data problem (parent bug PAE-1733).

With an explicit code, by-design rows are now unambiguous:
- by-design non-contributing rows carry `TEMPLATE_SECTION_DOES_NOT_CONTRIBUTE_TO_WASTE_BALANCE`
- rows evaluated and excluded keep their existing codes (`PRN_ISSUED`, `ORS_NOT_FOUND`, `MISSING_REQUIRED_FIELD`, etc.)
- included zero-tonnage rows and no-op adjustments keep an empty array

## Also

Removes the dormant, unused `createDateOnlyClassifier` helper. It returned `EXCLUDED` with an empty `reasons` array — the same ambiguity this PR removes — and was referenced only by its own tests. Deleting it stops the pattern reintroducing the ambiguity.

## Follow-up

The frontend must map the new code to check-page copy (tracked separately) before this deploys; until then the raw code string renders on the check page, exactly as `OUTSIDE_ACCREDITATION_PERIOD` does today.

[PAE-1733]: https://eaflood.atlassian.net/browse/PAE-1733?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


loveboat (approved):
Approve. Verified the two behaviour-tightening moves are safe: the `!schema` guard in classifyByPeriodStatus means removing the `schema?.` optional chaining can't NPE, and both WasteBalanceClassificationResult members declare `reasons` as required so dropping the `result?.reasons` guard is type-safe. createDateOnlyClassifier confirmed unused (definition + own test only). Strong integration coverage of all four by-design combos. Merge before FE #935.

loveboat (approved):

PR #1470, opened 2026-07-09
PAE-1339: [1/4] crush test type errors — fixtures, report-detail, parsers

loveboat:
reduce `lint:types:tests` errors 748 -> 485 (-263) across 7 test files, test-only, prod `[src]` unaffected. all touched files type-clean; full suite green.

- **server test fixtures** (`server-with-mock-db`, `server-with-db`): assert fixture shape via `TestAPI<{...}>`; type `db` as the spyable mock.
- **mongodb repo fixtures** (overseas-sites imports, organisations ors): `MongoFixtures` cast idiom; narrow `findById`/`findOne` nulls with `node:assert`.
- **aggregate-report-detail.test.js**: `asWasteRecords` cast helper for minimal test records; narrow optional `result.exportActivity`.
- **ors-spreadsheet-parser.test.js**: type the `buildOrsWorkbook` metadata/dataRow params (fixes `{}`/`never[]` inference).
- **row-continuity.test.js**: type the `ValidatedWasteRecord`/`WasteRecord` builders; optional-chain validation-issue reads; cast open-context `previousSummaryLog`.


src/overseas-sites/imports/repository/mongodb.test.js
```diff
@@ -148,6 +164,8 @@ describe('MongoDB ORS imports repository', () => {
     await repository.updateStatus('import-ttl-2', ORS_IMPORT_STATUS.PROCESSING)
 
     const found = await repository.findById('import-ttl-2')
+    assertPresent(found?.expiresAt)
```
dansweeting, line 167: `assertPresent(found?.expiresAt)` asserts the field, not the record. Every other call site in this PR asserts the fetched document (`assertPresent(found)`); this one narrows through the optional chain, so if `found` were ever null the chain yields `undefined` and we would be testing `expiresAt` rather than the record's existence. It satisfies tsc, but it is inconsistent with the convention elsewhere.

Suggest `assertPresent(found)` (the following `expect(found.expiresAt).toBeInstanceOf(Date)` already proves the date is non-null at runtime), or `assertPresent(found)` then `assertPresent(found.expiresAt)` if you also want to narrow the date for the type checker.

src/application/summary-logs/validations/row-continuity.test.js
```diff
@@ -109,7 +118,9 @@ describe('validateRowContinuity', () => {
 
     it('returns valid result when existingWasteRecords is null', () => {
       const wasteRecords = [createValidatedWasteRecord({ rowId: 'row-1' })]
-      const existingWasteRecords = null
+      const existingWasteRecords = /** @type {WasteRecord[]} */ (
```
dansweeting, line 121: The `null`/`undefined` -> `unknown` -> `WasteRecord[]` double-casts here and at line 136 exist only because `validateRowContinuity`'s JSDoc declares `@param {WasteRecord[]} existingWasteRecords` as non-nullable, yet the implementation's first line is `if (!existingWasteRecords || existingWasteRecords.length === 0)` and these two tests exist specifically to exercise the null/undefined paths. The production type is wrong, not the tests.

Prefer fixing the contract: change the JSDoc to `@param {WasteRecord[] | null | undefined}` (or default the param to `[]`), then delete both casts and pass `null`/`undefined` directly. That is a one-line prod change that makes the type match reality, rather than a test-only cast that hides it. Flagging because the ticket is framed as test-only, but this seems the better outcome.

src/reports/domain/aggregation/aggregate-report-detail.test.js
```diff
@@ -473,9 +491,12 @@ describe('#aggregateReportDetail', () => {
...
       ]
 
-      const result = aggregateReportDetail(records, exporterArgs)
+      const result = aggregateReportDetail(
+        asWasteRecords(records),
+        exporterArgs
+      )
 
-      expect(result.exportActivity.totalTonnageExported).toBe(11.47)
+      expect(result.exportActivity?.totalTonnageExported).toBe(11.47)
```
dansweeting, line 499: `exportActivity` is genuinely optional in the type and absent for reprocessor args, but within the exporter/accredited-exporter blocks it is always present. The PR handles that inconsistently: most assertions become `result.exportActivity?.x` (weaker, stops documenting the invariant) while a single `assertPresent(result.exportActivity)` is added at line 649.

This is not a correctness issue (an unexpected `undefined` still fails `.toBe(0)`/`.toStrictEqual([])`), but worth settling on one style per block before it becomes the copied pattern for parts 2 to 4. In the exporter blocks I would put a single `assertPresent(result.exportActivity)` at the top and keep the direct `.x` accesses, so it both narrows and asserts the invariant.

src/application/summary-logs/validations/row-continuity.test.js
```diff
@@ -230,15 +243,17 @@ describe('validateRowContinuity', () => {
...
-        'previous-summary-log-id'
-      )
+      expect(fatals[0]?.category).toBe(VALIDATION_CATEGORY.BUSINESS)
+      expect(fatals[0]?.code).toBe('SEQUENTIAL_ROW_REMOVED')
+      expect(fatals[0]?.message).toContain('row-1')
+      expect(fatals[0]?.message).toContain('cannot be removed')
+      expect(fatals[0]?.context?.location?.rowId).toBe('row-1')
+      expect(fatals[0]?.context?.location?.sheet).toBe('Received')
+      expect(
+        /** @type {PreviousSummaryLogRef} */ (
```
dansweeting, line 253: The inline `/** @type {PreviousSummaryLogRef} */ (fatals[0]?.context?.previousSummaryLog)` cast is repeated three times (here and around lines 370 and 375). The cast is justified (`context` is an untyped bag), but hoist it once per test: `const prev = /** @type {PreviousSummaryLogRef} */ (fatals[0]?.context?.previousSummaryLog)`, then assert `prev.id` / `prev.submittedAt`. No behaviour change, just less duplication.

src/overseas-sites/imports/repository/mongodb.test.js
loveboat: Fixed on #1483 (stack tip, to avoid a re-merge) — now `assertPresent(found)` then `assertPresent(found.expiresAt)`: the first narrows the record so the field is accessible, the second narrows the nullable date for `.getTime()`.

src/application/summary-logs/validations/row-continuity.test.js
loveboat: Agreed — fixed the contract on #1483 rather than the tests: `existingWasteRecords` is now `WasteRecord[] | null | undefined` (the impl already guards `!existingWasteRecords`), so both double-casts are gone and the tests pass `null`/`undefined` directly. `[prod]` typecheck stays clean.

src/reports/domain/aggregation/aggregate-report-detail.test.js
loveboat: Done on #1483 — `assertPresent(result.exportActivity)` at the top of each exporter block in this describe, with direct `.x` accesses (matching the accredited-exporter block at ~649).

src/application/summary-logs/validations/row-continuity.test.js
loveboat: Done on #1483 — hoisted to a single `const previousSummaryLog` per test.

dansweeting (approved):
Comments actioned on the stack tip (#1483) and verified. Approving.

PR #1471, opened 2026-07-09
PAE-1737: trim monthly report obligations to accreditation validFrom

timrutter:
Ticket: [PAE-1737](https://eaflood.atlassian.net/browse/PAE-1737)
## Summary

Trims an accredited operator's **monthly** report obligations so they start from the accreditation's `validFrom`, instead of always spanning the whole calendar year. Fixes operators being shown — and told to complete — monthly reports for periods before they were accredited (PAE-1737, first narrow slice of PAE-1703).

The `validFrom` front-trim is applied **at the source** in `generateReportingPeriods` (optional `fromDate` param), so every caller is bounded consistently rather than duplicating the filter:

- **Operator reports calendar** (`GET .../reports/calendar`) — bead `epr-reex-tcv.2`
- **Admin all-submissions export** (`buildSubmissionRows`) — bead `epr-reex-tcv.3`
- **Report compliance** (`generateReportCompliance`, which feeds the **public register**) — was missing the bound and still surfaced pre-accreditation months as obligations; now consistent (review feedback).

Stacked on #1467 (the `filterPeriodsFromDate` helper) — this branch includes those commits and targets `main`.

## Behaviour

- When the cadence is monthly (accredited), periods that ended **before** the accreditation `validFrom` are dropped; the period **containing** `validFrom` is kept (inclusive). Quarterly / registered-only operators are unchanged.
- Out of scope (deferred, full PAE-1703): cancellation / reg-only transition, the additive quarterly split, and any registration-date / `validTo` back-trim. `generateReportingPeriods`' `fromDate` param (paired with the helper's `toDate`) leaves the obvious extension point for those.

## Changes

- `src/reports/domain/generate-reporting-periods.js` — optional `fromDate` param bounds the generated periods (front trim) via `filterPeriodsFromDate`.
- `src/domain/organisations/registration-utils.js` — `activeAccreditationValidFrom(accreditation)`: the `validFrom` of a live (approved/suspended) accreditation, else `null`.
- `src/reports/routes/get.js`, `src/reports/application/report-submissions.js`, `src/reports/application/report-compliance.js` — each resolves the accreditation `validFrom` and passes it to `generateReportingPeriods`.
- Tests: `generate-reporting-periods.test.js` (fromDate bound), `get.test.js` (AC1–3/5/6), `report-submissions.test.js` (AC4), `report-compliance.test.js` + `public-register/integration.test.js` (compliance/register bound), `registration-utils.test.js` (helper).
- `.vite/helpers/build-approved-org.js` — optional date-range override so tests can accredit an org from a chosen date.

## Notes

- The trim runs before `mergeReportingPeriods`, which still appends any period that already carries a persisted report — a real submission is never hidden.
- No defensive "missing validFrom" branch in the callers: `validFrom` is schema-required on approved/suspended accreditations, so the helper relies on that invariant.

[PAE-1737]: https://eaflood.atlassian.net/browse/PAE-1737?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/reports/routes/get.js
```diff
@@ -102,7 +106,14 @@ export const reportsGet = {
      * support once outstanding historical reports are submitted.
      */
     const currentYear = new Date().getUTCFullYear()
-    const computedPeriods = generateReportingPeriods(cadence, currentYear)
+    const allPeriods = generateReportingPeriods(cadence, currentYear)
+
+    // An accredited operator owes monthly reports only from the date their
+    // accreditation began, so trim periods that ended before validFrom.
+    const validFrom = activeAccreditationValidFrom(registration.accreditation)
```
senthilinfy, line 111: Might be cleaner to push this into `generateReportingPeriods` itself so both cadences get bounded consistently at the source — validFrom/validTo from the registration for registered-only, from the accreditation for accredited. 

`generateReportingPeriods`  is called in other places as well for report-compliance. Otherwise logic would needs to be duplicated

timrutter: good call, have updated

senthilinfy (approved):

src/reports/domain/generate-reporting-periods.js
```diff
@@ -47,18 +48,29 @@ export function generateAllPeriodsForYear(cadence, year) {
...
  *
  * @param {Cadence} cadence
  * @param {number} year
  * @param {Date} [now] - Current date (defaults to new Date(), injectable for testing)
+ * @param {string | null} [fromDate] - ISO `YYYY-MM-DD` lower bound; periods ending before it are dropped
  * @returns {Array<{year: number, period: number, startDate: string, endDate: string, dueDate: string, report: null}>}
  */
-export function generateReportingPeriods(cadence, year, now = new Date()) {
-  return generateAllPeriodsForYear(cadence, year).filter((p) => {
+export function generateReportingPeriods(
```
senthilinfy, line 63: It will be good to refactor `generateReportingPeriods(cadence,fromDate,toDate)`, so then from/toDate can passed in as needed for given reg/acc window. This is just suggestion, as existing change is fixing things in right direction

PR #1476, opened 2026-07-09
PAE-1593: Derive report detail from summary-log row states

GraemeF:
Jira: [PAE-1593](https://eaflood.atlassian.net/browse/PAE-1593)

Repoints the report-detail aggregation off the legacy waste-records collection onto the summary-log row states committed to the waste balance ledger, per ADR-0037 stage 2. The report detail page now reflects a registration's state at its latest submitted summary log rather than the last-written waste-record data.

## What changes

- **Aggregation source.** `aggregateReportDetail` now folds a registration's waste-record states at its latest submitted summary log — membership keyed by `summaryLogId` — instead of folding each waste record's `versions[]`. The read goes through the existing `summaryLogRowStatesForRegistration` read model.
- **Report source metadata.** Which submission produced the current state, and when it was submitted, is resolved from the latest `SUMMARY_LOG_SUBMITTED` ledger event by a new `latestSubmittedSummaryLog` helper, replacing the version-timestamp fold.
- **Domain types.** The pure aggregators (`aggregateReportDetail`, the received/exported/sent-on sub-aggregators, and `filter-records-by-date`) now consume `WasteRecordState`, reading `wasteRecordType` in place of the retired `type` field.
- **Route wiring.** The report GET-detail and POST routes now wire the waste balance ledger and summary-log row state repositories in place of the waste-records repository.

## Notes

Row-state tonnages are already coerced to two decimal places when written, so the read side gets round-each-then-sum for free. The now-redundant rounding scattered through the report aggregation is deliberately left in place and will be removed in a separate follow-up so this change stays a pure read cutover.

The blocking `Lint Types` job is clean. The advisory `Lint Types - Tests` job carries only pre-existing `exportActivity possibly undefined` debt in the aggregation tests (unchanged from main); the overall test-typecheck error count drops on this branch.

[PAE-1593]: https://eaflood.atlassian.net/browse/PAE-1593?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/waste-balances/application/latest-submitted-summary-log.js
```diff
@@ -0,0 +1,32 @@
...
+ * with when it was submitted (the `SUMMARY_LOG_SUBMITTED` event's timestamp).
+ * Returns `null` when the ledger has no summary-log submission yet.
+ *
+ * @param {import('../repository/ledger-port.js').WasteBalanceLedgerRepository} ledgerRepository
+ * @param {{ registrationId: string, accreditationId: string | null }} ledgerId
+ * @returns {Promise<{ summaryLogId: string, submittedAt: Date } | null>}
+ */
+export const latestSubmittedSummaryLog = async (
+  ledgerRepository,
+  { registrationId, accreditationId }
```
GraemeF: This should be a ledger id

src/reports/application/report-service.js
```diff
@@ -446,16 +466,36 @@ async function getAggregatedReportDetail({
   cadence,
   period
 }) {
+  const accreditationId = registration.accreditationId ?? null
+  const ledgerId = { organisationId, registrationId, accreditationId }
+
+  // One head resolution serves both reads: the row states and the source
```
timrutter, line 472: **Read-skew fix (applied in `d61d5c3f`):** previously the ledger head was resolved twice — once inside `summaryLogRowStatesForRegistration` and again in `resolveSource`. A submission committing between the two lookups would pair submission N's rows with submission N+1's `source` metadata on an operator-facing report. The head now resolves once and serves both reads; `wasteRecordStatesForHead` is exported from the read model for callers that must stay consistent with a head they already hold. There's a regression test (`derives rows and source from the same submission when a new submission commits mid-read`) that pins this with a ledger stub whose head moves between lookups. Also saves one ledger query per report generation.

src/reports/domain/aggregation/aggregate-report-detail.js
```diff
@@ -11,6 +11,15 @@ import { aggregateWasteExported } from './aggregate-waste-exported.js'
 import { aggregateWasteSentOn } from './aggregate-waste-sent-on.js'
 import { coerceWasteRecordsForRead } from './coerce-waste-record.js'
 
+/**
+ * The slice of a waste-record state the report aggregation reads: the row's
+ * type and its coerced data. Narrower than the full `WasteRecordState` so the
+ * type doesn't demand fields (classification, row identity) the aggregation
+ * never consumes.
+ *
+ * @typedef {Pick<import('#waste-records/application/read-summary-log-row-states.js').WasteRecordState, 'wasteRecordType' | 'data'>} ReportableWasteRecordState
```
timrutter, line 20: **Type honesty (applied in `05a5b568`):** the aggregators were typed against the full `WasteRecordState`, demanding `classification` and `rowId` they never read — a claim the JSON fixtures could only satisfy through the `unknown` double-cast. This `Pick` narrows the aggregation's declared input to the two fields it consumes, so the fixtures genuinely satisfy the contract and the type documents that the report depends only on row type + data (which is also what makes this PR independent of the recorded-vs-live classification decision).

timrutter (commented):
Reviewed while you're away by claude: Two findings, both already addressed in commits pushed to this branch — inline comments below explain each. Agreed with the PR's central claim: report detail never consumes waste-balance classifications (its only mutable-context input, the ORS details map, stays live and unchanged), so this cutover is genuinely independent of the recorded-vs-live decision.

timrutter (approved):

PR #1475, opened 2026-07-09
PAE-1339: [2/4] crush test type errors — builders, mocks, fixtures

loveboat:
Ticket: [PAE-1339](https://eaflood.atlassian.net/browse/PAE-1339)
batch 2 of `lint:types:tests` cleanup — **stacked on #1470** (base: `PAE-1339-test-types`). test-only, prod `[src]` unaffected. all touched files type-clean; suites green.

cumulative with #1470: `lint:types:tests` 485 -> 340.

files (8):
- collate-users (-33): typed org/registration/accreditation builders
- migration-orchestrator (-19): vi.mocked wraps + typed builders + repo casts
- classify-helpers (-17): typed accreditation + context helper
- seed-scenarios (-17): assert find results + widen orgIds
- parser-workarounds (-16): Buffer.from conversion + typed schema/classifier
- fail-action (-20): typed mock request + fake Joi errors
- get-entra-user-roles (-12): typed setListsExplicit params
- form-submissions/mongodb (-9): typed mongoIt fixtures + counters collection

types use @import blocks. review/merge #1470 first.

[PAE-1339]: https://eaflood.atlassian.net/browse/PAE-1339?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1479, opened 2026-07-09
PAE-1339: [3/4] crush test type errors — aggregation, validations, fixtures

loveboat:
batch 3 of `lint:types:tests` cleanup — **stacked on #1475**. test-only **except one prod change** (see below); prod `[src]` typecheck clean; all touched files type-clean; suites green.

cumulative across the stack: `lint:types:tests` 340 -> 164.

> **Prod change:** `src/domain/organisations/accreditation.js` — added `orsFileUploads?` and `glassRecyclingProcess?` to `AccreditationBase`. These were type drift: the ORS work added them to the accreditation schema + transformer + fixtures but not the JSDoc type. This is a correctness fix (type now matches what the code produces/validates), and it cleared the same latent errors in several test files.

~20 test files across: aggregation, process-import(+file), prn-tonnage, material/registration/processing-type validations, form-submissions in-memory + mongodb, prn + org-schema validation, sent-on-loads-schema, merge-reporting-periods, cdp-uploader, and the forms-submission JSON-import sweep.

patterns: typed test-data builders/fixtures at the boundary; `assert`/optional-chain narrowing; `vi.mocked()` wraps; `with { type: 'json' }` import attributes; fixed stale enum/signature drift. @import blocks throughout. review/merge #1470, #1475 first.


src/application/common/material-aggregation.test.js
```diff
@@ -7,6 +8,20 @@ import {
...
+ * `availableAmount`), so the returned entries can't be statically typed. This
+ * narrows a looked-up entry to the metric fields the tests assert on.
+ *
+ * @param {Array<{ material: string }>} materials
+ * @param {string} material
+ * @returns {{ totalTonnage: number, availableAmount: number }}
+ */
+const findMaterial = (materials, material) =>
+  /** @type {{ totalTonnage: number, availableAmount: number }} */ (
+    /** @type {unknown} */ (materials.find((m) => m.material === material))
```
dansweeting, line 22: `findMaterial` casts away the `undefined` that `.find()` can return, so a missed lookup fails with `Cannot read properties of undefined` rather than a clear assertion. This file already uses `assertPresent` further down; suggest using it here too for the same narrowing plus a legible failure.

loveboat: Done on #1483 (stack tip, to avoid a re-merge) — `findMaterial` now does `assertPresent(found)` then the metric cast, so a missed lookup fails legibly instead of `Cannot read properties of undefined`.

dansweeting (approved):
Comment actioned on the stack tip (#1483) and verified. Approving.

PR #1483, opened 2026-07-10
PAE-1339: [4/4] crush test type errors — tail sweep to zero + type-helpers

loveboat:

Final PR of the PAE-1339 `lint:types:tests` stack — **stacked on #1479**.

Cumulative across the stack: test type errors **748 -> 0 (100%)**. Suites green, 100% coverage on new code.

Mostly test-only — JSDoc casts/annotations plus a shared `#test/type-helpers.js` (`assertPresent`, `invalidArg`, `partialMock`).

It also makes a few small **prod type-accuracy fixes** (JSDoc-only, no runtime change) where the declared type had drifted from what the runtime genuinely produces or accepts:
- `StatusHistoryEntry.updatedAt` -> `Date | string` (`getStatusHistoryDateTimes` normalises both via `new Date(...)`)
- `Accreditation.glassRecyclingProcess` -> `... | null`, plus `orsFileUploads?` / `reprocessingType?` (backed by the accreditation Joi schema)
- `RegAccKeyFields.glassRecyclingProcess` -> `... | null` (`getRegAccKeyValuePairs` uses `?.length`)
- forms `BaseOrganisation.version?` (set on DB-fetched orgs by `prepareForReplace`)
- overseas `OverseasSiteAddress` `line2` / `stateOrRegion` / `postcode` -> `string | null` (schema `.allow(null)`)

These are schema/runtime-backed, not path-of-least-resistance: the runtime already produces these values and consumers already null-guard them (optional chaining), so there's no behaviour change and no null-safety follow-up needed. (`operator-category` was the counter-example — `accreditation` is required-nullable there, so the fixtures carry `accreditation: null` and the prod type is unchanged.)

review/merge #1470, #1475, #1479 first.


src/domain/organisations/accreditation.js
```diff
@@ -1,9 +1,16 @@
-/** @import {GlassRecyclingProcess, RegAccStatus, User} from '#domain/organisations/model.js' */
+/** @import {GlassRecyclingProcess, RegAccStatus, ReprocessingType, User} from '#domain/organisations/model.js' */
+
+/**
+/**
```
dansweeting, line 3: There are two consecutive `/**` openers here, which leaves the `updatedAt` rationale in a malformed comment block, detached from the `StatusHistoryEntry` typedef it justifies. Suggest collapsing to a single `/** ... */` and moving it directly above the typedef.

src/routes/health.test.js
```diff
@@ -6,7 +6,7 @@ import { setupAuthContext } from '#vite/helpers/setup-auth-mocking.js'
 describe('GET /health', () => {
   setupAuthContext()
   it('returns 200 with success message', async () => {
-    const server = await createTestServer({ skipMongoDb: true })
+    const server = await createTestServer()
```
dansweeting, line 9: `skipMongoDb` was a dead option (not read anywhere in `create-test-server.js`), so removing it is harmless, but this health-check test now spins up Mongo it does not need. Deliberate?

loveboat: Good to check, but it doesn't spin up Mongo. `createTestServer` is fully in-memory — it registers only the `createInMemory*RepositoryPlugin` set, with no MongoDB plugin, `MongoClient`, or container. In this suite Mongo only starts via the explicit per-file `mongoIt` fixture (`#vite/fixtures/mongo.js`, vitest-mongodb), and testcontainers are only used for sqs/redis/cdp-uploader — the health test uses none of them. `skipMongoDb` was dead precisely because there's no Mongo in this path, so dropping it is a pure no-op; the test wasn't (and isn't) starting Mongo.

src/domain/organisations/accreditation.js
loveboat: Fixed — collapsed to a single `/** ... */` block directly above the typedef.

dansweeting (approved):
All stack comments actioned and verified on this tip. Approving.

PR #1499, opened 2026-07-13
PAE-1698: Mark draft report as stale on PRN cancellation

senthilinfy:
Ticket: [PAE-1698](https://eaflood.atlassian.net/browse/PAE-1698)
See [ADR-0042](https://github.com/DEFRA/epr-re-ex-service/blob/main/docs/architecture/decisions/0042-stale-reports-on-prn-cancellation.md) for the design.

- Extends the existing summary-log staleness mechanism (PAE-1240) to also flag active draft reports stale when a PRN moves to `awaiting_cancellation`, triggered inline on the reject route rather than via a new event channel. `report.stale` becomes a container for two independent, possibly simultaneous reasons (`summaryLogChanged` / `prnCancelled`), each set by a single targeted write so neither trigger clobbers the other. `assertNotStale` now surfaces the applicable reasons as an array on the 409 payload.
- Generalises the PRN-cancellation notifier into a reusable PRN lifecycle event bus (`packaging-recycling-notes/application/prn-events.plugin.js`, `request.prnEvents.onCancelled`), moved out of `reports` and owned by the domain that raises PRN transitions. The route firing a PRN event shouldn't need to know reports is the only listener - a new consumer registers a handler rather than editing the route, and events fan out via `Promise.all`.

[PAE-1698]: https://eaflood.atlassian.net/browse/PAE-1698?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/packaging-recycling-notes/application/update-status.js
```diff
@@ -231,6 +260,69 @@ export async function updatePrnStatus({
...
+      : await performLedgerWrite(ctx)
+
+  await prnMetrics.recordStatusTransition({
+    fromStatus: currentStatus,
+    toStatus: newStatus,
+    material: prn.accreditation.material,
+    isExport: prn.isExport
+  })
+
+  await notifyPrnCancelled(prnEvents, newStatus, updatedPrn)
```
loveboat: **Report-staleness side-effect can fail a committed PRN cancellation**

`notifyPrnCancelled` runs *after* the ledger write has committed, and its rejection propagates out of `updatePrnStatus`. `onCancelled` fans out with no error boundary (`prn-events.plugin.js` — `await Promise.all(handlers.map(...))`, no try/catch), and the handler awaits **two mongo writes** (`prn-cancellation-events.js`):
- `reportsRepository.markActiveReportsStaleForPrnCancellation` (updateMany + find)
- `auditMarkReportsStale → recordSystemLogs → insertMany`

(The CDP `safeAudit` call is fire-and-forget/non-throwing, so it's not in the risk path — only the two awaited mongo writes are.)

**Failure scenario:** a transient mongo error on either write, in the window after the PRN transition committed, makes `updatePrnStatus` reject. Both callers (`external-transition-handler.js` reject route, `status.js`) then return **500 to the external caller on a PRN that is already `AWAITING_CANCELLATION`/`CANCELLED`**, with the report left un-flagged. Retry doesn't recover it — `validateTransition` rejects the re-attempt since the PRN already moved. So the caller sees a failure for an operation that actually succeeded, and the report stays stale-but-unmarked.

**On ADR-0042:** the ADR weighs inline-vs-SQS purely on happy-path latency ("the operator's next page load needs the flag to already be set, which only a synchronous write guarantees") — it doesn't address this commit-then-fail ordering. And that justification doesn't quite hold here: the reject/cancel is an external call by a *different actor* (RPD/producer) than the operator who owns the draft, so there's no read-your-writes requirement tying the write to a later page load. A few seconds of async lag would be fine.

**Suggestion:** make the fan-out best-effort — wrap `onCancelled` in try/catch, log, don't rethrow — so a soft side-effect's failure can't fail (or appear to undo) the authoritative PRN transition. `markActiveReportsStaleForPrnCancellation` is already idempotent, so an out-of-band reconcile stays safe. Ordering and error semantics were implementation choices here, not ADR mandates.

src/reports/domain/stale.js
```diff
@@ -2,20 +2,80 @@ import Boom from '@hapi/boom'
...
+  if (
+    reasons.length > 0 &&
+    report.status.currentStatus !== REPORT_STATUS.SUBMITTED
+  ) {
     const err = Boom.conflict(
-      'Report cannot be submitted: summary log has changed'
+      'Report cannot be submitted: summary log has changed or a PRN was cancelled'
     )
-    err.output.payload.code = report.stale.reason
+    err.output.payload.code = reasons
```
loveboat, line 109: **Minor:** ADR-0042 (line 36) recommends a Joi schema asserting `payload.code`'s new array shape *inside `assertNotStale`* "so `payload.code`'s new array shape is contract-tested rather than left implicit." Here it's just `err.output.payload.code = reasons` with no schema — the array shape is only covered by unit assertions, not a contract. Low priority, but it's a direct ADR item not picked up.

loveboat (approved):
Approving. Review feedback all addressed: PRN-cancellation notify is now best-effort (try/catch swallow in update-status.js + Promise.allSettled in the plugin), so a stale-notification failure no longer fails a committed transition; and payload.code is Joi-validated in assertNotStale per ADR-0042. Clean, well-tested change.

PR #1503, opened 2026-07-14
PAE-000: typecheck strictness — hard gate + soft burndown ratchet

loveboat:
raises typecheck strictness: a hard gate now + a soft burndown for ratcheting later, with prod and tests controlled the same way. (rolls up the baseline + strict PRs.)

hard / blocking (existing strictness):
- enable zero-cost tsc flags: noImplicitReturns, noFallthroughCasesInSwitch, noImplicitOverride, noUnusedParameters
- prod and tests typecheck now both run via the shared lint-types-tests action as their own full-suite hard-gate jobs (fail-on: all), each with its own sticky comment
- align eslint no-unused-vars to args: all (matches tsc; neostandard defaults to args: none)

structure + soft / non-blocking (stricter):
- extract jsconfig.typecheck.base.json (shared compilerOptions); prod/tests/strict configs extend it
- add jsconfig.typecheck.strict.json listing the candidate strictness levers
- add a non-blocking typecheck-burndown job that posts a ranked sticky comment: which levers are ready to enforce (0 errors) and which still need chipping away, per surface

graduating a lever = move its line from strict.json into base.json (both surfaces) or the prod/tests config (one surface). depends on the epr-re-ex-service PR (shared actions). the burndown job is pinned to the branch for preview; revert to @main before merge.


senthilinfy (approved):
Approved — pending epr-re-ex-service#349 merge before this can land.

PR #1510, opened 2026-07-15
PAE-1660: Retain original submission date on the public register

dansweeting:
Ticket: [PAE-1660](https://eaflood.atlassian.net/browse/PAE-1660)
## What

The public register embeds a submitted date per reporting period. It previously read the period's *current* report (`mergedPeriod.report`). While a period is being resubmitted the current report is an unsubmitted draft carrying no submitted date, so a period that had in fact been submitted transiently blanked on a public-facing artefact — and once a resubmission completed, the register showed the correction's date rather than the original.

This derives the register date from the **earliest** submitted report for the period instead of the current one:

- adds `originalSubmittedDate()`, which takes the first entry from `selectSubmittedReports` (ordered by submission number ascending)
- a period with an in-flight resubmission keeps showing its original submitted date instead of blanking
- a completed resubmission still shows the original submitted date, not the correction's
- a never-submitted period remains blank, unchanged

## Why

The business decision on PAE-1660 is that the public register retains the original submission date: resubmissions are not reflected externally, while a complete record of all resubmissions continues to be maintained in the backend (the report-submissions feed) for auditability. This may be revisited in future.

[PAE-1660]: https://eaflood.atlassian.net/browse/PAE-1660?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/reports/application/report-compliance.js
+++ b/src/reports/application/report-compliance.js
@@ -1,7 +1,10 @@
 import { CADENCE } from '#reports/domain/cadence.js'
 import { periodKey } from '#reports/domain/period-key.js'
 import { generateReportingPeriods } from '#reports/domain/generate-reporting-periods.js'
-import { mergeReportingPeriods } from '#reports/domain/merge-reporting-periods.js'
+import {
+  mergeReportingPeriods,
+  selectSubmittedReports
+} from '#reports/domain/merge-reporting-periods.js'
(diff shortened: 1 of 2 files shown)


loveboat (approved):
🥇

PR #1512, opened 2026-07-16
PAE-000: [1/4] type reports subsystem against ports (ts7 prep)

loveboat:
Ticket: [PAE-000](https://eaflood.atlassian.net/browse/PAE-000)
part 1/4 of splitting the typescript 6->7 (tsgo) bump (#1502) into reviewable, type-green slices.

points loose `@param {object}` deps in the reports subsystem at their existing repository port types, plus the coupled test-file fixes. behaviour-preserving - jsdoc/type-only, no runtime change. green under both ts6 (current main) and ts7.

- reports domain, routes, application and monitoring
- prod + coupled test-config fixes together
- stack base; #1502 to be superseded by the final bump slice

PR #1513, opened 2026-07-16
PAE-000: [2/4] type waste/public-register/adapter deps against ports (ts7 prep)

loveboat:
part 2/4 of splitting the typescript 6->7 (tsgo) bump (#1502) into reviewable, type-green slices. stacked on #1512.

points loose `@param {object}` deps at their repository port types across:
- application/public-register, waste-balances, waste-records, adapters
- summary-log route helpers
- shared test-infra typing (`#test/type-helpers.js`, `#test/create-test-server.js`)

behaviour-preserving, jsdoc/type-only. green under both ts6 and ts7.


src/application/waste-records/sync-from-summary-log.js
```diff
@@ -247,15 +250,15 @@ const calculateMetrics = (wasteRecords) => {
  * zero-delta — the submission is still recorded, it just moves no tonnage.
  *
  * @param {object} params
- * @param {object} params.summaryLog
+ * @param {{ file: { id: string }, organisationId: string, registrationId: string }} params.summaryLog
```
dansweeting, line 253: Not blocking for a 2/4 slice. `summaryLog` is shaped three ways in this file: untyped in `resolveAccreditationId` (line 126), `{ file: { id, uri }, organisationId, registrationId }` in the context builder (lines 197-202), and inline again here. If one named type (the summary-log document type, or a `Pick` of it) covers what they read, prefer that. Same for `ParserCollection` in `exceljs-parser.js`: defined but only applied to `createParserState`, not the `draftCollection` helpers that consume it.

loveboat: Done for `ParserCollection` in 4233e245 — applied it to the `draftCollection` helpers, fully typed their remaining params (adding `CellValue` and `UnfilledValues` typedefs for the repeated shapes), and lifted the cell-normalisation out of `processRowCell` into a named `normaliseCellValue` (which also dropped a misleadingly-named `trimmedValue` that wasn't always a string).

On the `summaryLog` shapes: I looked at collapsing them to one named type but a single strict type doesn't cleanly cover all three — `SummaryLog.file` is a `FileUpload | StoredFile` union (only `StoredFile` has `uri`, which the transform reads) and `accreditationId` isn't on the model, so it'd need the file variant narrowed at the submit boundary (à la `SubmittableSummaryLog`). Left as-is here and noting it as a follow-up rather than widening this typing slice.

dansweeting (approved):
ParserCollection typing sorted; happy to take the summaryLog unification as a follow-up. Approving.

PR #1514, opened 2026-07-16
PAE-000: [3/4] type overseas/prn/repository deps against ports (ts7 prep)

loveboat:
part 3/4 of splitting the typescript 6->7 (tsgo) bump (#1502) into reviewable, type-green slices. stacked on #1513.

points loose `@param {object}` deps at their repository port types across:
- overseas-sites, queue-consumer, packaging-recycling-notes
- organisations/summary-logs repositories + contract test-data
- auditing, common helpers, forms-submission-data migrations

also includes two summary-log integration tests (deferred here to travel with the organisations test-data builder they depend on) and a small dual-compat fix to a credited-tonnage test induced by the tightened builder.

behaviour-preserving, jsdoc/type-only. green under both ts6 and ts7.


src/auditing/helpers.js
```diff
@@ -3,29 +3,45 @@ import { logger } from '#common/helpers/logging/logger.js'
...
  */
 
 /**
  * @param {import('#common/hapi-types.js').HapiRequest} request
+ * @returns {SystemLogActor}
  */
 function extractUserDetails(request) {
-  return request.auth?.credentials?.isMachine
+  /** @type {MachineCredentials | HumanCredentials} */
+  const credentials = request.auth.credentials
```
dansweeting, line 28: `extractUserDetails` now reads `request.auth.credentials` unconditionally and branches on `'isMachine' in credentials`, where the old code used optional chaining. If `auth`/`credentials` is ever absent this throws instead of returning undefined fields. All current callers are authenticated routes, so low risk, but worth confirming: is there any path (pre-auth extension, error handler, audit on an unauthenticated request) that reaches this without credentials? Also `'isMachine' in credentials` is a key-existence check, safe only because `MachineCredentials.isMachine` is typed `true`.

src/server/run-duplicate-accreditation-link-migration.js
```diff
@@ -237,16 +243,15 @@ const processOrganisation = async ({
 }
 
 /**
- * @param {object} server
+ * @param {Server} server
```
dansweeting, line 246: `@param {Server}` is the stock hapi Server (no `db`/`locker`/`featureFlags`), so laundering through `unknown` to `MigrationServer` is self-inflicted. `HapiServer` (`#common/hapi-types.js`) already declares those fields and `MigrationServer` is built from it. Change `@param {Server}` to `@param {HapiServer}` and cast straight to `MigrationServer`, dropping the `unknown` step. Same at the second function (lines 288-292).

src/common/helpers/collections/seed-scenarios.js
```diff
@@ -274,14 +276,13 @@ async function buildActiveOrgWithSuspendedAccreditation(
     (r) => r.status === REG_ACC_STATUS.APPROVED
   )
   const matchingAccIndex = org.accreditations.findIndex(
-    (acc) => acc.material === approvedReg.material
+    (acc) => acc.material === approvedReg?.material
```
dansweeting: `approvedReg?.material` silences a case the comment above says can't happen. If the invariant ever breaks, this yields `undefined` → `matchingAccIndex` = -1 → silently seeds an org with no suspended accreditation, where before it threw. Seed-only so small blast radius, but prefer asserting the invariant (e.g. `assertPresent`) over `?.` so a broken guarantee fails loudly.

src/overseas-sites/application/process-import-file.js
```diff
@@ -3,8 +3,20 @@ import { SpreadsheetValidationError } from '#adapters/parsers/summary-logs/excel
...
- * @import { SystemLogsRepository, SystemLogHumanActor } from '#repositories/system-logs/port.js'
  * @import { TypedLogger } from '#common/hapi-types.js'
+ * @import { OverseasSitesRepository } from '#overseas-sites/repository/port.js'
+ * @import { OrganisationsRepository } from '#repositories/organisations/port.js'
+ * @import { SystemLogsRepository, SystemLogHumanActor } from '#repositories/system-logs/port.js'
+ */
+
+/**
+ * @typedef {{
+ *   errors: Array<object>
```
dansweeting: Every error produced here is `{ field?: string, message: string }` (see `failureResult`), so the bare `object` forces a cast in the test at `process-import-file.test.js`. Tighten it and the cast goes away:

```suggestion
 *   errors: Array<{ field?: string, message: string }>
```

loveboat: Done in 212a0754 — tightened to `Array<{ field?: string, message: string }>` and dropped the now-unnecessary cast in the test.

src/common/helpers/collections/seed-scenarios.js
loveboat: Done in 212a0754 — swapped `?.` for a `node:assert` on the invariant so a broken guarantee fails loudly. Went with `assert` rather than a bare `throw` because it stays a single always-executed statement (the throw lives inside the library), keeping the file at 100% coverage — a hand-rolled `if (!x) throw` adds a branch the tests never hit since the invariant always holds.

src/auditing/helpers.js
loveboat: Confirmed — all ~13 callers are authenticated audit paths, and `HapiRequest.auth.credentials` is typed non-optional (`MachineCredentials | HumanCredentials`), so credentials is always present; there's no pre-auth/error-handler/unauthenticated path that reaches `extractUserDetails` without it, and no test exercises a credential-less call. And yes — `'isMachine' in credentials` is safe precisely because `MachineCredentials.isMachine` is typed `true` and `HumanCredentials` has no such key. Leaving as-is; restoring the old fallback (undefined-valued actor) would be type-invalid against `SystemLogActor` anyway.

src/server/run-duplicate-accreditation-link-migration.js
loveboat: Held off on this one. The caller (`server.js` `onPostStart`) passes the stock hapi `Server`, which isn't assignable to the hand-rolled `HapiServer` typedef (required `featureFlags`, app-specific `app`/`logger`/`events`) — retyping the param to `HapiServer` breaks that call site and all the mock-server tests. And `Server -> MigrationServer` genuinely needs the `unknown` step under tsgo (TS2352, "neither type sufficiently overlaps"). So the launder is load-bearing: it's the single bridge from the runtime stock `Server` (decorated with db/locker/featureFlags only at runtime) to `MigrationServer`. Open to a cleaner idiom if you have one, but `HapiServer` on the param doesn't compile.

src/overseas-sites/application/process-import-file.js
loveboat: Refined in 8e45dd4c — extracted a named `OrsFileError` typedef (`{ field?: string, message: string }`) reused across `OrsFileResult.errors` (`OrsFileError[]`) and `failureResult`'s param, rather than the inline object. No existing type fit — `ValidationErrorIssue` is a richer, different shape. (Aside on `Array<>` vs `[]`: repo convention is `Array<…>` for inline object element types, `X[]` for primitives/named types — hence `OrsFileError[]`.)

dansweeting (approved):
Accepting the reasoning on the HapiServer/unknown launder — the rest is actioned. Approving.

PR #1515, opened 2026-07-16
PAE-000: [4/4] bump typescript to 7.0.2 + submit/ors domain narrowing

loveboat:
part 4/4 (capstone) of splitting the typescript 6->7 (tsgo) bump (#1502) into reviewable slices. stacked on #1514. this is the slice that flips the checker to ts7; with #1512-#1514 merged the whole repo is type-green (prod + tests, 0 errors).

**the bump**
- typescript 6.0.3 -> 7.0.2, lockfile regenerated on current main

**domain decisions** (the errors that needed judgement, not mechanical porting):
- `submit.js`: introduce `SubmittableSummaryLog` - a summary log in SUBMITTING status always has `organisationId`/`registrationId`/`createdAt` (optional earlier in the lifecycle). narrowed by cast after the status check; no runtime change (deliberately not a throwing guard - established behaviour is to proceed).
- `summary-logs` metrics: `processingType` made optional on `ProcessingTypeDimensions` - it is genuinely absent for some log types and `buildDimensions` already tolerates undefined. matches `StatusTransitionDimensions`.
- `ors-import-commands.js`: a userless ORS import is genuinely reachable (the dispatcher enqueues `{ importId }` with no user), so rather than cast, `execute` now defaults `payload.user ?? SYSTEM_USER` - a shared system actor extracted from the two migrations that had it copy-pasted. schema stays `.optional()` (a `.required()` schema would silently *drop* userless messages at the consumer, not process them). userless imports now write a system-attributed audit record instead of `createdBy: undefined`.
- `add-or-update` test: drop a now-unused `@ts-expect-error` (ts7 errors on unused directives). `mockOrganisation` is inferred as `Object & Partial<Organisation>`, so `users` was already optional and was never checked against the real `Organisation` type; runtime coverage stays intact via the `replace` assertions.
- small dual-compat cast in a credited-tonnage ledger test.

all tests pass, 100% coverage. supersedes #1502.


src/server/queue-consumer/ors-import-commands.js
```diff
@@ -48,10 +61,13 @@ export const importOverseasSitesHandler = {
       systemLogsRepository: deps.systemLogsRepository,
       logger: deps.logger,
       orsImportMetrics,
-      user: payload.user
+      user: /** @type {SystemLogHumanActor} */ (payload.user)
```
dansweeting: This casts `payload.user` to `SystemLogHumanActor` while the schema keeps `user` optional (line 50) and the dispatcher deliberately enqueues without a user: `const payload = user ? { importId, user } : { importId }` (sqs-command-executor.js:143). So a schema-valid message with no user makes `processOrsImport` receive `user: undefined` typed as present, which lands as `createdBy: undefined` in the system log (only warns, never errors) — a silent gap in the audit record.

Two options: tighten the schema to `userSchema.required()` (which also makes the dispatcher's `user ?` branch dead), or type `deps.user` as `SystemLogHumanActor | undefined` and handle absence. Either way the PR-body rationale ("joi `.optional()` is lenient") isn't the real reason it's safe — that's the single caller passing a stamped `createdBy`, which is exactly what a second caller could break.

src/routes/v1/organisations/user/add-or-update-organisation-user.test.js
```diff
@@ -275,7 +275,6 @@ describe('addOrUpdateOrganisationUser', () => {
       await addOrUpdateOrganisationUser(
         mockRequest,
         mockTokenPayload,
-        // @ts-expect-error intentionally testing an organisation with no users property
         mockOrganisation
```
dansweeting, line 278: Removing the now-unused `@ts-expect-error` is correct (ts7 errors on unused directives). Flagging only that the PR body's reason ("ts7 no longer flags the call") isn't quite it: `mockOrganisation` is inferred as `Object & Partial<Organisation>`, so `users` is already optional here and the arg was never checked against the real `Organisation` type. Runtime coverage is intact (both tests still assert `replace` is called with a freshly built `users` array), it just isn't obvious from the diff — worth a line in the description.

src/server/queue-consumer/ors-import-commands.js
loveboat: Good catch on the dispatcher — you're right the userless path is real. Fixed in ead8a9ad by **handling absence** rather than tightening the schema. Tightening to `required` would actually be worse: `consumer.js` validates the payload and on failure just error-logs and `return null` (drops the message) — it's not routed to `onFailure` — so a required-user schema would silently drop the import instead of processing it.

So: schema stays `.optional()` (matches the dispatcher's `user ?` branch), and `execute` now does `user: payload.user ?? SYSTEM_USER`, where `SYSTEM_USER` is a shared system actor I extracted from the two migrations that had it copy-pasted. A userless import now writes a **system-attributed** audit record instead of `createdBy: undefined`. Added execute tests for both branches. (Confirmed the actor isn't schema-validated on insert — `createdBy` is `any`, and `normaliseActorRole` passes non-human actors straight through.) Also corrected the PR body.

src/routes/v1/organisations/user/add-or-update-organisation-user.test.js
loveboat: Thanks — you're right, updated the PR description. The real reason is that `mockOrganisation` is inferred as `Object & Partial<Organisation>`, so `users` was already optional and the arg was never checked against the real `Organisation` type; removing the directive is just the ts7 unused-directive fix, and runtime coverage stays intact via the `replace` assertions.

dansweeting (approved):
Both comments actioned. Approving.

PR #1518, opened 2026-07-17
PAE-1755: Fix report stale-status normalisation dropping legacy summary-log data

senthilinfy:
Ticket: [PAE-1755](https://eaflood.atlassian.net/browse/PAE-1755)
## Summary
- Fixes a prod 500 (`"stale.uploadedAt" is not allowed`) on report reads for reports marked stale before PAE-1698 (old flat `stale` shape) and then re-flagged afterwards, which left legacy fields sitting alongside the new nested shape.
- `normaliseStale` now correctly upgrades/strips the legacy flat shape (`{ uploadedAt, reason, summaryLogId }`) to the current nested shape (`{ summaryLogChanged?, prnCancelled? }`) wherever a report is read, so the strict response schema never sees the old fields.

Ref: PAE-1755

[PAE-1755]: https://eaflood.atlassian.net/browse/PAE-1755?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/reports/domain/stale.js
+++ b/src/reports/domain/stale.js
@@ -2,6 +2,10 @@ import Boom from '@hapi/boom'
 import Joi from 'joi'
 import { REPORT_STATUS } from '#reports/domain/report-status.js'
 
+/**
+ * @import { ReportStale, StaleSummaryLogChanged } from '#reports/repository/port.js'
+ */
+
 export const STALE_REASON = Object.freeze({
   SUMMARY_LOG_CHANGED: 'summary_log_changed',
   PRN_CANCELLED: 'prn_cancelled'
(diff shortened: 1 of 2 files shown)


dansweeting (approved):

loveboat (approved):

src/reports/domain/stale.js
```diff
@@ -44,34 +44,27 @@ export const staleReasons = (stale) => {
...
- * flat `{ uploadedAt, reason, summaryLogId }` object; new writes only ever
- * produce `{ summaryLogChanged?, prnCancelled? }`. Applied at the repository
- * read boundary so every caller sees the current shape regardless of when
- * the document was written — no bulk migration needed, since `stale` only
- * ever applies to active drafts, which are short-lived by construction.
+ * Normalises a report's `stale` field to the current nested shape.
+ * Upgrading the legacy flat shape (`{ uploadedAt, reason, summaryLogId }`) where needed.
  *
- * @param {Record<string, unknown> | undefined} stale
+ * @param {any} stale
```
loveboat: Nice — this drops the `any` casts, but the swap trades them for a couple of things that cut against our conventions:

- `@param {any} stale` weakens the param (it was `Record<string, unknown> | undefined`). We avoid bare `any` — a single whole-object cast on the return keeps it type-safe without per-field casts.
- `if (!stale) return undefined` — single-line `if` without braces.
- The doc comment lost the root-cause explanation (the dot-path `$set` merge) — that's the bit worth keeping for the next reader.

Suggested form, `any`-free and keeping the why. Add an `@import` block at the top of the file (no inline `import()`):

```js
/**
 * @import { ReportStale, StaleSummaryLogChanged } from '#reports/repository/port.js'
 */
```

then:

```js
/**
 * Normalises a report's `stale` field to the current nested shape. Old
 * documents carry a flat `{ uploadedAt, reason, summaryLogId }` object; new
 * writes only ever produce `{ summaryLogChanged?, prnCancelled? }`. Applied at
 * the repository read boundary so every caller sees the current shape.
 *
 * Always rebuilds from just the two known nested fields — a dot-path `$set`
 * (e.g. `stale.summaryLogChanged`) merges into an existing `stale` rather than
 * replacing it, so an old-shape doc touched again can carry both the legacy
 * keys and the new nested key. Picking only the known fields drops the legacy
 * leftovers before they hit the strict response schema.
 *
 * @param {Record<string, unknown> | undefined} stale
 * @returns {ReportStale | undefined}
 */
export const normaliseStale = (stale) => {
  if (!stale) {
    return undefined
  }

  const { summaryLogChanged, prnCancelled, reason: _reason, ...rest } = stale

  if (summaryLogChanged || prnCancelled) {
    return /** @type {ReportStale} */ ({
      ...(summaryLogChanged ? { summaryLogChanged } : {}),
      ...(prnCancelled ? { prnCancelled } : {})
    })
  }

  return {
    summaryLogChanged: /** @type {StaleSummaryLogChanged} */ (rest)
  }
}
```

Not blocking — logic and tests are correct.

senthilinfy: Applied — dropped the `any` casts (kept a single whole-object cast on the return, per the suggested form), added the `@import` block, and restored brace style on the early return. Kept the shorter doc comment for now rather than the longer root-cause version — happy to expand it back if that's preferred.

PR #1519, opened 2026-07-17
PAE-1755: enrich report response-validation logging

loveboat:
Ticket: [PAE-1755](https://eaflood.atlassian.net/browse/PAE-1755)

stacked on #1518 (the fix). two observability outcomes from the `stale.uploadedAt` 500:

- **A — response-validation failures:** report GET/POST routes used Hapi's built-in `failAction: 'error'`, which logs only the first Joi message (`"stale.uploadedAt" is not allowed`) — no report id, no failing field. added a shared `reportResponseFailAction` that throws a `CdpBoom` carrying a searchable `code` + `event.action` and a PII-safe `reason` (route params + Joi field-path:rule-type). wired into `get-detail`, `post`, `submissions`.
- **B — legacy-stale read signal:** `mapReport` now warns when it normalises a legacy/hybrid `stale` doc on read (`legacy_stale_shape_normalised`, with report id + stripped keys), so the migration tail is observable and the back-compat branch in `normaliseStale` is safe to time for removal.

pure detector `legacyStaleKeys` added to the stale domain. new event actions in `common/enums/event.js`.

[PAE-1755]: https://eaflood.atlassian.net/browse/PAE-1755?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/reports/repository/helpers.js
+++ b/src/reports/repository/helpers.js
@@ -16,9 +16,10 @@ const BARE_DATE_LENGTH = 10
 
 /**
  * Normalises a report's `stale` field to the current nested shape on read.
- * Warns when the document still carries the legacy flat shape, so the number
- * of un-migrated documents is observable — the back-compat branch in
- * `normaliseStale` is safe to remove once this stops firing (PAE-1755).
+ * Logs at info when the document still carries the legacy flat shape, so the
+ * number of un-migrated documents is observable — the back-compat branch in
+ * `normaliseStale` is safe to remove once this stops firing (PAE-1755). Info,
+ * not warn: old documents are expected, so this must not raise alerts.
  *
  * @template {{ id?: string, stale?: Record<string, unknown> }} T
  * @param {T} report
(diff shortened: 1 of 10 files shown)


src/reports/repository/helpers.js
```diff
@@ -4,18 +4,43 @@ import {
...
-export const mapReport = (report) =>
-  report.stale ? { ...report, stale: normaliseStale(report.stale) } : report
+export const mapReport = (report) => {
+  if (!report.stale) {
+    return report
+  }
+  const strippedKeys = legacyStaleKeys(report.stale)
+  if (strippedKeys.length > 0) {
+    logger.warn({
+      message: 'Normalised legacy stale shape on report read',
```
senthilinfy, line 35: Better to log as info as warn will get alerted for old docs?

senthilinfy (approved):

src/reports/repository/helpers.js
loveboat: Yep, good point - lets keep those alerts down!

PR #1520, opened 2026-07-17
PAE-1747: Startup diagnostic for pre-CPA reports needing resubmission

dansweeting:
Ticket: [PAE-1747](https://eaflood.atlassian.net/browse/PAE-1747)
## Why

Closed-Period Amendment (CPA) is the rule that flags a submitted report as *requires resubmission* when a later summary-log upload changes waste-record data already folded into that report for a now-closed period. CPA is built but not yet enabled in production (its feature flag is off), so no report has ever been flagged this way in prod, even though the pattern has occurred many times in the historical data.

Before CPA is switched on, we need to size the affected population: how many already-submitted reports CPA *would* flag as requiring resubmission. That sizing informs whether to retrospectively backfill the flag when CPA goes live.

## What

Adds a read-only startup diagnostic, feature-flagged **off** by default, that scans the estate and logs the population of existing reports CPA would flag. It writes nothing and backfills no flags: enabling it in an environment only produces sizing figures in the logs.

- **Domain finder** (`src/reports/monitoring/pre-cpa-resubmission.js`): reconstructs the CPA rule from the summary-log row-state collection (ADR-0037). Per registration it diffs consecutive submitted uploads, maps each changed row to its period, and records a finding when that period's report was already submitted. Faithful to CPA, so it catches zero-net adjustments and oscillations — a retrospective count, not a prediction.

- **Startup runner** (`src/server/run-pre-cpa-resubmission-report.js`): flag-gated, under a cross-instance lock. Logs each affected report, a sizing summary (scanned / findings / affected orgs / registrations), an invariant probe, and a warning for any submitted report missing its timestamp (excluded from sizing, not dropped).

- **Feature flag** `preCpaResubmissionReport` (`FEATURE_FLAG_PRE_CPA_RESUBMISSION_REPORT`), default off, wired through config, both feature-flag adapters and the port type.

- **Server wiring**: invoked from `onPostStart` alongside the existing startup diagnostics.

## Sample log output

Captured by driving the real `runPreCpaResubmissionReport` over a fake estate. Each line is emitted as `logger.info({ message })` / `logger.warn({ message })`, so in CDP these appear as structured ECS JSON with the string below as the `message` field. Identifiers and dates are illustrative.

**Affected reports found** (one line per report, then the sizing summary, then a clean invariant probe):

```
INFO   Pre-CPA resubmission (retrospective): org a1b2c3d4-0001-4000-8000-000000000001 / registration reg-plastic-4471, report rep-plastic-jun26 (Jun 2026, monthly) — closed period restated by summary log SL-plastic-0011 uploaded after the report was submitted 2026-07-01T09:14:00.000Z
INFO   Pre-CPA resubmission (retrospective): org a1b2c3d4-0001-4000-8000-000000000001 / registration reg-paper-4472, report rep-paper-may26 (May 2026, monthly) — closed period restated by summary log SL-paper-0009 uploaded after the report was submitted 2026-06-10T11:00:00.000Z
INFO   Pre-CPA resubmission (retrospective): org f9e8d7c6-0002-4000-8000-000000000002 / registration reg-glass-5510, report rep-glass-apr26 (Apr 2026, monthly) — closed period restated by summary log SL-glass-0006 uploaded after the report was submitted 2026-05-02T10:30:00.000Z
INFO   Pre-CPA resubmission sizing: scanned 3 submitted reports, 3 would require resubmission, across 2 organisations / 3 registrations. Retrospective — not a prediction of the next upload.
INFO   Invariant check: 0 IGNORED restatements fell in a closed reported period (expected 0)
```

**Clean estate** (nothing to resubmit — just the summary and the probe):

```
INFO   Pre-CPA resubmission sizing: scanned 1 submitted reports, 0 would require resubmission, across 0 organisations / 0 registrations. Retrospective — not a prediction of the next upload.
INFO   Invariant check: 0 IGNORED restatements fell in a closed reported period (expected 0)
```

**Data anomalies surfaced, not swallowed** (an outside-accreditation restatement in a closed reported period, and a submitted report with no submission timestamp — both warned with their report ids):

```
INFO   Pre-CPA resubmission sizing: scanned 1 submitted reports, 0 would require resubmission, across 0 organisations / 0 registrations. Retrospective — not a prediction of the next upload.
WARN   Invariant check: 1 IGNORED restatements fell in a closed reported period (expected 0) — reports rep-ignored-jun26
WARN   Data integrity: 1 submitted reports missing a submittedAt were skipped from sizing — reports rep-nodate-mar26
```

## Scope and known limitations

The diagnostic is a sizing tool, not a data migration. Two edge cases are knowingly out of scope because they cannot change the flag-off, read-only outcome and are of low, uncertain incidence:

- A registration with multiple historical accreditation ids (re-accreditation, or accredited-then-cancelled): only the current and registered-only ledgers are read.
- A cancelled accreditation that retains its number: its template resolves to monthly cadence while the report cadence is quarterly.

Both are documented in the finder's JSDoc.

## Performance note

For each registration that has submitted reports, the scan runs a short series of MongoDB queries — one to look up the registration, one to load its summary logs, and one per submitted log (per ledger) to load that log's row states. They run one after another and are all indexed, so the runtime comes from the number of queries rather than the cost of each. The in-memory comparison is cheap: the row-state documents are small, and only one registration's data is held at a time.

This is fine as shipped. The diagnostic is fire-and-forget (started from `onPostStart` without `await`, so it never blocks startup or healthchecks), runs under a lock so only one instance per deploy executes it, and is off by default behind a flag — a one-off sizing run, not a recurring job. On a large estate it could still take a few minutes, holding the lock for that time.

If that ever becomes a problem, the simplest fix is to process several registrations at once (say 10–20) rather than one at a time, which would cut the run time roughly in proportion with no database changes. It is left out here to keep a throwaway diagnostic simple — the durable fix is to retire this module once CPA detection itself moves onto row states.

[PAE-1747]: https://eaflood.atlassian.net/browse/PAE-1747?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


loveboat (changes requested):
Non-ASCII em-dashes (—) in emitted log strings and source

loveboat (approved):

PR #1533, opened 2026-07-20
PAE-1764: Reclassify summary log row states live at read time

GraemeF:
Ticket: [PAE-1764](https://eaflood.atlassian.net/browse/PAE-1764)

A committed row state carries the waste-balance classification stamped when it was submitted. That reading is the right answer for reproducing the credit its submission committed, and the wrong answer for any question about the present: approving an overseas site or amending an accreditation's validity period changes which rows count, and nothing restates the stored reading until the operator submits again.

This adds the read path that re-derives classification against current rules and reference data, in the shape consumers already receive, so a consumer moves from stamped to live readings by changing which function it calls.

## What this adds

`reclassifyWasteRecordStates(states, { processingType, accreditation, overseasSites })` is pure: it replaces each row's stamped classification with one derived from the context it is given, and returns rows in the existing `{ rowId, wasteRecordType, data, classification }` shape.

`liveClassifiedRowStatesForRegistration({ ...ledgerId, repositories })` composes that with the reads the context comes from. It is a drop-in for `summaryLogRowStatesForRegistration` — same arguments plus the repositories, same return shape.

Neither has a production caller yet; the consumers are separate tickets.

## Where the template and accreditation come from

The reporting template is taken from the row states, which record the template they were submitted under — a summary log is one uploaded workbook, and every row in it reports under that workbook's template. The accreditation is resolved from the ledger partition being read.

Neither is expected to differ from what the registration currently names. If it turns out they should track the registration instead, that is a small change to this one function.

## Supporting changes

`classifyRecordForWasteBalance` now takes `processingType` as an explicit argument instead of reading it back out of `data`. The processing type names the template a registration reports under rather than describing the load, so it arrives as context. The parameter is also narrowed to the record fields the classifier actually reads — its type and data.

The waste-balance classifiers no longer require `ROW_ID` among the fields they check. A row's identity column is not part of its content, and the stored projection strips `ROW_ID` in favour of the top-level `rowId`, so re-classifying stored data would otherwise report the identity column as a missing required field. `ROW_ID` remains in `requiredHeaders` and in upload-time validation, where it belongs.

`processingType` is narrowed from `string` to the template union in the stored row-state typedefs.

## Reads this changes today

None. The write path is untouched — at submission time the workbook is the authority for the template, which is exactly what it already reads.

[PAE-1764]: https://eaflood.atlassian.net/browse/PAE-1764?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/waste-records/application/reclassify-waste-record-states.js
```diff
@@ -0,0 +1,42 @@
...
+ */
+export const reclassifyWasteRecordStates = (
+  states,
+  { processingType, accreditation, overseasSites }
+) =>
+  states.map((state) => ({
+    ...state,
+    classification: classifyRecordForWasteBalance(
+      { type: state.wasteRecordType, data: state.data },
+      processingType,
```
crosslandwa, line 38: whys is this from a closure, rather than `state.processingType`?
 
in the calling code the value passed (as part of the second parameter to `reclassifyWasteRecordStates`) comes from

```
const [{ processingType }] = rowStates
```

this approach _looks like_ it adds extra complexity (or am I missing something?)

crosslandwa (approved):

src/waste-records/application/reclassify-waste-record-states.js
GraemeF: Went back and forth on this a bit and probably ended halfway :laughing: 

Some of this stuff deals with spreadsheet rows rather than domain objects, so I'm trying to move it to a richer domain model and separate the identity from the content. The row id and processing type are not really properties of the row, they're identity that was bundled in with the content. I'll make sure it gets better next time...

PR #1536, opened 2026-07-20
PAE-1652: Add operator-initiated report resubmission

senthilinfy:
Ticket: [PAE-1652](https://eaflood.atlassian.net/browse/PAE-1652)
## Summary
- Nest `resubmissionRequired` by trigger (`closedPeriodRestated`, `operatorRequested`) so independent triggers can coexist
- Add `markSubmittedReportRequiringResubmissionByOperator` repository method, conditional on submitted status and not already flagged
- Add `POST .../submissions/{submissionNumber}/request-resubmission` endpoint, operator-scoped, gated behind `closedPeriodAdjustments`
- Surface `canRequestResubmission` on the report detail response, derived from the same eligibility rules the endpoint enforces

See ADR-0043: https://github.com/DEFRA/epr-re-ex-service/blob/main/docs/architecture/decisions/0043-operator-initiated-report-resubmission.md

[PAE-1652]: https://eaflood.atlassian.net/browse/PAE-1652?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


loveboat (commented):
A few non-blocking observations from a read-through of the resubmission flow. The migration risk I first suspected around the `resubmissionRequired` reshape checks out fine — CPA has never been enabled in prod (`FEATURE_FLAG_CLOSED_PERIOD_ADJUSTMENTS` unset, convict default false), so there is no legacy flat-shaped data to migrate. The three inline notes below are the ones worth a look; only the first is really about avoidable work on a hot path.

src/reports/routes/get-detail.js
```diff
@@ -21,6 +22,72 @@ import { SCOPES } from '#common/helpers/auth/constants.js'
...
+  reportsRepository,
+  report,
+  organisationId,
+  registrationId
+) {
+  if ('diagnostics' in report) {
+    return false
+  }
+
+  const periodicReports = await reportsRepository.findPeriodicReports({
```
loveboat: **Duplicate `findPeriodicReports` on the report-detail hot path.**

This fetches `findPeriodicReports` a second time per detail view. `fetchOrGenerateReportForPeriod` -> `fetchReportBySubmissionNumber` (`report-service.js:118`) already calls it before this handler runs, so a stored-report GET now runs that per-registration aggregation twice.

Two things:
- It's **unconditional on the feature flag**. `resubmissionEligibility` returns `FEATURE_DISABLED` on its first line, so with `closedPeriodAdjustments` off (the prod default today) this second aggregation is fetched and immediately discarded on every detail view. An `isClosedPeriodAdjustmentsEnabled()` short-circuit before the fetch removes that.
- Even with the flag on it's a duplicate. `canRequestResubmission(periodicReports, ...)` / `isLatestSubmissionOf` were deliberately extracted to run against "a periodicReports list already fetched once by the caller" (per their JSDoc), but the route re-queries rather than threading through the list `fetchReportBySubmissionNumber` already loaded.

Not a blocker, but `findPeriodicReports` is a full aggregation rather than a point read, so worth avoiding the extra round-trip.

src/reports/routes/request-resubmission.js
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * Lets an operator self-trigger resubmission on their own submitted report.
+ * Operator-scoped, unlike `unsubmit` which is admin-scoped.
+ */
+export const reportsRequestResubmission = {
+  method: 'POST',
+  path: reportsRequestResubmissionPath,
+  options: {
+    auth: getAuthConfig([SCOPES.organisationWrite]),
```
loveboat, line 20: **Endpoint is gated on `closedPeriodAdjustments` only, not the operator UI flag.**

`requestOperatorResubmission` -> `resubmissionEligibility` checks only `isClosedPeriodAdjustmentsEnabled()`. `operatorInitiatedResubmission` lives solely in frontend config, so enabling `closedPeriodAdjustments` in prod opens this POST endpoint independently of the UI flag — the frontend flag can be off while the endpoint is callable.

Low risk given it's `organisationWrite`-scoped to the operator's own org and the write is idempotent (nicely covered by the 403-for-service-maintainer test). Flagging mainly so it's a conscious decision: the two flags can't be reasoned about as one gate. A line in ADR-0043 noting the backend intentionally has no knowledge of the UI flag would cover it.

src/reports/routes/request-resubmission.js
```diff
@@ -0,0 +1,72 @@
...
+      organisationId,
+      registrationId,
+      year,
+      cadence,
+      period,
+      submissionNumber,
+      requestedBy: extractChangedBy(request.auth.credentials)
+    })
+
+    await auditReportRequestResubmission(request, {
```
loveboat, line 52: **Audit runs after the flag write, awaited.**

If `recordSystemLog` throws here, the report is already flagged but the caller gets a 500. On the frontend the POST catches only 409, so a 500 surfaces as an error page; a retry then hits `canRequestResubmission === false` and the make-changes GET 404s, so that specific entry point is gone (the period still shows `REQUIRES_RESUBMISSION`, so the operator can still resume via the reports list — they aren't stranded).

This matches the existing `auditReportCreate` ordering, so not a regression — just more visible now that success redirects into a draft. Noting as a possible follow-up in case audit ordering/robustness is worth revisiting for the self-service flows.

loveboat (approved):

src/reports/routes/request-resubmission.js
senthilinfy: This was intentional to restrict new flag to frontend alone. As there's no other way other than FE to reach the end point. 

I agree there's a risk, but think its acceptable for the short lived feature flag

src/reports/routes/get-detail.js
senthilinfy: really good point. addressing it. I'ts not just duplication/efficiency. The canRequestResubmissionFlag field should be returned on all fetches inside fetchOrGenerateReportForPeriod. Otherwise the field is not returned for all report fetch's via report-service.

src/reports/routes/request-resubmission.js
senthilinfy: I will leave this as its for now. lets tackle it as follow up if you don't mind

PR #1540, opened 2026-07-21
PAE-1765: Classify the waste records export live at read time

GraemeF:
Ticket: [PAE-1765](https://eaflood.atlassian.net/browse/PAE-1765)
The waste records CSV/FOI export resolves the accreditation and the registration's overseas sites as they stand at export time, but since it moved onto the summary log row states collection it rendered the waste-balance classification stamped when each row was submitted. The two drift apart: amend an accreditation's validity period or approve an overseas site, and a row can read "Accredited: Yes" beside "Included in Waste Balance: false — OUTSIDE_ACCREDITATION_PERIOD", contradicting itself within one line of the extract.

Each row's classification is now re-derived through `reclassifyWasteRecordStates` against the same context the surrounding columns read, so the whole row answers as of the same moment. No rows are added or dropped — the classification feeds the three waste-balance columns rather than a filter, so only their values change. The stamped classification stays in the collection as the record of the credit each submission committed.

The registration's processing type is hoisted once as classification context, since one summary log is one uploaded workbook, but every rendered column continues to read the row's own value. `coerceForExport` is narrowed to the three fields it actually reads, and `reclassifyWasteRecordStates` now states that it returns one state per input in order — the property that lets the export pair a classification back to its row.

Two behaviours the export documented but did not pin are now covered: the Operator Processing Type column against a summary log holding several types, and a submitted summary log that committed no rows.

[PAE-1765]: https://eaflood.atlassian.net/browse/PAE-1765?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/waste-records-export/application/stream-csv-export.js
+++ b/src/waste-records-export/application/stream-csv-export.js
@@ -6,6 +6,8 @@ import { resolveAccreditation } from '#domain/organisations/registration-utils.j
 import { findSchemaForProcessingType } from '#domain/summary-logs/table-schemas/index.js'
 import { coerceRowData } from '#domain/summary-logs/table-schemas/validation-pipeline.js'
 import { latestSubmittedSummaryLogId } from '#waste-balances/application/latest-submitted-summary-log-id.js'
+import { toWasteRecordState } from '#waste-records/application/read-summary-log-row-states.js'
+import { reclassifyWasteRecordStates } from '#waste-records/application/reclassify-waste-record-states.js'
 import {
   buildHeaderRow,
   buildDataRow,
@@ -60,7 +62,7 @@ const sortRowStates = (a, b) => {
  * top-level field on the row state, is merged back onto the data both to select
  * the schema and to fill its metadata column.
  *
- * @param {SummaryLogRowState} rowState
+ * @param {Pick<SummaryLogRowState, 'data' | 'processingType' | 'wasteRecordType'>} rowState
  * @returns {Record<string, any>}
  */
 const coerceForExport = ({ data, processingType, wasteRecordType }) => {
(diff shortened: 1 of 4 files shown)


src/waste-records-export/application/stream-csv-export.js
```diff
@@ -149,7 +154,19 @@ async function* streamRegistrationRows({
 
   const rowStatesSorted = [...rowStates].sort(sortRowStates)
 
-  for (const rowState of rowStatesSorted) {
+  // One summary log is one uploaded workbook, so every row in it reports under
+  // that workbook's template, and the rows record which one.
+  const [{ processingType }] = rowStatesSorted
```
crosslandwa, line 159: This pulls out the `processingType` once and passes it - inconsistent with what is implemented here

https://github.com/DEFRA/epr-backend/pull/1535/changes#diff-0b427377c3880f5a787b0ca8083e152d71e34051908d0f3f593853817b453a48L55-L81

GraemeF: Yep, as discussed I'll follow up to tidy this now I've realised we do need to carry it with the rows

src/waste-records-export/application/stream-csv-export.js
```diff
@@ -158,7 +175,7 @@ async function* streamRegistrationRows({
         data: coerceForExport(rowState),
         wasteRecordType: rowState.wasteRecordType,
         rowId: rowState.rowId,
-        classification: rowState.classification,
+        classification: liveStates[index].classification,
```
crosslandwa: I wonder if putting something like

```
for (const storedRowState of rowStatesSorted) {
  const [rowState] = reclassifyWasteRecordStates(
    toWasteRecordState(storedRowState),
    { processingType, accreditation, overseasSites }
  )
  yield await encodeRow(...

}
```

is worthwhile, to avoid the lookup by `index` (which frightens me a bit given there's sorting and whatnot going on here)

crosslandwa (approved):

PR #1552, opened 2026-07-23
PAE-1619: accreditation status-history endpoint + suspended-state verification

timrutter:
Ticket: [PAE-1619](https://eaflood.atlassian.net/browse/PAE-1619)
## Summary

- Adds a unified admin endpoint `POST /v1/organisations/{orgId}/registrations/{regId}/accreditations/{accId}/status-history` for accreditation status changes, replacing the earlier single-purpose `PATCH .../suspend` endpoint (per review discussion below)
- The payload is a Joi discriminated union on `status` — one arm per supported transition, each declaring the parameters that transition needs. Only `{ "status": "suspended" }` is supported so far; future transitions (approve, cancel, …) add arms rather than endpoints
- The route validates the transition itself via `assertRegAccStatusTransitionValid` (422 on an invalid transition; schema failures also map to 422 via the server-level failAction)
- The route enforces the registration→accreditation link from the path: 404 when the registration doesn't exist or isn't linked to the accreditation (per review), and the organisation is read once rather than twice
- The status change uses the existing mechanism for updating statuses; future changes will improve this but for now this will do

## Changes

- `src/routes/v1/organisations/accreditation-status-history.js` — the route (renamed from `accreditation-suspend.js`), now takes the target status from the validated payload, checks the registration/accreditation link, and does a single `findById` read
- `src/routes/v1/organisations/accreditation-status-history.schema.js` — the discriminated-union payload schema
- `src/routes/v1/organisations/accreditation-status-history.test.js` — reworked route tests plus payload-validation and mismatched-registration cases
- `src/routes/v1/organisations/index.js` — route export renamed

[PAE-1619]: https://eaflood.atlassian.net/browse/PAE-1619?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/routes/v1/organisations/accreditation-suspend.js
```diff
@@ -0,0 +1,77 @@
...
+import { auditOrganisationUpdate } from '#root/auditing/organisations.js'
+import { assertRegAccStatusTransitionValid } from '#domain/organisations/status.js'
+import { REG_ACC_STATUS } from '#domain/organisations/model.js'
+
+/** @typedef {import('#repositories/organisations/port.js').OrganisationsRepository} OrganisationsRepository */
+/** @typedef {import('#repositories/organisations/port.js').OrganisationReplacement} OrganisationReplacement */
+/** @typedef {import('#repositories/system-logs/port.js').SystemLogsRepository} SystemLogsRepository */
+
+export const accreditationSuspendPath =
+  '/v1/organisations/{organisationId}/registrations/{registrationId}/accreditations/{accreditationId}/suspend'
```
GraemeF: I know you mentioned this earlier, but did you consider a `POST` to `.../status`? It feels like we're adding a new suspension to a bunch of status resources.

timrutter: the problem with that is different status chagnes require different parameters.  So will and I decided to make it one endpoint per status change.  This one happens to have no parameters so it does look a bit overkill.  It will eventually have an applies from date however.  and maybe an "reason" text field, if we go down that route.

timrutter: eg approval for reg or acc requires the reg or acc id.

GraemeF: They can have different shapes, that's no problem. Use a discriminated union in Joi (like we do for the SQS messages, I think).

```
POST .../statusHistory

{
  "status": "approved",
  "accId": "123"
}
```

or whatever

GraemeF: But wouldn't those be taken from the path anyway? I'm confused

timrutter: ok i think you're right, i wasnt massively keen on the many endpoints anyway.

timrutter: also don't mix up acc id and acc number... the number has to be passed in, the id comes from the path. on approval

src/routes/v1/organisations/accreditation-status-history.js
```diff
@@ -0,0 +1,76 @@
...
+   *    organisationsRepository: OrganisationsRepository,
+   *    systemLogsRepository: SystemLogsRepository,
+   *    params: { organisationId: string, registrationId: string, accreditationId: string }
+   * }} request
+   * @param {import('@hapi/hapi').ResponseToolkit} h
+   * @returns {Promise<import('@hapi/hapi').ResponseObject>}
+   */
+  handler: async (request, h) => {
+    const { organisationsRepository } = request
+    const { organisationId, accreditationId } = request.params
```
loveboat: `registrationId` is in the path but never used or validated — the accreditation is suspended purely by org + acc id. A garbage or wrong `registrationId` (non-existent, or linking to a different accreditation) still returns 200, and since path params aren't Joi-validated it isn't even shape-checked.

Either enforce the link:
```js
const registration = initial.registrations.find((r) => r.id === registrationId)
if (registration?.accreditationId !== accreditationId) {
  throw Boom.notFound('Accreditation not found for registration')
}
```
or drop `registrationId` from the path. Worth a mismatched-`registrationId` test either way — every current test uses the correct one, so this path is uncovered.

src/routes/v1/organisations/accreditation-status-history.js
```diff
@@ -0,0 +1,76 @@
...
+    const { status } = request.payload
+
+    const [accreditation, initial] = await Promise.all([
```
loveboat: `findAccreditationById` already calls `findById` internally (mongodb.js:406), so this reads the same org doc twice. Can collapse to one read:
```js
const initial = await organisationsRepository.findById(organisationId)
const accreditation = initial.accreditations.find((a) => a.id === accreditationId)
if (!accreditation) {
  throw Boom.notFound(`Accreditation with id ${accreditationId} not found`)
}
```
Also removes the window where the two reads could observe different versions before the transition check.

src/routes/v1/organisations/accreditation-status-history.schema.js
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * Discriminated union on `status`: each target status declares its own arm
+ * with the parameters that transition requires. Only suspension is supported
+ * so far; add an arm per status as transitions gain endpoints.
+ */
+const suspendedSchema = Joi.object({
+  status: Joi.string().valid(REG_ACC_STATUS.SUSPENDED).required()
+})
+
+export const accreditationStatusHistoryPayloadSchema = Joi.alternatives()
```
loveboat, line 13: Single-arm `alternatives()` is behaviourally identical to exporting `suspendedSchema` directly. Per the YAGNI convention (inline first, extract at the 2nd use case), this could be the plain object schema for now and become an `alternatives` union when the second transition arm actually lands. Comment captures the intent nicely — just flagging the house rule; non-blocking.

loveboat (commented):
A few notes on the new status-history route (non-blocking except perhaps the first). Detail inline.

src/routes/v1/organisations/accreditation-status-history.schema.js
timrutter: next addition to this will come very soon...so will leave

src/routes/v1/organisations/accreditation-status-history.js
timrutter: Fixed in d0ae4e8f — the handler now returns 404 when the registration doesn't exist on the organisation or isn't linked to the accreditation, with tests for both cases. Kept registrationId in the path for consistency with the sibling routes (PRN status, waste-balance-events) which use the same shape.

src/routes/v1/organisations/accreditation-status-history.js
timrutter: Fixed in d0ae4e8f — collapsed to a single findById read; the accreditation (and now registration) lookups run against that one snapshot.

loveboat (approved):

PR #1555, opened 2026-07-24
PAE-1621: reinstate accreditation via status-history endpoint

timrutter:
Ticket: [PAE-1621](https://eaflood.atlassian.net/browse/PAE-1621)
## Summary

- Extends the `POST .../accreditations/{accId}/status-history` discriminated union with an `suspendedToApproved` option, enabling accreditation reinstatement (ACC5: `suspended → approved`)

[PAE-1621]: https://eaflood.atlassian.net/browse/PAE-1621?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/routes/v1/organisations/accreditation-status-history.schema.js
```diff
@@ -2,14 +2,21 @@ import Joi from 'joi'
...
   status: Joi.string().valid(REG_ACC_STATUS.SUSPENDED).required()
 })
 
+const suspendedToApprovedSchema = Joi.object({
+  status: Joi.string().valid(REG_ACC_STATUS.APPROVED).required()
+})
+
 export const accreditationStatusHistoryPayloadSchema = Joi.alternatives()
-  .try(suspendedSchema)
+  .try(approvedToSuspendedSchema, suspendedToApprovedSchema)
```
loveboat, line 21: The arms are named by transition (`approvedToSuspendedSchema` / `suspendedToApprovedSchema`) but only discriminate on the target `status` — the from-state is enforced in the handler, not here. That works while each transition has a distinct target. Once two transitions share a target status — `created -> approved` (grant) vs `suspended -> approved` (reinstate), which the doc comment anticipates — `Joi.alternatives().try()` can't tell them apart on `status` alone and the arms will shadow each other, so it'll need a real discriminator (e.g. presence of `accreditationNumber`). Non-blocking; flagging while the union is still target-status-keyed.

loveboat (commented):
One note on the union's discriminator, inline.

src/routes/v1/organisations/accreditation-status-history.schema.js
timrutter: yeah i did ponder this might be better to pass the to and from status for clarity but that gives a concrete reason to do so.  will sort that in the pr after this one (created -> approved)

loveboat (approved):

src/routes/v1/organisations/accreditation-status-history.schema.js
timrutter: i will resolve this in the next PR that will introduce the condition you are describing

PR #1556, opened 2026-07-24
PAE-1617: grant accreditation via status-history with explicit from/to

timrutter:
Ticket: [PAE-1617](https://eaflood.atlassian.net/browse/PAE-1617)
## Summary

- Extends the status-history endpoint with the `created → approved` (grant) transition. The payload union is now discriminated on an explicit `{ fromStatus, toStatus }` pair — each supported transition is its own arm, and the handler returns 422 when the accreditation is not actually in `fromStatus`
- The grant arm takes `appliesFrom` (YYYY-MM-DD) and `accreditationNumber` (non-empty). The number is issued at grant and must be unique across all organisations' accreditations — new `findByAccreditationNumber` port method (MongoDB + in-memory + contract tests); 422 when already in use. Uniqueness is app-enforced only (no unique index) — documented in the handler
- `validFrom` is set to `appliesFrom`; **`validTo` is not changed** — approval 422s via the existing schema guard if the accreditation data carries no `validTo`, and 422s when `appliesFrom` is after the existing `validTo` so an inverted validity window cannot be written
- Date strings are validated by a shared `isoDateString()` Joi schema with a round-trip check, so impossible calendar dates (e.g. `2026-02-30`, which `Date` silently rolls to March 2) are rejected. Used for both the `appliesFrom` payload field and the stored `validFrom`/`validTo` schema, replacing two duplicated validators

[PAE-1617]: https://eaflood.atlassian.net/browse/PAE-1617?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/routes/v1/organisations/accreditation-status-history.js
+++ b/src/routes/v1/organisations/accreditation-status-history.js
@@ -27,7 +27,12 @@ export const accreditationStatusHistory = {
   },
 
   /**
-   * @param {import('#common/hapi-types.js').HapiRequest<{ status: RegAccStatus }> & {
+   * @param {import('#common/hapi-types.js').HapiRequest<{
+   *    fromStatus: RegAccStatus,
+   *    toStatus: RegAccStatus,
+   *    appliesFrom?: string,
+   *    accreditationNumber?: string
+   * }> & {
    *    organisationsRepository: OrganisationsRepository,
    *    systemLogsRepository: SystemLogsRepository,
    *    params: { organisationId: string, registrationId: string, accreditationId: string }
@@ -38,7 +43,8 @@ export const accreditationStatusHistory = {
   handler: async (request, h) => {
     const { organisationsRepository } = request
     const { organisationId, registrationId, accreditationId } = request.params
-    const { status } = request.payload
+    const { fromStatus, toStatus, appliesFrom, accreditationNumber } =
+      request.payload
 
     const initial = await organisationsRepository.findById(organisationId)
 
@@ -58,11 +64,35 @@ export const accreditationStatusHistory = {
       )
     }
 
+    if (accreditation.status !== fromStatus) {
+      throw Boom.badData(
+        `Cannot transition accreditation from ${fromStatus}: its status is ${accreditation.status}`
+      )
+    }
+
     // Route-level check is required, not belt-and-braces: the repository's
     // assertAndHandleItemStateTransition skips validation when the status is
     // unchanged, so suspended -> suspended via replace() would otherwise be a
     // silent no-op instead of the required 422.
-    assertRegAccStatusTransitionValid(accreditation.status, status)
+    assertRegAccStatusTransitionValid(fromStatus, toStatus)
+
+    // Granting issues the accreditation number, which must not already be in
+    // use by any accreditation in any organisation.
+    if (accreditationNumber) {
+      const holder =
+        await organisationsRepository.findByAccreditationNumber(
+          accreditationNumber
+        )
+      if (holder) {
+        throw Boom.badData(
+          `Accreditation number ${accreditationNumber} is already in use`
+        )
+      }
+    }
+
+    const grantFields = accreditationNumber
+      ? { accreditationNumber, validFrom: appliesFrom }
+      : {}
 
     const { id, version, ...orgFields } = initial
 
@@ -73,7 +103,11 @@ export const accreditationStatusHistory = {
         acc.id === accreditationId
           ? // The transition assert above guarantees the requested status is
             // reachable from the current one, so the result is a valid item.
-            /** @type {typeof acc} */ ({ ...acc, status })
+            /** @type {typeof acc} */ ({
+              ...acc,
+              status: toStatus,
+              ...grantFields
+            })
           : acc
       )
     }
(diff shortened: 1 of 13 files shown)


loveboat (commented):
Three notes on the grant path, inline. #1 is the one worth acting on.

src/routes/v1/organisations/accreditation-status-history.schema.js
```diff
@@ -1,22 +1,48 @@
 import Joi from 'joi'
 import { REG_ACC_STATUS } from '#domain/organisations/model.js'
 
+const dateSchema = Joi.string()
+  .pattern(/^\d{4}-\d{2}-\d{2}$/)
+  .custom((value, helpers) => {
```
loveboat: This validator accepts rollover dates: `new Date('2026-02-30T00:00:00.000Z')` isn't NaN (it rolls to Mar 2), so `2026-02-30` / `2026-04-31` pass and get stored verbatim as `validFrom`, then silently shift when parsed downstream. It also duplicates `dateRequiredWhenApprovedOrSuspended()` in `schema/helpers.js:94`, which has the identical gap — so the stored-data schema doesn't catch it either. The frontend's `toIsoDate` already guards this via a round-trip; the endpoint (the authority) doesn't.

Suggest adding a round-trip check (`date.toISOString().slice(0,10) === value`) to the shared helper, extracting its core as e.g. `isoDateString()`, and reusing it here — fixes the new `appliesFrom` and the pre-existing `validFrom`/`validTo` gap in one place, and drops the duplication.

src/routes/v1/organisations/accreditation-status-history.js
```diff
@@ -58,11 +64,35 @@ export const accreditationStatusHistory = {
...
+          accreditationNumber
+        )
+      if (holder) {
+        throw Boom.badData(
+          `Accreditation number ${accreditationNumber} is already in use`
+        )
+      }
+    }
+
+    const grantFields = accreditationNumber
```
loveboat: Nothing enforces `validFrom <= validTo`. Grant sets `validFrom = appliesFrom` freely and leaves `validTo` untouched, so an `appliesFrom` after the existing `validTo` (e.g. `validTo: 2026-12-31`, `appliesFrom: 2027-06-01`) produces an inverted validity window with no rejection — the accreditation schema only checks presence + format per field. Intended, or should grant 422 when `appliesFrom > validTo`?

src/routes/v1/organisations/accreditation-status-history.js
```diff
@@ -58,11 +64,35 @@ export const accreditationStatusHistory = {
...
     // Route-level check is required, not belt-and-braces: the repository's
     // assertAndHandleItemStateTransition skips validation when the status is
     // unchanged, so suspended -> suspended via replace() would otherwise be a
     // silent no-op instead of the required 422.
-    assertRegAccStatusTransitionValid(accreditation.status, status)
+    assertRegAccStatusTransitionValid(fromStatus, toStatus)
+
+    // Granting issues the accreditation number, which must not already be in
+    // use by any accreditation in any organisation.
+    if (accreditationNumber) {
```
loveboat: This is check-then-act with no DB backstop — there's no unique index on `accreditations.accreditationNumber` (and one is awkward: multikey over `accreditations[]` with many nulls needs partial+sparse), so two concurrent grants of the same number can both pass here and both write. Low concurrency since it's admin-issued, but the "globally unique" invariant isn't DB-guaranteed — worth a comment noting that. (The match-any-status behaviour, permanently reserving a number even after cancellation, reads as intentional per the port doc — just confirming.)

src/routes/v1/organisations/accreditation-status-history.schema.js
timrutter: Verified — in Node `2026-02-30T00:00:00.000Z` rolls to March 2 (months out of range do NaN, days roll over), so both branches of the check are real. Extracted `isoDateString()` with the round-trip check into `src/common/validation/iso-date-schema.js` and reused it here and in `dateRequiredWhenApprovedOrSuspended()`, dropping both duplicated validators. Fixed in 81846e60.

src/routes/v1/organisations/accreditation-status-history.js
timrutter: Not intended — grant now returns 422 when `appliesFrom` is after the existing `validTo` (which it still does not change). When `validTo` is absent the replace continues to 422 via the schema guard requiring it on approved accreditations. Fixed in 81846e60.

src/routes/v1/organisations/accreditation-status-history.js
timrutter: Match-any-status is intentional — a number stays reserved even after cancellation, per the port doc. Added a comment at the check documenting that uniqueness is app-enforced only, with no unique index backstop (81846e60). Went comment-only rather than adding the index: a partial multikey index over `accreditations[]` is awkward as you say, and would not catch a duplicate within a single document's array anyway; this is an admin-issued, low-volume path.

loveboat (approved):

PR #1561, opened 2026-07-27
PAE-1705: remove suspended status from registrations

timrutter:
## Summary

- Registrations can no longer be suspended (PAE-1705): the shared registration/accreditation status-transition table is split into `VALID_REG_TRANSITIONS`  and `VALID_ACC_TRANSITIONS` 
- Cancelling a registration now **force-cancels** its linked accreditation when that accreditation is approved or suspended; a linked accreditation that was never live (created/rejected) is left untouched
- Reports & public register: split the status sets so a registration's own status can never be included as suspended. `REPORTABLE_STATUSES` drops `suspended`, and the public register transformer now has separate registration and accreditation inclusion sets so cancelled/suspended linked accreditations stay published (BR-E8) while a registration is never included as suspended

## Production data check

Verified in prod: zero registrations have `suspended` as a current status and zero registrations carry a `suspended` entry in `statusHistory`, so removing `suspended` from the registration schema cannot break reads or replaces of existing documents.

[PAE-1705]: https://eaflood.atlassian.net/browse/PAE-1705?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/application/public-register/public-register-transformer.js
+++ b/src/application/public-register/public-register-transformer.js
@@ -20,7 +20,15 @@ import chunk from 'lodash.chunk'
 import { TEST_ORGANISATION_IDS } from '#common/helpers/parse-test-organisations.js'
 import { config } from '#root/config.js'
 
-const INCLUDED_STATUSES = new Set([
+// A registration's own status can never be suspended (suspension is an
+// accreditation-only concept), so suspended is excluded here.
+const REGISTRATION_INCLUDED_STATUSES = new Set([
+  REG_ACC_STATUS.APPROVED,
+  REG_ACC_STATUS.CANCELLED
+])
+
+// Cancelled and suspended accreditations remain on the public register (BR-E8).
+const ACCREDITATION_INCLUDED_STATUSES = new Set([
   REG_ACC_STATUS.APPROVED,
   REG_ACC_STATUS.SUSPENDED,
   REG_ACC_STATUS.CANCELLED
@@ -92,13 +100,18 @@ function getLinkedAccreditation(registration, accreditations) {
   return registration.accreditationId
     ? accreditations.find(
         (acc) =>
-          acc.id === registration.accreditationId && isInPublishableState(acc)
+          acc.id === registration.accreditationId &&
+          isAccreditationInPublishableState(acc)
       )
     : null
 }
 
-function isInPublishableState(item) {
-  return INCLUDED_STATUSES.has(item.status)
+function isRegistrationInPublishableState(item) {
+  return REGISTRATION_INCLUDED_STATUSES.has(item.status)
+}
+
+function isAccreditationInPublishableState(item) {
+  return ACCREDITATION_INCLUDED_STATUSES.has(item.status)
 }
 
 function isTestOrg(org) {
@@ -115,22 +128,24 @@ function processBatch(batch, reportComplianceData) {
   return batch
     .filter((org) => !isTestOrg(org))
     .flatMap((org) =>
-      org.registrations.filter(isInPublishableState).map((registration) => {
-        const accreditation = getLinkedAccreditation(
-          registration,
-          org.accreditations
-        )
-        const entry = entries.get(registration.id)
-        const reportComplianceFields = entry
-          ? buildReportComplianceFields(periods, entry)
-          : {}
-        return transformRegAcc(
-          org,
-          registration,
-          accreditation,
-          reportComplianceFields
-        )
-      })
+      org.registrations
+        .filter(isRegistrationInPublishableState)
+        .map((registration) => {
+          const accreditation = getLinkedAccreditation(
+            registration,
+            org.accreditations
+          )
+          const entry = entries.get(registration.id)
+          const reportComplianceFields = entry
+            ? buildReportComplianceFields(periods, entry)
+            : {}
+          return transformRegAcc(
+            org,
+            registration,
+            accreditation,
+            reportComplianceFields
+          )
+        })
     )
 }
 
--- a/src/application/public-register/public-register-transformer.test.js
+++ b/src/application/public-register/public-register-transformer.test.js
@@ -215,17 +215,6 @@ describe('transform', () => {
     }
   )
 
-  it('should exclude suspended registrations', async () => {
-    const registration = createTestRegistration({
-      status: REG_ACC_STATUS.SUSPENDED
-    })
-    const org = createTestOrganisation({ registrations: [registration] })
-
-    const rows = await transform([org], { periods: [], entries: new Map() })
-
-    expect(rows).toHaveLength(0)
-  })
-
   it.each([
     {
       status: REG_ACC_STATUS.APPROVED,
--- a/src/application/public-register/public-register-transformer.test.js
+++ b/src/application/public-register/public-register-transformer.test.js
@@ -181,10 +181,6 @@ describe('transform', () => {
       expectedStatus: 'Approved'
     },
     {
-      status: REG_ACC_STATUS.SUSPENDED,
-      expectedStatus: 'Suspended'
-    },
-    {
       status: REG_ACC_STATUS.CANCELLED,
       expectedStatus: 'Cancelled'
     }
@@ -218,6 +214,17 @@ describe('transform', () => {
       ])
     }
   )
+
+  it('should exclude suspended registrations', async () => {
+    const registration = createTestRegistration({
+      status: REG_ACC_STATUS.SUSPENDED
+    })
+    const org = createTestOrganisation({ registrations: [registration] })
+
+    const rows = await transform([org], { periods: [], entries: new Map() })
+
+    expect(rows).toHaveLength(0)
+  })
 
   it.each([
     {
--- a/src/domain/organisations/registration.js
+++ b/src/domain/organisations/registration.js
@@ -52,7 +52,7 @@
 /**
  * @typedef {RegistrationBase & {
  *  registrationNumber: string;
- *  status: Extract<RegAccStatus, 'approved'|'suspended'>;
+ *  status: Extract<RegAccStatus, 'approved'>;
  *  validFrom: string;
  *  validTo: string;
  * }} RegistrationApproved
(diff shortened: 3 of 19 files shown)


src/application/public-register/public-register-transformer.js
```diff
@@ -20,7 +20,15 @@ import chunk from 'lodash.chunk'
 import { TEST_ORGANISATION_IDS } from '#common/helpers/parse-test-organisations.js'
 import { config } from '#root/config.js'
 
-const INCLUDED_STATUSES = new Set([
+// A registration's own status can never be suspended (suspension is an
+// accreditation-only concept), so suspended is excluded here.
+const REGISTRATION_INCLUDED_STATUSES = new Set([
+  REG_ACC_STATUS.APPROVED,
```
crosslandwa, line 26: How big is the blast radius of splitting `REG_ACC_STATUS` into `REGISTRATION_STATUS` and `ACCREDITATION_STATUS`?

timrutter: Measured it: ~380 references to `REG_ACC_STATUS` across 37 files — 78 in 16 production files, the rest in tests/contract tests. All the string values are shared (`created/approved/cancelled/rejected`, plus accreditation-only `suspended`), so there's no data or API impact — it would be a pure source-level rename.

The mechanical 80% is find-and-replace. The awkward 20% is code that's genuinely generic over both entities and would need either a union type or duplicating:
- `schema/base.js` + `schema/helpers.js` — shared status/statusHistory Joi schemas and the `statusWhen` helpers
- the `RegistrationOrAccreditation` domain type and generic functions like `getLastStatusUpdateDate`
- the status-transition plumbing

Worth noting this PR already enforces the distinction at the points that matter: the registration Joi schema no longer accepts `suspended`, the transition tables are separate (`VALID_REG_TRANSITIONS` / `VALID_ACC_TRANSITIONS`), and the typedefs narrow with `Extract<RegAccStatus, ...>`. So the split would buy naming clarity (and a narrower `Registration['status']` type) rather than new runtime safety. I'd do it as a follow-up rather than grow this PR.

loveboat: I was going to ask the same as if they now differ we can use the type system to prevent us accidentally setting a `suspended` status for a registration

src/repositories/organisations/schema/status-transition.js
```diff
@@ -66,36 +65,59 @@ export const assertOrgStatusTransition = (existing, updated) => {
...
 ) => {
-  const statusUpdates = new Map()
+  /** @type {Set<string>} */
+  const cascadeCancelledIds = new Set()
 
   for (const reg of registrations) {
-    if (reg.accreditationId && isRegistrationCancelledOrSuspended(reg)) {
-      statusUpdates.set(reg.accreditationId, reg.status)
+    if (reg.accreditationId && isRegistrationCancelled(reg)) {
+      cascadeCancelledIds.add(reg.accreditationId)
```
loveboat: The cascade force-cancels the linked accreditation regardless of its current status, and `cascadeCancelledIds` then exempts it from the transition table entirely. The doc comment above says "whether that accreditation is approved or suspended" - the code is wider than that.

Confirmed by running the scenario on both branches: registration approved and linked, its accreditation left at `created`, then cancel the registration.

- this branch: succeeds silently, accreditation ends `cancelled` with `statusHistory` `[created, cancelled]`
- main, same payload: `422 Cannot transition registration/accreditation status from approved to cancelled`
- main via the suspend route (the only path to cancelled there): `422 ... from created to suspended`

Main blocked the cascade onto a non-approved accreditation both ways. The exemption removes that guard rather than replacing it.

Not an edge case: `validateAccreditationLinkExists` / `LinkMatches` / `LinkUniqueness` do not require the linked accreditation to be approved, and `findEligibleAccreditations` (`link-form-submissions.js:156`) deliberately links non-approved accreditations, filtering approved ones out. Registration approved + accreditation still `created` is the normal state while an accreditation application is pending. The same path also resurrects a `rejected` accreditation into `cancelled`, overwriting the rejection in its history.

Gating the cascade on the accreditation's current status would match the stated intent:

```js
const CASCADEABLE = new Set([REG_ACC_STATUS.APPROVED, REG_ACC_STATUS.SUSPENDED])
// only add to cascadeCancelledIds when CASCADEABLE.has(acc.status)
```

A test for the `created`-accreditation case would pin this down - there is currently no direct unit test for this function's new return shape.

src/repositories/organisations/helpers.js
```diff
@@ -87,7 +111,8 @@ export const mapDocumentWithCurrentStatuses = (org) => {
   rest.status = /** @type {OrganisationStatus} */ (getCurrentStatus(rest))
 
   for (const item of rest.registrations) {
-    item.status = getCurrentStatus(item)
+    // Greenfield: no stored registration history contains suspended (PAE-1705)
```
loveboat, line 114: This comment is the load-bearing assumption for the whole change, and two things depend on it.

1. A document whose derived registration status is `suspended` reads back fine, silently drops out of the public register and reports, and then 422s on every subsequent `replace()` - the schema no longer accepts `suspended`, and the payload echoes the registration's own status back. That blocks updates to the entire organisation, not just that registration.

2. `statusHistoryItemSchema` (`schema/base.js:67`) still lists `SUSPENDED` and is shared by both entities, so registration history containing `suspended` remains valid backend-side. The admin PR (DEFRA/epr-re-ex-admin-frontend#504) removes `suspended` from the registration `statusHistory` enum, and that validator drives the admin raw-JSON editor. Any registration ever suspended and then reinstated or cancelled carries a permanent `suspended` history entry the backend accepts and the admin editor will reject, so the admin cannot save edits to that organisation. The two repos disagree on what a valid registration is.

Suggest querying each env for `registrations.statusHistory.status: 'suspended'` and putting the count in the PR body rather than asserting it in a comment. If history entries do exist, either narrow the backend statusHistory schema per entity to match the admin FE, or drop the history enum change from the admin PR.

The `/** @type {Registration['status']} */` cast on the next line also launders the assumption past tsc.

src/domain/organisations/registration.js
```diff
@@ -52,7 +52,7 @@
 /**
  * @typedef {RegistrationBase & {
  *  registrationNumber: string;
- *  status: Extract<RegAccStatus, 'approved'|'suspended'>;
+ *  status: Extract<RegAccStatus, 'approved'>;
```
loveboat, line 55: `getReportableRegistrations` returns `Array<{ registration: RegistrationApproved }>` but `REPORTABLE_STATUSES` is now `{approved, cancelled}`, so narrowing `status` here makes that return type wrong for the cancelled half. Pre-existing mismatch, tighter after this change.

Related: `registrationNumber`, `validFrom` and `validTo` are typed non-null on this typedef, but the schema allows them null for cancelled registrations (`requiredWhenApproved`). Nothing dereferences them off the registration in `report-submissions.js` or `report-compliance.js` today, so no live bug - but a `ReportableRegistration` alias, or widening the return annotation, would stop that drifting.

Minor: `Extract<RegAccStatus, 'approved'>` with a single member is just `'approved'`.

loveboat (commented):
Three findings. The first is confirmed by running the scenario on both branches.

src/application/public-register/public-register-transformer.test.js
```diff
@@ -219,6 +215,17 @@ describe('transform', () => {
     }
   )
 
+  it('should exclude suspended registrations', async () => {
```
crosslandwa: This looks superfluous - a registration can't be suspended...

src/application/public-register/public-register-transformer.js
timrutter: i'll do it in a separate PR.

src/repositories/organisations/helpers.js
timrutter: there are zero suspended registration status history entries, i have checked this in prod.  I think this is what that is getting at but its a bit AI speak

src/application/public-register/public-register-transformer.test.js
crosslandwa: I'm going to assume this test goes away in the _other_ PR (where `REG_ACC_STATUS` splits into REGISTRATION_STATUS and ACCREDITATION_STATUS) - the type system would catch `createTestRegistration({ status: REG_ACC_STATUS.SUSPENDED })` as an error (without the need for a runtime test)

crosslandwa (approved):

src/repositories/organisations/schema/status-transition.js
timrutter: Fixed in b21599fe. The cascade now only fires when the accreditation's payload status is approved or suspended — created/rejected accreditations are left untouched and excluded from `cascadeCancelledIds` (so a genuine user-driven transition on them still hits the transition table). Checking the payload status also keeps the reinstatement-held-at-cancelled behaviour intact. Added a unit test for the function's return shape and a contract test for the created-accreditation repro.

src/repositories/organisations/helpers.js
timrutter: Prod check is now recorded in the PR body: zero registrations with a suspended current status and zero suspended registration statusHistory entries. On the backend/admin statusHistory enum divergence: with zero existing data both schemas are satisfiable today, and the backend can no longer produce a suspended registration entry; narrowing the backend statusHistory enum per entity will land with the REGISTRATION_STATUS/ACCREDITATION_STATUS split follow-up PR.

src/domain/organisations/registration.js
timrutter: Fixed in b21599fe. Added a `ReportableRegistration` typedef (status `'approved' | 'cancelled'`, keeping registrationNumber/validFrom/validTo non-null — a cancelled registration was previously approved, so it carries them) and used it for `getReportableRegistrations` and `buildRow`. Left `Extract<RegAccStatus, 'approved'>` as-is in the registration typedefs: it stays anchored to the enum, so the planned enum-split PR will surface those sites.

src/application/public-register/public-register-transformer.test.js
timrutter: Removed in b21599fe — a registration can no longer be suspended, and exclusion of non-included statuses is already covered by 'should filter out registrations with unapproved statuses'.

PR #1564, opened 2026-07-27
PAE-1705: split REG_ACC_STATUS into per-entity status enums

timrutter:

Follow-up to #1561, as discussed in [this review thread](https://github.com/DEFRA/epr-backend/pull/1561#discussion_r3658021030).

## Summary

- Split the shared `REG_ACC_STATUS` enum into `REGISTRATION_STATUS` (created/approved/cancelled/rejected) and `ACCREDITATION_STATUS` (those plus suspended), so the type system now prevents a registration from ever being assigned `suspended` — previously only runtime validation caught this
- Corresponding JSDoc types `RegistrationStatus` / `AccreditationStatus`, plus a `RegOrAccStatus` union for the few places genuinely generic over both (status history, user collation)
- Narrowed the `statusHistory` item schemas per entity for the admin raw-JSON editor (`organisation-json-schema-overrides`): registration history no longer accepts `suspended`, aligning the backend with the admin frontend's registration `statusHistory` enum
- Pure rename at runtime — every status string value is unchanged, so there is no data or API impact. Where generic code compares against values shared by both enums (e.g. `approved`), the registration constants are used

[PAE-1705]: https://eaflood.atlassian.net/browse/PAE-1705


crosslandwa (approved):

src/domain/organisations/registration-utils.js
```diff
@@ -1,18 +1,21 @@
...
-/** @import { GlassRecyclingProcess, Material, Organisation, RegAccStatus } from '#domain/organisations/model.js' */
+/** @import { AccreditationStatus, GlassRecyclingProcess, Material, Organisation, RegistrationStatus } from '#domain/organisations/model.js' */
 /** @import { Registration, ReportableRegistration } from '#domain/organisations/registration.js' */
 /** @import { Accreditation } from '#domain/organisations/accreditation.js' */
 
 const TEST_ORGANISATIONS = new Set(TEST_ORGANISATION_IDS)
 
-const REPORTABLE_STATUSES = /** @type {Set<RegAccStatus>} */ (
-  new Set([REG_ACC_STATUS.APPROVED, REG_ACC_STATUS.CANCELLED])
+const REPORTABLE_STATUSES = /** @type {Set<RegistrationStatus>} */ (
```
loveboat, line 13: `/** @type {X} */ (expr)` is a type *assertion*, not an annotation. `Set<string>` and `Set<RegistrationStatus>` are mutually assignable, so tsc accepts whatever is inside the parentheses.

Checked on this branch - added `ACCREDITATION_STATUS.SUSPENDED` to this set, left the cast as-is, ran `sk_types.sh`: **zero errors**, prod and tests.

Moving the same type off the cast and onto an annotation catches it:

```js
/** @type {Set<RegistrationStatus>} */
const REPORTABLE_STATUSES = new Set([
  REGISTRATION_STATUS.APPROVED,
  REGISTRATION_STATUS.CANCELLED
])
```

With that form, the same injected `SUSPENDED` gives:

```
error TS2322: Type 'Set<"approved" | "cancelled" | "suspended">' is not assignable to type 'Set<RegistrationStatus>'.
```

in both prod and tests. Same pattern applies to `ACTIVE_ACCREDITATION_STATUSES` below and to `CASCADEABLE_ACC_STATUSES` in `schema/status-transition.js`. Three one-line moves and the split starts enforcing the rule instead of describing it.

loveboat (commented):
The split itself is the right shape: per-entity enums, `RegOrAccStatus` kept only for genuinely generic code, `ReportableRegistration` added, per-entity statusHistory schemas wired into the admin JSON-schema overrides. The cascade guard from #1561 is in too.

The part that does not hold is the claim that "the type system now prevents a registration from ever being assigned suspended". I checked that on the branch and it does not - details inline on `registration-utils.js`. Three separate causes, each a small fix, and together they are the difference between the split documenting the rule and enforcing it.

One thing outside the diff: the generated admin schema now differs from what epr-re-ex-admin-frontend#504 committed by hand. Details on `schema/base.js`.

src/application/public-register/public-register-transformer.js
```diff
@@ -23,15 +26,15 @@ import { config } from '#root/config.js'
 // A registration's own status can never be suspended (suspension is an
 // accreditation-only concept), so suspended is excluded here.
 const REGISTRATION_INCLUDED_STATUSES = new Set([
-  REG_ACC_STATUS.APPROVED,
-  REG_ACC_STATUS.CANCELLED
+  REGISTRATION_STATUS.APPROVED,
```
loveboat, line 29: These two sets have no type at all, and `isRegistrationInPublishableState(item)` / `isAccreditationInPublishableState(item)` below have no JSDoc, so `item.status` is `any`.

Checked: adding `ACCREDITATION_STATUS.SUSPENDED` to `REGISTRATION_INCLUDED_STATUSES` typechecks clean. This is the file where the two sets were first split apart, so it is the one most worth pinning down.

```js
/** @type {Set<RegistrationStatus>} */
const REGISTRATION_INCLUDED_STATUSES = new Set([...])

/** @param {Registration} item */
function isRegistrationInPublishableState(item) { ... }
```

Annotating the predicate is what makes the set annotation bite - without it, `.has(any)` is unchecked regardless of the set's type.

src/domain/organisations/status.js
```diff
@@ -21,24 +22,36 @@ const VALID_ORG_TRANSITIONS = {
...
+    ACCREDITATION_STATUS.SUSPENDED,
+    ACCREDITATION_STATUS.CREATED
+  ],
+  [ACCREDITATION_STATUS.SUSPENDED]: [
+    ACCREDITATION_STATUS.APPROVED,
+    ACCREDITATION_STATUS.CANCELLED
   ],
-  [REG_ACC_STATUS.CANCELLED]: [REG_ACC_STATUS.APPROVED],
-  [REG_ACC_STATUS.REJECTED]: [REG_ACC_STATUS.CREATED]
+  [ACCREDITATION_STATUS.CANCELLED]: [ACCREDITATION_STATUS.APPROVED],
```
loveboat, line 53: Two things about the tables and the asserters built from them.

1. `makeAssertStatusTransitionValid` still takes `Record<string, string[]>` and returns `(fromStatus: string, toStatus: string) => void`. Checked: `assertRegistrationStatusTransitionValid('suspended', 'not-a-status')` typechecks clean. Making the factory generic over the entity's status type would make that a compile error, which is the same win as the sets.

2. There is no test tying a table's keys to its enum. With `validTransitions[fromStatus] ?? []` a missing key does not throw - every transition out of that status silently 422s instead of failing loudly. Now that both are derivable from one source:

```js
expect(Object.keys(VALID_REG_TRANSITIONS).toSorted()).toStrictEqual(
  Object.values(REGISTRATION_STATUS).toSorted()
)
```

plus the accreditation equivalent. That is the test that stops the two declarations drifting.

src/repositories/organisations/schema/registration.js
```diff
@@ -72,10 +72,10 @@ export const registrationSchema = Joi.object({
   id: idSchema,
   status: Joi.string()
     .valid(
-      REG_ACC_STATUS.CREATED,
-      REG_ACC_STATUS.APPROVED,
-      REG_ACC_STATUS.CANCELLED,
-      REG_ACC_STATUS.REJECTED
+      REGISTRATION_STATUS.CREATED,
```
loveboat, line 75: This list is still restated rather than derived, and it is inconsistent with `makeStatusHistoryItemSchema` in `schema/base.js`, which this PR changes to build from `Object.values(...)`.

```js
.valid(...Object.values(REGISTRATION_STATUS))
```

Same for `accreditation.js`. That removes the last independent declaration of "which statuses a registration can hold" - after which the enum, the schema and the transition table all come from one place, which was the point of the split.

src/repositories/organisations/schema/base.js
```diff
@@ -64,20 +65,32 @@ export const linkedDefraOrganisationSchema = Joi.object({
...
-      ORGANISATION_STATUS.ACTIVE,
-      REG_ACC_STATUS.CANCELLED,
-      REG_ACC_STATUS.REJECTED,
-      REG_ACC_STATUS.SUSPENDED
-    )
-    .required(),
-  updatedAt: Joi.date().iso().required(),
-  updatedBy: idSchema.optional()
-})
+const makeStatusHistoryItemSchema = (statuses) =>
```
loveboat, line 68: Narrowing these per entity is the right fix and it supersedes the hand-edit in epr-re-ex-admin-frontend#504 - the admin `organisation.json` is a generated artefact (`package.json` `schema:sync-local` runs `schema:generate` then copies `.schemas/organisation.update.json` across and recompiles ajv).

Worth flagging that the generated output is not identical to what #504 committed. Ran `npm run schema:generate` on this branch and diffed against #504's `organisation.json`:

```
1190a1191 >   "active",
1606a1608 >   "active",
```

The old shared `statusHistoryItemSchema` had `ORGANISATION_STATUS.ACTIVE` bolted on, so `active` leaked into the registration and accreditation statusHistory enums in the generated schema. The new per-entity schemas are pure `Object.values(...)`, so it is gone. That looks correct - `active` is an organisation status - but it is an unannounced change, and it means #504's hand-edited file will not match the next `schema:sync-local`. Suggest regenerating #504 from this branch rather than keeping the manual edit, and calling the `active` removal out in the description.

Still open from the #1561 thread: if any existing registration statusHistory contains `suspended`, this makes it invalid on the backend side too, so that organisation becomes unsaveable in both layers rather than one. The per-env count for `registrations.statusHistory.status: 'suspended'` is still the gating fact.

PR #1567, opened 2026-07-28
PAE-1705: make the split status enums enforceable by the type system

loveboat:

Follow-up to #1561 and #1564. The enums were split, but nothing enforced them — `assertRegistrationStatusTransitionValid('suspended', 'not-a-status')` compiled fine. This makes the split hold.

- casts (`/** @type {Set<X>} */ (new Set(...))`) replaced with annotations — a cast is mutually assignable, so it accepted any member
- public register status sets and their predicates were untyped, so `.has(any)` was unchecked
- transition asserters were generic over `string`; now generic over the entity's status union, with the tables typed `Record<RegistrationStatus, RegistrationStatus[]>` — which also makes enum key coverage a compile-time guarantee
- `getCurrentStatus` was `any` in, asserted union out, on the line every status flows through; now generic, removing three casts at call sites
- swept remaining `status: string` JSDoc in reg/acc code onto the per-entity unions
- deduped the active-accreditation set (approved/suspended), which existed three times: two sets and one inline `||`

Verified by injecting a non-member into each set and the asserter and confirming `lint:types` rejects all four.

Test fixtures that no longer typecheck are fixed test-side by annotating the tables, so literals are checked against the union. The nine `suspended` rows in the registration transition table are unrepresentable now, replaced by two tests for the runtime guard (source and target branches).

396 files, 7075 tests, 100% coverage.

[PAE-1705]: https://eaflood.atlassian.net/browse/PAE-1705?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/application/public-register/public-register-transformer.js
+++ b/src/application/public-register/public-register-transformer.js
@@ -3,6 +3,8 @@
  */
 
 /** @import {AccreditationStatus, Organisation, RegistrationStatus} from '#domain/organisations/model.js' */
+/** @import {Accreditation} from '#domain/organisations/accreditation.js' */
+/** @import {Registration} from '#domain/organisations/registration.js' */
 /** @import {PublicRegisterRow} from './types.js' */
 /** @import {ReportComplianceData} from '#reports/application/report-compliance.js' */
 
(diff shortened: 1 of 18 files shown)


timrutter (approved):
claude happy with this

PR #1570, opened 2026-07-28
PAE-1790: stop shared code importing feature internals

GraemeF:
Under the modular monolith structure (ADR-0027), `src/common/` is our shared layer: it may be imported by any feature, and must not import a feature's internals in return. Three places broke that rule. No behaviour changes.

**Database seeding no longer lives in the shared Mongo plugin.** `mongo-db-plugin.js` was importing the concrete organisations Mongo repository so it could build one and run non-production seed data — shared infrastructure reaching into a feature's adapter. The seeding now lives at the composition root in `src/server/seed/`, and the plugin takes a `seedDatabase` step as an option, wired in `server.js`. `common/helpers/collections/create-update.js` keeps only the form-collection and lock-index setup; the seed fixtures and scenarios moved out with the rest.

**Summary-log metrics moved into the summary-logs application layer.** `common/helpers/metrics/summary-logs.js` typed its dimensions against `#domain/summary-logs`, and every consumer already sat inside the feature. It is now `src/application/summary-logs/metrics.js`.

**Defra ID organisation-scope resolution moved into the organisations application layer.** `getDefraUserRoles` and `getOrgMatchingUsersToken` are organisation-scoped authorisation, not generic auth, and pulled `ORGANISATION_STATUS` into shared code. They now live in `src/application/organisations/`, and `getJwtStrategyConfig` receives the role resolver from the auth plugin rather than importing it. Generic JWT verification and issuer routing stay shared. The injected collaborator is typed, so a resolver of the wrong shape now fails at the wiring point.

Three shared files still reach into feature modules and are deliberately left alone. `hapi-types.js` type-imports three repository ports plus the waste-balance service — an application module rather than a port — to type the decorated request; `helpers/formatters.js` imports the material and tonnage vocabulary that currently sits under `#domain/organisations/`. Fixing either here means a 100+ file rewrite colliding with several branches in flight. `helpers/dates/accreditation.js` pulls `ACCREDITATION_STATUS` from the organisations domain — the same offence as the auth file this PR moves, and much cheaper, but it lands more naturally with the organisations module than with the shared-layer clean-up. All three are tracked separately.

Heads-up for rebasing: this PR renames `src/common/helpers/metrics/summary-logs.test.js` to `src/application/summary-logs/metrics.test.js`, which [#1348](https://github.com/DEFRA/epr-backend/pull/1348) modifies.


src/common/helpers/auth/get-jwt-strategy-config.test.js
```diff
@@ -10,8 +10,12 @@ import {
 import Boom from '@hapi/boom'
 import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest'
 import { SCOPES, ADMIN_ROLES } from './constants.js'
+import { getDefraUserRoles } from '#application/organisations/get-defra-user-roles.js'
```
loveboat: This reintroduces `common` -> `#application` in `common`'s own test. Sweeping the branch, it's the only unit test under `src/common` reaching into a feature -- the only other case is the mongo plugin integration test booting the whole server.

The seam you've added means a `vi.fn()` works here, and it costs no coverage: `src/application/organisations/get-defra-user-roles.test.js` already has 10 cases covering the full scope matrix, including "does not assign standard user scope when user is linked to an active organisation, but request does not specify an organisation" -- which is what `returns standard user scope for valid Defra ID tokens` re-asserts below via `stubRequest()`. Stubbing drops a duplicate assertion and leaves this file testing issuer routing.

src/common/helpers/auth/get-jwt-strategy-config.js
```diff
@@ -1,17 +1,19 @@
...
+
 /**
  * Configures JWT authentication strategy for both Entra ID and Defra ID
- * @param {Object} oidcConfigs - OIDC configuration for both identity providers
- * @param {Object} oidcConfigs.entraIdOidcConfig - Entra ID OIDC configuration
- * @param {Object} oidcConfigs.defraIdOidcConfig - Defra ID OIDC configuration
+ * @param {Object} authConfig - Identity provider configuration and collaborators
+ * @param {Object} authConfig.entraIdOidcConfig - Entra ID OIDC configuration
+ * @param {Object} authConfig.defraIdOidcConfig - Defra ID OIDC configuration
+ * @param {(tokenPayload: DefraIdTokenPayload, request: *) => Promise<UserRoleAndScopes>} authConfig.getDefraUserRoles - Resolves the roles a Defra ID user holds for the requested organisation
```
loveboat: `request: *` is `any`, and this is the one place it matters -- injecting the collaborator is exactly where a wrong shape stops being caught. The first param does real work under `strictFunctionTypes`; this slot doesn't.

`HapiRequest` from `#common/hapi-types.js` is common -> common, so no boundary cost, and it adds no new errors: `validate` below is itself `@param {*} request`, so `any` flows in, and a contextually-typed stub is fine under `noImplicitAny: false`.

The three `{Object}` annotations give no property checking either -- `entraIdOidcConfig.jwks_uri` and `.issuer` compile clean against `Object`. A TS object type-literal for `authConfig` would type the new seam and both OIDC configs in one edit.

src/common/helpers/plugins/mongo-db-plugin.js
```diff
@@ -1,15 +1,12 @@
 import { createMongoClient } from '#common/helpers/mongo-client.js'
-import { createOrganisationsRepository } from '#repositories/organisations/mongodb.js'
```
loveboat, line 2: Worth locking this in -- nothing stops the next one going back the other way. `eslint-plugin-import-x@4.17.1` is already a dependency and ships `no-restricted-paths` (built on `moduleVisitor`, so static, re-export and dynamic imports are all covered).

A zone of `src/common` -> `src/{application,routes,repositories,server,plugins,adapters}` + `src/waste-balances` has **zero violations** in non-test code on this branch -- it could go in with this PR. Tests need a `**/*.integration.test.js` override (the integration test in this directory dynamically imports `#server/server.js`, which is inherent to it).

Caveat: the rule only sees import syntax, so it can't guard the `@import` leaks in `hapi-types.js` or `vi.mock('#application/...')` strings.

src/common/helpers/plugins/mongo-db-plugin.js
```diff
@@ -18,6 +15,12 @@ export const mongoDbPlugin = {
     version: '1.0.0',
     /**
      * @param {import('../../hapi-types.js').HapiServer} server
+     * @param {{
```
loveboat, line 21: Inline `import('mongodb').Db` / `import('../../hapi-types.js').HapiServer` here, while the two new files in this PR (`seed-database.js`, `create-seed-data.js`) use `@import` blocks. Same for the rewritten `@typedef {import('...')}` shims in the moved auth files -- those paths changed anyway, so converting is close to free.

src/server/seed/seed-database.js
```diff
@@ -0,0 +1,19 @@
...
+import { isProductionEnvironment } from '#root/config.js'
+import { createSeedData } from './create-seed-data.js'
+
+/** @import {Db} from 'mongodb' */
+
+/**
+ * @param {Db} db
+ * @returns {Promise<void>}
+ */
+export async function seedDatabase(db) {
```
loveboat, line 11: `createOrganisationsRepository()` calls `ensureCollection()`, which issues 4 `createIndex` calls; `src/repositories/organisations/mongodb.plugin.js:16` calls the same factory. That runs twice at every startup, prod included.

Pre-existing, but this 19-line file at the composition root is where `if (isProductionEnvironment()) return` would drop the redundant pass in prod and put the production guard at the point of wiring rather than inside `createSeedData`'s body.

src/application/organisations/get-defra-user-roles.js
```diff
@@ -3,7 +3,7 @@ import { ORGANISATION_STATUS } from '#domain/organisations/model.js'
 import { getOrgMatchingUsersToken } from './get-users-org-info.js'
```
loveboat, line 3: The stated rationale for moving this file is that it pulled `ORGANISATION_STATUS` into shared code (line 2 above). `src/common/helpers/dates/accreditation.js:2` does the same with `ACCREDITATION_STATUS` from `#domain/organisations/model.js` and isn't in the deferred list -- it looks far cheaper to move than `formatters.js` (one domain dependency, no 100-file blast radius). Either move it or add it to the list.

Also: the description says `hapi-types.js` keeps "three repository ports" -- there are three, plus a fourth reach into `#waste-balances/application/waste-balance-service.js`, which is an application module rather than a port.

src/server/seed/create-seed-data.js
```diff
@@ -0,0 +1,128 @@
...
+import accreditationFixture from '#data/fixtures/accreditation.json' with { type: 'json' }
+import organisationFixture from '#data/fixtures/organisation.json' with { type: 'json' }
+import registrationFixture from '#data/fixtures/registration.json' with { type: 'json' }
+
+import eprOrganisation1 from '#data/fixtures/common/epr-organisations/sample-organisation-1.json' with { type: 'json' }
+import eprOrganisation2 from '#data/fixtures/common/epr-organisations/sample-organisation-2.json' with { type: 'json' }
+import eprOrganisation3 from '#data/fixtures/common/epr-organisations/sample-organisation-3.json' with { type: 'json' }
+import eprOrganisation4 from '#data/fixtures/common/epr-organisations/sample-organisation-4.json' with { type: 'json' }
+
+import { ObjectId } from 'mongodb'
```
loveboat: Nit: third-party import sits below the JSON fixtures. Move it up with the other module imports.

loveboat (commented):
Comments inline. Behaviour-neutral as far as I can tell -- seeding order and the `isProduction` guard are unchanged, `mongoDbPlugin` has one registration site so the new required option can't go missing, and no stale references to any moved path remain.

loveboat (approved):

PR #1571, opened 2026-07-28
PAE-1790: stop the summary-logs domain importing into the application layer

GraemeF:

The summary-logs domain layer imported upwards into the application layer. Three moves put each piece of code in its correct layer. No behaviour changes. Each commit is a file move plus the changes to the import paths.

**The row-transformer factory.** Thirteen table schemas in `src/domain/summary-logs/table-schemas/` imported `createRowTransformer` from `#application/waste-records/row-transformers/`. The factory has no dependency on the application layer. It makes sure that a row ID is present, then it shapes a waste record. Only the directory was wrong. The factory moves to `src/domain/waste-records/row-transformers/`, and the thirteen imports change only their prefix.

**The loads typedefs.** `src/domain/summary-logs/model.js` reached into `#application/summary-logs/load-counts.js` and `period-status.js` for the `Loads` and `LoadsByReportingPeriod` types. Both types describe the summary-log document in storage, and the domain already owns the Joi schemas that validate those shapes. The typedefs move to `loads-schema.js` and `loads-by-period-status-schema.js`, next to those schemas. A shape and its validator are now in one place. The application layer keeps the code that counts and classifies, and imports the types downwards.

**The failure transitions.** `src/domain/summary-logs/mark-as-failed.js` typed a repository port from inside the domain layer. The file does repository work: it reads the log, it checks the status, then it writes the transition back. That work is a use case, so the file moves to `src/application/summary-logs/`.

**One duplicated constant.** The move of the `Loads` typedefs showed a duplicated cap. The schema held a literal `.max(100)`, and `load-counts.js` declared its own `MAX_ROW_IDS = 100` for the truncation that produces those row IDs. The domain schema now exports the constant and the producer imports it. The validator and the code it validates cannot drift apart.

**What the moves leave behind.** `grep -rn "#application" src/domain/` now returns nothing across the full domain tree. Three test files in `src/domain/summary-logs/table-schemas/` still import a contract test-data builder from `#repositories`. Those imports are fixtures, not production layering, so they stay.

Two edges stay in place on purpose. `table-schemas/exporter/received-loads-for-export.js` and `reports/domain/aggregation/aggregate-waste-exported.js` both import `#overseas-sites/domain/approval.js`. That is a reach across modules, not an inversion of the layers. The overseas-sites module-interface work tracks it separately.

**Type checks.** Both type-check configurations report zero errors, the same as `main`. This repository sets `noImplicitAny` to false, so a clean `tsc` is no evidence on its own that a JSDoc type edge resolves. Each of the two moved type edges therefore got a poison test. A rename of the imported typedef makes the build fail, which confirms that the edge is live.


src/domain/summary-logs/loads-schema.js
```diff
@@ -6,6 +6,27 @@ import Joi from 'joi'
  * Used by both repository (storage validation) and route (response validation)
  */
 
+/**
+ * @typedef {Object} LoadCategory
+ * @property {number} count - Total count of loads
+ * @property {string[]} rowIds - Row IDs (truncated to the schema's cap)
```
loveboat: The type and the validator it now sits beside disagree: this says `string[]`, the schema below accepts `Joi.alternatives().try(Joi.string(), Joi.number())`. `loadsSchema` is on both the storage path (`repositories/summary-logs/schema.js:89`) and the route response contract (`routes/.../response.schema.js:69`), so a persisted doc with numeric rowIds would serialise numbers under a contract declaring strings.

Evidence points at tightening the schema rather than widening the type: the `Joi.number()` branch dates to the original commit (0436d5d, PAE-425) rather than being added later to tolerate stored data, the producer path is string-typed throughout (`addToCategory` is `@param {string} rowId`, `createRowTransformer` does `String(rowData[rowIdField])`), and every fixture in `get.test.js` uses strings. Your call which way, but it should be one or the other.

Worth noting nothing asserts these hand-written typedefs agree with the Joi schemas they mirror, which is why this sat unnoticed until the move put them in one file.

src/application/summary-logs/load-counts.js
```diff
@@ -31,13 +18,6 @@ import { RECORD_CHANGE, recordChangeFor } from './record-change.js'
...
  */
 
-/**
- * @typedef {Object} Loads
- * @property {LoadValidity} added - Loads added in this upload
- * @property {LoadValidity} unchanged - Loads unchanged from previous uploads
- * @property {LoadValidity} adjusted - Loads adjusted in this upload
- */
-
 const MAX_ROW_IDS = 100
```
loveboat: This cap is duplicated as a literal `.max(100)` in `#domain/summary-logs/loads-schema.js`. The sibling file added in this PR already solves exactly that:

```js
// Per-bucket cap on listed rows. The producer (period-status.js) truncates to
// this and the schema validates it; sharing one constant keeps them in step.
export const MAX_ROWS_PER_BUCKET = 100
```

-- imported by `period-status.js:3` and its test. Now that `Loads` lives beside `loadsSchema`, exporting `MAX_ROW_IDS` from the domain schema and importing it here is a three-line change that finishes the job.

Related: the typedef comment was reworded to "truncated to the schema's cap", which isn't accurate -- the schema doesn't truncate, `addToCategory` below does, and the schema independently validates the same number. The original wording named the actual mechanism. Sharing the constant would make the new wording true.

loveboat (commented):
Verified against `fe96f4f`. The description understates one result: `src/domain/**` non-test now has zero imports from `#application`, `#routes`, `#repositories`, `#server`, `#plugins` or `#adapters` -- the whole domain layer, not just summary-logs. Old paths are fully removed (`src/application/waste-records/row-transformers/` is gone rather than half-emptied), all 13 table-schema importers are repointed with no stragglers, and the one remaining reference to `#application/summary-logs/load-counts.js` is a runtime `createEmptyLoads` import in an integration test, which is correct since that function stays put.

The destination for the factory adds no new cross-module edge either: the table schemas already import `#domain/waste-records/model.js` for `WASTE_RECORD_TYPE`, so summary-logs -> waste-records at domain level pre-exists. The move converts 13 layering inversions into edges that were already there.

Two comments inline, both on `loads-schema.js` / `load-counts.js`. Neither is introduced by this PR -- co-locating the type with its validator is what makes them visible, which is arguably the point.

Carried over from #1570: with `src/domain/**` now at zero upward imports, an `import-x/no-restricted-paths` zone for `src/domain` -> `src/{application,routes,repositories,server,plugins,adapters}` passes today, needing only an override for the three test files importing `#repositories/organisations/contract/test-data.js`. Same rule instance as the `src/common` zone -- one config block could pin both.

src/application/summary-logs/load-counts.js
GraemeF: Done in 506d558d. `MAX_ROW_IDS` is now exported from `#domain/summary-logs/loads-schema.js`, `loadCategorySchema` uses it in its `.max()`, and `load-counts.js` imports it instead of declaring its own.

On the comment wording — you're right that the schema doesn't truncate, so I've restored the original `truncated to MAX_ROW_IDS`, which is now a live reference in the same file rather than a dangling name.

One thing that fell out of the move and isn't in your comment: `loads-by-period-status-schema.js` carried `// Matches MAX_ROW_IDS in load-counts.js`, which went stale the moment the constant moved. Repointed to `loads-schema.js` in the same commit.

src/domain/summary-logs/loads-schema.js
GraemeF: I dug into this rather than pick a direction, because the evidence for tightening doesn't survive the history — and I'd have made a read-breaking change on it.

**The producer has not always stringified.** `String(rowData[rowIdField])` arrived on 2026-06-19 in 772883f0 (#1349). Before that, `createRowTransformer` returned `rowId: rowData[rowIdField]` raw. That commit is PAE-1576, a Bug, and its description is the point: *"ExcelJS reads numeric `ROW_ID` cells as numbers, so `createRowTransformer` produces records whose `rowId` is a number rather than a string."* Numeric row IDs in real uploads are the premise of the ticket, not a hypothetical.

**So there's roughly a seven-month window of exposure.** `loads` gained `rowIds` on 2025-11-28 in 0436d5dd (PAE-425) — the same commit that added the `Joi.number()` branch you spotted. Anything uploaded between then and 2026-06-19 with a numeric `ROW_ID` cell could have persisted a number here.

Checking the pre-fix tree rather than inferring from today's:

- `addToCategory` takes `record.rowId` raw, and `classify-and-persist.js` has no `String(` anywhere on the path.
- The alternatives branch *preserves* the number rather than casting it — `[1000, 'row-2']` validates clean and comes back `[1000, "row-2"]`.
- The upload didn't abort earlier either: the legacy waste-records schema did have `rowId: Joi.string().required()`, but `repositories/waste-records/mongodb.js` only ever imported `validateOrganisationId` and `validateRegistrationId`, so `validateWasteRecord` was never on the write path. That's PAE-1576's own "currently masked in production".
- No migration or backfill has ever written `loads` — the classification path is the only writer.

**One correction to where the risk sits.** `repositories/summary-logs/validation.js` validates on insert and update only, so repository reads survive a tightened schema. The exposure is the GET route: it sets `response: { schema }` with no `failAction`, so hapi's default applies and a failure is a 500 — one legacy document with a numeric row ID means a dead check page surfacing as a server error. And tightening really is rejection, not coercion: `Joi.string().validate(1000)` errors.

**What I'd suggest instead of either direction: coerce at the boundary.**

```js
rowIds: Joi.array()
  .items(Joi.alternatives().try(Joi.string(), Joi.number().cast('string')))
  .max(MAX_ROW_IDS)
  .required()
```

`[1000, 'row-2']` becomes `["1000", "row-2"]` with no error. That makes the typedef's `string[]` genuinely true at the boundary instead of picking a side of the disagreement, keeps legacy documents readable, needs no data migration, and since the same schema is on the write path any document rewritten later self-heals to strings.

I've deliberately not made that change in this PR — it's a behaviour change on a response contract, sitting in a PR whose whole point is that it changes no behaviour, and it wants a decision rather than my judgement. Happy to do it here if you'd prefer it landed with the file it belongs to; otherwise it'll be tracked separately. I also couldn't get a count of affected documents in a live environment from where I'm sitting, so I'd rather not tighten on inference alone.

Your last point stands regardless, and it's the more valuable one — nothing asserts these typedefs agree with the schemas they mirror, which is the only reason this needed archaeology at all. That's tracked separately as its own piece of work.

loveboat: Archaeology accepted, and I checked the code rather than the commit messages. `772883f0^:create-row-transformer.js:26` is `rowId: rowData[rowIdField]` raw, and `772883f0^:load-counts.js:118` appends `record.rowId` verbatim with no `String(` anywhere on that path. My "the producer stringifies" was read off today's tree, so the tightening advice was wrong -- it would reject on read, not coerce.

**The cast doesn't do what it says, though.** I ran the four variants against this repo's hapi 21.4.9 and joi 18.2.1, handler emitting `[1000, 'row-2']`:

```
current   (string|number),      no modify   -> 200 {"rowIds":[1000,"row-2"]}
tightened (string only),        no modify   -> 500
cast      (string|number.cast), no modify   -> 200 {"rowIds":[1000,"row-2"]}
cast      (string|number.cast), modify:true -> 200 {"rowIds":["1000","row-2"]}
```

hapi assigns the validated value back only when `modify` is true (`@hapi/hapi/lib/validation.js:226-235`), and `config.js:171` declares `modify` with no default -- a real route resolves to `modify=undefined`. `get.js:29` sets `response: { schema }` and nothing else, so the coerced copy is binned and the original payload is serialised. Rows 2 and 3 above are the same on the wire.

That makes the cast worse than the current honest mismatch: the schema stops complaining while the number still arrives, so it reads as fixed when the frontend still gets `1000` under a contract saying `string[]`. `rowIds.includes('1000')` is false; `rowId.startsWith(...)` throws.

**Two things that go with your diagnosis.** `sample` also resolves to `undefined` and the guard is `if (settings.response.sample)`, so every response is validated -- one bad document means that check page is dead every time, not intermittently. And the 500 is right for the right reason: response `failAction` resolves to `'error'`, and the global `routes.validate.failAction` at `server.js:64` doesn't reach it, being under `validate` rather than `response`.

**One thing that changes the shape of the risk.** There's a TTL index on the collection (`repositories/summary-logs/mongodb.js:44`, `expireAfterSeconds: 0`) and `domain/summary-logs/status.js:42-53` gives every status a TTL of a day or a week -- except `submitted`, which is `null`. So the only documents from that window still present are submitted ones, and `get.js:52` emits `loads` regardless of status. The population is much smaller than seven months of uploads implies, but it never drains: the survivors are exactly the permanent ones.

**On the count you couldn't get** -- this should give it directly:

```js
db.getCollection('summary-logs').countDocuments({ $or:
  ['added','unchanged','adjusted'].flatMap(c =>
    ['valid','invalid','included','excluded'].map(v =>
      ({ [`loads.${c}.${v}.rowIds`]: { $type: 'number' } }))) })
```

`$type: 'number'` on a dotted array path matches if any element is a number. If it comes back zero, tightening is safe and correct after all. Non-zero and it's normalise-on-read. `./sk_cdp_logs.sh epr-backend --since 30d -q 'http.response.status_code:500 AND url.path:*summary-logs*'` would show whether it's already firing.

Agreed on keeping it out of a behaviour-neutral PR. Worth the follow-up ticket saying "normalise on the read path, or `response: { schema, modify: true }` accepting that coercion then applies to the whole response" rather than the cast alone, which would land looking like a fix and change nothing.

timrutter (approved):

PR #1572, opened 2026-07-28
PAE-1790: consolidate organisation lifecycle rules in the domain

GraemeF:

Four rules about an organisation's lifecycle were decided outside the organisations domain. Three sat in the repository schema layer — an organisation cannot be approved without an approved registration, cannot be activated without a linked Defra organisation, and cancelling a registration force-cancels its linked accreditation. The fourth was in the link route handler, which decided that linking an organisation to a Defra organisation activates it, and wrote the new status inline.

All four now live in `src/domain/organisations/lifecycle.js`, in the same decide/fold shape as `src/waste-balances/domain/commands.js`: a decider returns its outcome as data, carrying a typed refusal reason, and never throws. Each caller turns a refusal into the error its own callers expect.

One lifecycle rule deliberately stays where it is. Which organisation statuses are linkable at all is applied inside the repository adapters, because the filter has to be part of the Mongo query to be selective, and it remains duplicated across the two implementations. Moving it needs the domain to own the predicate and the query to derive from it, which is a larger change than this one and is tracked separately.

## What each layer does now

`src/domain/organisations/lifecycle.js` holds `decideOrganisationStatusChange`, which folds the transition table and both status conditions into one decision; `decideOrganisationLink`, which either refuses or returns the activated organisation with the Defra organisation attached; and the registration-to-accreditation cascade, moved across unchanged with its unit tests.

`src/repositories/organisations/schema/status-transition.js` drops from 136 lines to 60. What remains is the translation from a refusal to a Boom error — one entry per reason, preserving the exact messages and status codes — plus the generic per-item transition guard, untouched. It stays at this layer because `assertOrgStatusTransition` fires inside `prepareForReplace`, which every writer of an organisation goes through, and the repository contract asserts that Boom comes out of `replace` directly.

`src/routes/v1/organisations/link.js` no longer reasons about lifecycle. It assembles the decision's inputs, maps a refusal to `Boom.conflict` and an acceptance to `replace`.

## Types

Because `noImplicitAny` is off in this repo, a green `tsc` proves nothing on its own, so every type edge this branch writes or moves was poison-tested — renamed to a non-existent typedef, checked that errors appear, restored. Both `jsconfig.typecheck.json` and `jsconfig.typecheck.tests.json` are clean.

The accepted link decision is typed `Organisation & { linkedDefraOrganisation: LinkedDefraOrganisation }`, so the route reads the linked details without an assertion — the decider always attaches them, and the type says so.

Typing the decider's inputs honestly turned out to need the write path describing itself. `decideOrganisationStatusChange`'s second parameter was annotated `Organisation` and is never one: it receives the output of `validateOrganisationUpdate`, whose schema makes `status` optional and forbids `id`. Naming that shape as `OrganisationUpdate` then showed that `collateUsers` has the same problem — it is declared to take an `Organisation` and is handed an update, whose registrations and accreditations have had their status moved into `statusHistory`. So `registration.js` and `accreditation.js` gain `RegistrationUpdate` and `AccreditationUpdate` for what the update schemas actually validate, `updateStatusHistoryForItems` becomes generic so each item keeps its own shape, and `collate-users.js` and `schema/helpers.js` carry write-path signatures instead of read-path ones. Two of those files are outside the original change; they are here because the type they had was silently wrong and this is what made it visible.

One consequence is worth flagging rather than burying. The accreditation cascade tests each accreditation's status against the set of live statuses, and it reads that status from the update payload. Once the type said `status` is optional, the existing `Set.has(undefined)` stopped compiling. It is written out as an explicit `status !== undefined` check with a test pinning it, so runtime behaviour is identical — but it means an accreditation whose update happens not to mention its status is skipped by the cascade. That is pre-existing and is tracked separately.

## Notes for review

The decision's discriminant is `outcome` rather than `status`, which the waste-balances deciders use. In this domain `status` is the name of the thing being decided about, so `decision.status === 'accepted'` beside `organisation.status === 'active'` reads poorly; the reason is recorded on the typedef.

`assertOrgStatusTransitionValid` becomes the predicate `isOrgStatusTransitionValid`. The decider needs a predicate, and a throwing twin with no callers would be dead code. Its registration and accreditation siblings still throw and are unchanged — that inconsistency is tracked separately.

`status-transition.js` no longer passes `{ statusCode: 422 }` to `Boom.badData`. That is Boom's `data` argument rather than a status override, `badData` is already 422, and nothing in `src/` reads `boom.data` — `boom-error-logger.js` carries a comment saying so deliberately. It is unobservable to any client, since Boom does not put `data` in `output.payload`.

Behaviour is otherwise unchanged: same status codes, same error messages, same cascade outcomes.


src/domain/organisations/lifecycle.js
```diff
@@ -0,0 +1,209 @@
...
+
+export const LIFECYCLE_REFUSAL = Object.freeze({
+  INVALID_TRANSITION: 'invalid-transition',
+  NO_APPROVED_REGISTRATION: 'no-approved-registration',
+  NOT_LINKED_TO_DEFRA_ORGANISATION: 'not-linked-to-defra-organisation',
+  NOT_LINKABLE: 'not-linkable'
+})
+
+/** @type {{ outcome: typeof LIFECYCLE_DECISION.ACCEPTED }} */
+const ACCEPTED = { outcome: LIFECYCLE_DECISION.ACCEPTED }
```
loveboat: Every accepted `decideOrganisationStatusChange` returns this same object. `LIFECYCLE_DECISION` and `LIFECYCLE_REFUSAL` are both `Object.freeze`d two lines above; this isn't, so a caller that annotates or mutates a returned decision corrupts every subsequent accept process-wide. `Object.freeze` costs nothing and matches the surrounding style.

src/domain/organisations/status.js
```diff
@@ -54,16 +54,13 @@ const VALID_ACC_TRANSITIONS = {
...
-      `Cannot transition organisation status from ${fromStatus} to ${toStatus}`
-    )
-  }
-}
+/**
+ * @param {string} fromStatus
+ * @param {string} toStatus
+ * @returns {boolean}
+ */
+export const isOrgStatusTransitionValid = (fromStatus, toStatus) =>
```
loveboat, line 70: An unknown `fromStatus` gives `undefined.includes(...)` -- a TypeError and a 500 rather than a refusal. The table does cover all four `ORGANISATION_STATUS` values including `ACTIVE: []`, so this needs a stored record with a missing or unrecognised status: pre-existing, probably unreachable, and identical to what `assertOrgStatusTransitionValid` did.

Raising it only because this PR newly documents "the deciders never throw" on `lifecycle.js` and newly adds `@param {string}` where `OrganisationStatus` exists (the old function carried no JSDoc at all). Typing both parameters `OrganisationStatus` makes totality a type-level guarantee instead of an assumption, and matches `status.test.js`, which already declares its table as `[OrganisationStatus, OrganisationStatus, boolean][]`.

src/domain/organisations/lifecycle.js
```diff
@@ -0,0 +1,209 @@
...
+ * @param {Organisation} organisation
+ * @returns {boolean}
+ */
+const hasApprovedRegistration = (organisation) =>
+  organisation.registrations.some(
+    (registration) => registration.status === REGISTRATION_STATUS.APPROVED
+  )
+
+/**
+ * @param {Organisation} organisation
```
loveboat: Nit: the documented parameter name doesn't exist in the signature below, which destructures. Its sibling `hasApprovedRegistration` takes the whole object -- destructure both or neither.

loveboat (commented):
Verified against `858c4d0`. Behaviour preservation holds on the points that could have slipped: decision order is identical, both conditions still judge the incoming payload rather than the existing record (the old `requireApprovedRegistration` named its parameter `existing` but was called with `updated` -- `hasApprovedRegistration(requested)` keeps the real behaviour and drops the misleading name), messages and status codes are byte-identical including the `Boom.badData(msg, { statusCode: 422 })` quirk where the second argument is `data` rather than options, and the link route still spreads `organisation` rather than the matched entry from `linkableOrganisations`.

Deleting `status-transition.test.js` loses nothing: its four cascade tests moved verbatim, and all three refusal -> Boom mappings are covered with exact messages and status codes by `repositories/organisations/contract/org-status.contract.js:42,65,129`.

Also confirmed: no stale `assertOrgStatusTransitionValid` references, `src/domain/organisations/**` non-test has zero upward imports, and `Record<OrganisationStatusRefusal, ...>` makes the error map exhaustive at type level -- a new refusal reason without a matching entry fails `tsc`.

Three comments inline; only the first has runtime bite.

PR #1575, opened 2026-07-28
PAE-1783: add unexported tonnage startup diagnostic

loveboat:

read-only startup diagnostic sizing how many stored reports the "packaging waste received but not exported" fix would change. runs before the fix lands, and doubles as the verifier after it.

## what it recomputes

- applies the AC rule per load — column S minus column T, blank T as zero — deliberately reimplemented rather than calling the live path, so it is valid either side of the fix
- reads rows at the summary log each report was built from (`source.summaryLogId`), not the current head, so findings isolate the calculation bug from later data changes
- scans monthly reports only, which confines the population to accredited exporters: a registration reports monthly once it holds an accreditation number and quarterly before that
- adds `subtractTonnage` alongside `addTonnage`, which the fix itself will also need

## what it reports

findings are split four ways:

| kind | meaning |
|---|---|
| `mismatch` | stored figure disagrees with the rule — a backfill could correct it in place |
| `source-missing` | rows are not there to recompute from, with the cause named |
| `recompute-failed` | rows resolved but not readable as tonnages |
| `lookup-failed` | reading threw — a fact about the run, not the data, and may clear on a re-run |

each report failure is contained to its own finding, so one unreadable report does not cost the run every figure already computed.

per mismatch, and totalled across the estate:

- rows in period, rows carrying unexported tonnage, rows exporting more than they received, rows with no received tonnage recorded
- affected exporters (accredited exporter registrations) and the organisations behind them, counted over mismatches only; exporters whose reports could not be resolved are reported separately

breakdowns, each emitted as a finished log line:

- **by month** — reports, net delta, and the understated/overstated split, so opposing corrections in a month do not cancel out of sight
- **by status** — separates submitted reports, which would need resubmitting, from drafts that regenerate on their own
- **largest deltas** — top corrections by size, to distinguish one outlier from estate-wide drift

## logging

CDP indexes an allowlisted set of ECS fields, so a figure logged as a property is dropped at ingest. every breakdown is therefore aggregated before it is logged. each line carries an `event.action` so it can be filtered without parsing message text, and finding lines carry `event.reference` set to the report id.

## known limitations

- only the registration's current accreditation is probed, so a report built under a superseded one resolves no rows and reads as `source-missing`
- an over-exporting row is clamped to zero and counted; how such rows should be handled is still open with the business

gated by `FEATURE_FLAG_UNEXPORTED_TONNAGE_REPORT` (default off), single-pod lock, read-only throughout.

to use:
1. flag on for one deploy
2. check logs for "Unexported tonnage" scoped to that serviceVersion
3. flag off.


src/config.js
```diff
@@ -352,23 +352,29 @@ const baseConfig = {
       default: false,
       env: 'FEATURE_FLAG_DEV_ENDPOINTS'
     },
-    staleIssuedTonnageReport: {
-      doc: 'Feature Flag: Run the estate-wide stale issued-tonnage discrepancy diagnostic on startup (PAE-1665)',
+    preCpaResubmissionBackfill: {
+      doc: 'Feature Flag: Backfill requires-resubmission on stale pre-CPA submitted reports on startup (PAE-1768).',
       format: Boolean,
       default: false,
-      env: 'FEATURE_FLAG_STALE_ISSUED_TONNAGE_REPORT'
```
dansweeting, line 359: Do we always add FFs into the `config` AND the FF port? 🤔

src/server/run-unexported-tonnage-report.test.js
```diff
@@ -0,0 +1,411 @@
...
+      )
+    )
+  })
+
+  it('logs no largest-deltas line when nothing is wrong', async () => {
+    const server = buildServer(emptyEstateApp())
+
+    await runUnexportedTonnageReport(server)
+
+    expect(logger.info).not.toHaveBeenCalledWith({
```
dansweeting: This assertion can never fail. `infoLine` (line 107) builds every log call as `{ message, event: { category, action, ... } }`, but this matcher passes a bare `{ message: expect.stringContaining('largest deltas') }`. `toHaveBeenCalledWith` matches the whole argument structurally, so a `{ message }`-only object never equals a real `{ message, event }` call, and `not.toHaveBeenCalledWith` is trivially true whether or not the line was logged. The suppression path (`if (largest.length > 0)`) is therefore untested.

Match the real shape instead, e.g.:

```js
expect(logger.info).not.toHaveBeenCalledWith(
  expect.objectContaining({
    event: expect.objectContaining({ action: 'unexported_tonnage_largest_deltas' })
  })
)
```

src/server/run-unexported-tonnage-report.test.js
```diff
@@ -0,0 +1,411 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+import { logger } from '#common/helpers/logging/logger.js'
+import { runUnexportedTonnageReport } from './run-unexported-tonnage-report.js'
+
+vi.mock('#common/helpers/logging/logger.js', () => ({
+  logger: {
+    info: vi.fn(),
+    warn: vi.fn(),
```
dansweeting: Production never calls `logger.warn` (only `info` and `error`), so this `warn: vi.fn()` mock and the two `expect(logger.warn).not.toHaveBeenCalled()` assertions (lines 167 and 207) guard a branch that does not exist. Harmless but misleading to a future reader. Suggest removing them.

src/config.js
loveboat: I don't really see the purpose of the FF port tbh. It seems to add a lot of code for little benefit, but happy to be convinced otherwise

dansweeting (approved):

dansweeting (approved):

PR #1576, opened 2026-07-28
PAE-1783: calculate not-exported tonnage per load as received minus exported

loveboat:

# ⚠️ DO NOT MERGE UNTIL LOGIC AGREED

fixes the accredited-exporter "packaging waste received but not exported" figure. the prod sizing run found 253 reports affected across 74 exporters, with a net delta of -217,049.2 tonnes: 226,279.86 overstated against 9,230.66 understated. correcting the rule mostly moves figures **down**, not up.

the calc did not implement the AC rule. it dropped any load whose `DATE_OF_EXPORT` fell in the reporting period and summed column S for the rest, so a load carrying an export date but no exported tonnage read as fully exported and contributed nothing. export *date* was standing in for export *tonnage*.

now, per load received in the period: column S minus column T, blank T as zero, summed. no load is dropped for carrying an export date.

- clamped at zero for a load reporting more exported than received — a data error, not waste owed back. decided with the business on 10 Aug; the 2,284 such rows in prod have their overshoot recorded by the over-exported loads diagnostic (#1610) rather than folded into this figure
- `startDate` / `endDate` were only used by the removed date filter, so they are dropped from `calculateTonnageReceivedNotExported` and from `aggregateWasteExported`'s params. period scoping is unchanged — `wasteReceivedRecords` is already sliced by `DATE_RECEIVED_FOR_EXPORT` in `sliceRecordsByPeriod`
- registered-only exporters are untouched: aggregation still returns `null` and the operator types the figure in on the frontend. reprocessors have no `exportActivity` at all

the accredited fixture in `exporter.test.js` moves from 0 to 63.67 — all three of its January loads carry in-period export dates, and one of them (S=44, T=50) is the exported-exceeds-received case, so the clamp is covered by real data.

this does not correct history: `getReportDetail` returns the stored document when one exists, so the 245 already-submitted reports keep their figures until each is resubmitted or backfilled. the -217,049.2 is the size of the latent discrepancy, not what changes on deploy.

no frontend change; the value is already displayed.


src/reports/domain/aggregation/aggregate-waste-exported.js
```diff
@@ -98,29 +104,81 @@ function getTonnageRepatriated(repatriatedRecords) {
...
+ *
+ * @param {Record<string, any>} data
+ * @returns {RoundedTonnage}
+ */
+function tonnageStillOnSite(data) {
+  const received = toRoundedTonnage(data.TONNAGE_RECEIVED_FOR_EXPORT)
+  const exported = toRoundedTonnage(data.TONNAGE_OF_UK_PACKAGING_WASTE_EXPORTED)
+  return greaterThan(received, exported)
+    ? subtractTonnage(received, exported)
+    : ZERO_TONNAGE
```
dansweeting, line 121: `tonnageStillOnSite` clamps a load to zero when exported > received. Per the decision with David, don't clamp: carry the negative net so over-exports stay visible. Edge case to settle: a blank received (exported present) reads as zero and would then contribute `-exported`, folding the "missing data" anomaly into the over-export one that `isOverExported` keeps separate. Stay at zero, or net negative? Either way the "clamped" / "silently discarded" wording here, in `countOverExportedLoads`, and in `warnIfLoadsOverExported` needs updating, since nothing would be discarded.

loveboat: Clamping (not to be confused with glamping) is the business direction as of today. 

`countOverExportedLoads` and `warnIfLoadsOverExported` have gone from this PR - moved into a new startup diagnostic in #1610

dansweeting (approved):

PR #1577, opened 2026-07-28
PAE-1573: Add registration and waste balance to the PRN tonnage API

GraemeF:
Ticket: [PAE-1573](https://eaflood.atlassian.net/browse/PAE-1573)

`GET /v1/prn-tonnage` now returns the registration behind each accreditation and the waste it is holding, so the report can show what an operator has in stock next to what they have issued PRNs for. Rendering the columns is a separate change under the same ticket ([epr-re-ex-admin-frontend #508](https://github.com/DEFRA/epr-re-ex-admin-frontend/pull/508)); the two deploy together.

New fields: `registrationNumber` (string), `registrationType` (`REPROCESSOR_INPUT` | `REPROCESSOR_OUTPUT` | `EXPORTER`), `wasteBalance` (number), `availableWasteBalance` (number). `organisationId` is renamed `orgId`: it has always carried the organisation's public reference number rather than the Mongo id, and `orgId` is the name the rest of the API already uses for that value. Row fields now read down the hierarchy, matching the column order agreed on the ticket.

Two things worth a reviewer's attention.

**The join runs off the PRN's own `registrationId`.** Every PRN carries it as a required field, and it is the key the ledger is written under, so the report reads the same partition the writes used. It previously derived the registration in reverse, filtering the organisation's registrations for the one pointing back at the accreditation. A row is now one (organisation, registration, accreditation), keyed on ids alone so the row key and the ledger key agree; the organisation name, accreditation number and material come from the organisation lookup rather than the snapshot each PRN carries.

**The aggregation output is parsed where it enters the application layer.** Mongo hands back an untyped document; validating it once against what the pipeline must have resolved gives the row a named type for the rest of the flow, and replaces the `$ifNull` defaults the pipeline used to apply. The response schema cannot cover this — the ledger id never appears in the response. Field rules mirror the organisation schema, so a reprocessor registration must carry `input` or `output` and an exporter is forbidden a reprocessing type.

Balances come from the ledger port (`findLatestInLedger`), one indexed point lookup per row. This follows `src/waste-balances/routes/get.js` rather than the `$lookup` in `aggregate-available-balance.js`, which reaches past the port for the ledger collection name and is a known layering breach being unpicked under [PAE-1790](https://eaflood.atlassian.net/browse/PAE-1790).

[PAE-1573]: https://eaflood.atlassian.net/browse/PAE-1573?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


timrutter (approved):

src/application/prn-tonnage/aggregate-prn-tonnage.js
```diff
@@ -113,20 +197,143 @@ const buildSortStage = () => ({
...
+  const aggregatedRows = await db
     .collection(PRNS_COLLECTION)
     .aggregate(pipeline)
     .toArray()
 
+  const rows = await Promise.all(
+    aggregatedRows.map((row) =>
+      buildReportRow(
+        ledgerRepository,
+        Joi.attempt(
```
loveboat, line 341: `Joi.attempt` throws, and it runs inside `.map()`, so the first unreportable row rejects `aggregatePrnTonnage` and the route returns 500 — every organisation's rows are lost, not just the offending one. Before this PR a row with an unresolvable organisation or tonnage band degraded to `null` and still appeared.

Probed against a real Mongo with two organisations, one healthy (balance 900, accepted PRNs) and one whose registration carries `registrationNumber: null`:

```
{"ok":false,"message":"Unreportable PRN tonnage row for accreditation ACC-acc-cancelled: \"registrationNumber\" must be a string"}
```

The healthy operator's row goes with it, and `fetchJsonFromBackend` turns the 500 into a generic error page — so the admin gets no report and no indication of which organisation caused it. Joi's default `abortEarly` also means one field of one row per attempt, so remediation is one 500 at a time.

The state is reachable. Probing the schemas directly, `registrationNumber` and `reprocessingType` are only enforced on an update carrying `status: 'approved'`:

| write | `registrationNumber` / `reprocessingType` |
|---|---|
| insert (`status` is `forbidden()`) | not required, defaults `null` |
| update, `status: 'approved'` | required |
| update, no status (the update fork makes it optional) | not required, defaults `null` |
| update, `status: 'cancelled'` | not required, defaults `null` |

A cancelled registration is reportable per `ReportableRegistration` in `#domain/organisations/registration.js` ("approved, or cancelled after approval... carries its registration number"), and cancelling cascades to its accreditation while the PRNs it already issued stay in the collection. `src/data/fixtures/common/epr-organisations/sample-organisation-1.json` also carries `registrationNumber: null` on both registrations, and no `reprocessingType` key at all on the reprocessor one.

`public-register-transformer.js` handles this by constraining the population rather than validating it — filtering to `REGISTRATION_INCLUDED_STATUSES` (approved, cancelled) and excluding `TEST_ORGANISATION_IDS`. `summary-log-uploads-report/generate-report.js` degrades instead, with `registration.registrationNumber || ''`. This pipeline does neither, so test organisations in prod — the ones most likely to hold incomplete data — can take the report down. Filtering by reportable status and dropping test orgs in the pipeline would make the strict schema safe, because the population would then be constrained to rows that must satisfy it.

src/application/prn-tonnage/aggregate-prn-tonnage.js
```diff
@@ -56,8 +130,9 @@ const buildMatchStage = () => ({
 const buildGroupStage = () => ({
   $group: {
     _id: {
-      orgId: '$organisation.id',
+      organisationId: '$organisation.id',
       orgName: '$organisation.name',
+      registrationId: '$registrationId',
       accId: '$accreditation.id',
       accNumber: '$accreditation.accreditationNumber',
       material: '$accreditation.material'
```
loveboat: Three of these six group-key fields — `orgName`, `accNumber`, `material` — are snapshots copied onto each PRN when it is created (`packaging-recycling-notes/routes/post.js:216` writes `org.companyDetails.name`), not read live. The `ledgerId` assembled in `buildAddFieldsStage` uses ids only, so the row key can split on something the balance key cannot see, and `balancesFor` then runs twice against one ledger and returns the same figure both times.

Probed with two PRNs on one accreditation differing only in the org-name snapshot, against a single ledger event of 250:

```
rowCount: 2
  Acme Reprocessing      accepted 40  wasteBalance 250
  Acme Reprocessing Ltd  accepted 10  wasteBalance 250
summedColumn: 500   actualLedgerBalance: 250
```

Tonnages still sum correctly; the balances do not, so totalling the Waste balance column double-counts. `companyDetails` is mutable through the replace path, so an operator renaming mid-year is enough to trigger it.

Grouping on ids only and taking the name, number and material from the `$lookup` would keep the row key and the ledger key aligned.

src/application/prn-tonnage/aggregate-prn-tonnage.js
```diff
@@ -113,20 +197,143 @@ const buildSortStage = () => ({
...
+export const aggregatePrnTonnage = async (db, ledgerRepository) => {
   const pipeline = buildAggregationPipeline()
 
-  const rows = await db
+  const aggregatedRows = await db
     .collection(PRNS_COLLECTION)
     .aggregate(pipeline)
     .toArray()
 
+  const rows = await Promise.all(
```
loveboat, line 337: One `findLatestInLedger` per row with no concurrency cap, across every accreditation nationally. Each lookup is index-backed and individually cheap, but the fan-out holds the connection pool for the report's duration. Worth measuring the row count against prod before this ships — and if it is large, a batch read on the ledger port would keep the layering intact rather than reaching for a `$lookup`.

loveboat (commented):
Three findings here, one on epr-re-ex-admin-frontend#508 under the same ticket.

The first two are the substantive ones: the report now fails whole rather than per-row on data the organisation schema permits, and one accreditation's waste balance can be reported on more than one row. Both verified against a real Mongo — probe output is in the comments.

src/application/prn-tonnage/aggregate-prn-tonnage.js
GraemeF: Yes we spotted this too, decided not to bundle it as it isn't new

src/application/prn-tonnage/aggregate-prn-tonnage.js
GraemeF: tldr, but I don't think there's any scenario where this can happen, so it will fail loudly on purpose to prompt investigation.

loveboat (approved):

PR #1584, opened 2026-07-30
PAE-1573: bound the PRN tonnage report's ledger read fan-out

GraemeF:
Ticket: [PAE-1573](https://eaflood.atlassian.net/browse/PAE-1573)
The PRN tonnage report gives every row its own waste-balance ledger read, and used to release them all at once. `mongoOptions` sets no `maxPoolSize`, so the MongoDB driver's default of 100 applies. The public register currently lists around 370 reprocessor and exporter registrations, so a report covering them already asks for several times the pool in one go: it checks out every connection and every other request on the pod queues behind it. Reads now go out ten at a time.

## Bounding the reads

Rows resolve in batches of ten, one batch after another, so the report holds a fixed number of connections however many accreditations it covers.

The bound is per invocation, not per process: ten concurrent calls to this route would still reach the pool ceiling. That is an accepted limit rather than an oversight. This is an admin-scoped report, and a process-wide guarantee would want a lock or a cached result rather than a larger constant.

The trade is latency. Each wave costs the slowest of its ten reads, so the report gets slower by roughly the number of waves times a point-lookup round trip. The route sets no timeout, so if the report ever grows enough for that to matter it will fail as an undiagnosed 500 — worth a timeout and a measurement before the row count grows much further, rather than something this change settles.

## Validating before reading, not during

Row validation now happens up front, before any ledger read is spent. Previously an unreportable row threw out of the `.map` after earlier rows had already started their reads, leaving those unawaited: wasted work in the ordinary case, and an unhandled rejection in the case where one of them also failed. The report still rejects the same row with the same message.

## Making unreportable rows unrepresentable

Two shapes the module could describe but could never legitimately hold:

`material` and `tonnageBand` were typed as bare strings even though the response schema already constrains both to the domain's finite sets — the inbound parse was weaker than the outbound one, so a bad value failed late at response validation rather than early with its accreditation named. Both are now the domain types, and the inbound schema constrains them to match.

A registration's processing types were one shape with an optional reprocessing type, which made "reprocessor carrying no reprocessing type" representable — the exact combination the schema already forbids, and one the type-of-registration derivation quietly folded into reprocessor-input. Exporter and reprocessor are now separate shapes, so that combination cannot be built and the reprocessor branch is exhaustive by construction.

## Why the reads stay behind the port

The alternative is to resolve the balances inside the aggregation, with a correlated `$lookup` that sorts each partition by event number and takes one. The adjacent waste balance availability report already does exactly that, in one round trip, and it would very likely be faster here too.

It is not done here because that report reaches past the ledger's repository port and imports the events collection name from the Mongo adapter directly. Copying that would spread a layering breach there is already work open to remove. Bringing the same shape behind the port — a bulk latest-per-ledger read the adapter implements — is the version worth having, and it is a fair follow-up to this change rather than a reason to hold it: it is a performance decision that wants measurements, and this change is what stops the report exhausting the pool in the meantime.

<details>
<summary>How the bound is held, and how it is guarded</summary>

`inBatchesOf` splits the rows immutably and an async `reduce` folds the batches. Awaiting the rows reported so far *before* touching the next batch is what holds each batch back: `reduce` invokes every reducer synchronously, so each one must suspend on its predecessor before calling `batch.map`, or all batches launch at once.

Three guards, each confirmed to fail against the mistake it describes. The first wraps the real Mongo ledger repository in a recorder and asserts the peak in-flight count is exactly the bound, so it fails against an unbounded fan-out, against fully sequential reads, and against a flipped fold. The second asserts row order survives the batching, which the old `Promise.all` gave for free. The third asserts no ledger read is spent at all when a later row fails validation, which fails if validation moves back inside the fan-out.

`lodash.chunk` would have been the house helper here, but it ships no type declarations and made both the batch and its rows implicitly `any` under the strict type-check configuration, in a file that was otherwise clean. The repo carries no `@types/*` dependencies, so the split is written out instead.

The type changes were checked by poisoning each edge and confirming it reddens, since `noImplicitAny` is off in this repo and a green run proves little on its own. One limit remains: `Joi.attempt` returns `any`, so the agreement between the row schema and its typedef is maintained by hand — everything downstream of that assertion is checked, the assertion itself is not. That is the existing pattern across the repo rather than anything this branch introduces.
</details>

[PAE-1573]: https://eaflood.atlassian.net/browse/PAE-1573?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/application/prn-tonnage/aggregate-prn-tonnage.js
+++ b/src/application/prn-tonnage/aggregate-prn-tonnage.js
@@ -1,5 +1,4 @@
 import Joi from 'joi'
-import chunk from 'lodash.chunk'
 import { PRN_STATUS } from '#packaging-recycling-notes/domain/model.js'
 import {
   REPROCESSING_TYPE,
@@ -49,6 +48,10 @@ export const REGISTRATION_TYPE = Object.freeze({
   REPROCESSOR_OUTPUT: 'REPROCESSOR_OUTPUT',
   EXPORTER: 'EXPORTER'
 })
+
+/**
+ * @typedef {typeof REGISTRATION_TYPE[keyof typeof REGISTRATION_TYPE]} RegistrationType
+ */
 
 /**
  * How many ledger reads the report holds in flight at once. Every report row
@@ -300,11 +303,33 @@ const balancesFor = async (ledgerRepository, ledgerId) => {
 }
 
 /**
+ * A row as the report publishes it: the hierarchy read down from organisation
+ * to accreditation, then the balances it holds, then its PRN tonnage by status.
+ *
+ * @typedef {Object} ReportRow
+ * @property {string} organisationName
+ * @property {string} orgId
+ * @property {string} registrationNumber
+ * @property {RegistrationType} registrationType
+ * @property {string} accreditationNumber
+ * @property {string} material
+ * @property {string} tonnageBand
+ * @property {number} wasteBalance
+ * @property {number} availableWasteBalance
+ * @property {number} awaitingAuthorisationTonnage
+ * @property {number} awaitingAcceptanceTonnage
+ * @property {number} awaitingCancellationTonnage
+ * @property {number} acceptedTonnage
+ * @property {number} cancelledTonnage
+ */
+
+/**
  * Reads down the hierarchy — organisation, registration, accreditation — before
  * the figures, so the response carries the report's column order.
  *
  * @param {WasteBalanceLedgerRepository} ledgerRepository
  * @param {AggregatedRow} aggregatedRow
+ * @returns {Promise<ReportRow>}
  */
 const buildReportRow = async (
   ledgerRepository,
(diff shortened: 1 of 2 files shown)


src/application/prn-tonnage/aggregate-prn-tonnage.js
```diff
@@ -322,6 +370,41 @@ const buildReportRow = async (
...
+/**
+ * Resolves the rows a batch at a time, so the report never has more than
+ * `LEDGER_READ_CONCURRENCY` ledger reads outstanding however many
+ * accreditations it covers. Awaiting the rows reported so far before touching
+ * the next batch is what holds each batch back until its predecessor lands.
+ *
+ * @param {WasteBalanceLedgerRepository} ledgerRepository
+ * @param {AggregatedRow[]} aggregatedRows
+ */
+const buildReportRows = (ledgerRepository, aggregatedRows) => {
```
loveboat, line 393: The deferral reads right, and the batching is the correct thing to ship now. Recording the follow-up's shape here since the PR body is the only place this reasoning survives, and the obvious implementation is the wrong one.

**Don't copy `findLatestSubmittedSummaryLogPerLedger`'s shape.** Its `$sort` -> `$group`/`$first` only performs because `kind_number_report` (`{kind: 1, number: -1}`) backs the `$match` on kind. A latest-per-ledger read has no `kind` to filter on, so the same shape collscans. Measured against 370 ledgers x 1000 events (370k docs): 205ms and linear in ledger depth, versus 18ms for the shape below. It spills to disk rather than erroring -- `allowDiskUseByDefault` is true from 6.0 -- so it degrades quietly instead of failing loudly. Bounding it with `$or` over the ledger ids is worse again at 657ms, since the sort still covers every event in every matched ledger.

**The shape that holds up** is `$documents` + a correlated `$lookup`, with the ledger ids arriving as data rather than as a foreign collection -- so the adapter still names only the collection it owns, and no import of `WASTE_BALANCE_EVENTS_COLLECTION_NAME` is needed:

```js
db.aggregate([
  { $documents: ledgerIds },
  { $lookup: {
      from: WASTE_BALANCE_EVENTS_COLLECTION_NAME,
      let: { orgId: '$organisationId', regId: '$registrationId', accId: '$accreditationId' },
      pipeline: [
        { $match: { $expr: { $and: [ /* the three ids */ ] } } },
        { $sort: { number: -1 } },
        { $limit: 1 },
        { $project: { _id: 0, closingBalance: 1 } }
      ],
      as: 'latest'
  } }
])
```

Explain against the same 370k-document collection: `totalKeysExamined: 370`, `totalDocsExamined: 370`, `collectionScans: 0`, `indexesUsed: ["partition_number"]`. One seek per ledger on the existing unique slot index, flat in ledger depth, no blocking sort. Percona 7.0.28/8.0.17 on CDP, so `$documents` (5.1+) is available; it needs `db.aggregate()` rather than `collection.aggregate()`, and `createMongoLedgerRepository` already has `db` in scope at line 307.

**The reason to do it is connections, not latency.** At today's row count the one-call version is 19ms against 39ms -- not worth a ticket on its own. What it changes is that a report holds one connection rather than ten, which turns the limit this PR accepts:

> The bound is per invocation, not per process: ten concurrent calls to this route would still reach the pool ceiling.

into a bound that does hold per process, without the lock or cached result that would otherwise be needed. That makes the follow-up the completion of this change rather than a separate optimisation.

**Two traps for whoever picks it up.** A ledger with no events must still report `ZERO_BALANCES` -- `$lookup` yields an empty array there, and dropping the row instead of zeroing it is a silent wrong answer, so it wants pinning in `findLatestPerLedger.contract.js` for both adapters. And results should be keyed back to rows by ledger id rather than by position; the ordering test added here would not catch a positional mismatch.

Measurements are from a throwaway local `mongo:8.3.7` seeded with the index set from `ensureLedgerCollection`, so treat them as relative rather than as predictions for prod.

loveboat (commented):
Inline comments below.

PR #1608, opened 2026-08-10
PAE-1812: Stop approved records losing their numbers, and detect any that have

GraemeF:
Ticket: [PAE-1812](https://eaflood.atlassian.net/browse/PAE-1812)
A production accreditation lost its accreditation number, and the PRN tonnage report has been returning a 500 rather than a report ever since. This closes the write path that takes the number off, and adds a startup check that finds any record already in that state.

## The write path

`PUT /v1/organisations/{id}` returns 200 and silently nulls six fields on an approved registration and accreditation when the update fragment omits the items' `status`: `accreditationNumber`, `validFrom`, `validTo` and `reprocessingType` on the accreditation, `registrationNumber` and `reprocessingType` on the registration. No status change is involved, and the status history is left reading `approved` afterwards.

The requiredness of each of those fields is conditional on the item's status. Status is a derived field that the update schemas accept as optional, so an update that omits it resolves every one of those conditions to the optional arm, where the schema default writes `null` over the stored value.

`withStatusFromExisting` supplies the stored status before validation, so the conditions stay keyed to the record as it stands and the update is rejected instead. The same omission also made `validateApprovals` and the accreditation link checks skip silently, and they now run on every update.

An accreditation is issued a number once and keeps it for the rest of its life. Only the status it is granted into requires the number, so every other state can otherwise be saved without it. `validateAccreditationNumbersRetained` compares each update against the stored record and refuses one that drops a number already held, whatever the status. That covers the reported sequence — cancel an accreditation, which keeps the number, then save it again without — and a number put on an accreditation that was never approved, which no reading of the status history could see. A stored number may still be corrected, only not removed.

## The check

`UNNUMBERED_ACCREDITATION` joins the startup organisation validation sweep. It flags an accreditation whose status history reached approved or suspended and which now carries no valid accreditation number: the number must have existed at that point, so its absence means the value was lost rather than never issued.

Reading the history rather than the current status is what catches a demoted accreditation, which keeps its PRNs and still breaks the report. The rule walks the organisation's accreditations rather than its registrations, because the reports reach an accreditation through the identifiers recorded on a PRN, and an accreditation a registration has since been pointed away from breaks them just the same.

<details>
<summary>Detail</summary>

**The assertion is a non-empty string, not merely a non-null value.** The production failure was `"accreditationNumber" must be a string`, which fires for a number as well as for `null`. Severity is `error`.

**The status history is evidence, not proof.** The admin editor accepts a corrected status on a stored entry, so an edit that replaces the active entry with one the transition rules also allow silences the rule for that accreditation.

**`src/repositories/organisations/helpers.js`** — `withStatusFromExisting`, applied to both item arrays before the update schema runs.

**`src/repositories/organisations/schema/helpers.js`** — `validateAccreditationNumbersRetained`, beside the other cross-item invariants, called with the stored accreditations alongside the update.

**`src/domain/organisations/validation/rules/unnumbered-accreditation.js`** — the rule, following the shape of its siblings: a `CODE`/`SEVERITY_LEVEL` pair, a module-local `evaluate`, one exported rule object. **`rules/index.js`** registers it.

**`src/application/summary-log-uploads-report/generate-report.test.js`** — this reached its null-number state by approving an organisation and then stripping the numbers back off, which the repository now refuses. It is rebuilt on an organisation that was never approved, which is the reachable route to the same read state.

**`src/domain/organisations/accreditation.js` and `src/repositories/organisations/port.js`** — `accreditationNumber` on a non-active accreditation now admits `null`, which is what the storage schema writes for it. The readers already handled `null` at runtime; the types said otherwise.

</details>

https://eaflood.atlassian.net/browse/PAE-1812

[PAE-1812]: https://eaflood.atlassian.net/browse/PAE-1812?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/domain/organisations/validation/rules/unnumbered-accreditation.test.js
```diff
@@ -0,0 +1,173 @@
...
+
+/** @param {AccreditationStatus[]} statuses */
+const historyOf = (...statuses) =>
+  statuses.map((status) => ({ status, updatedAt: '2026-01-01' }))
+
+const issueFor = (accreditationId) => ({
+  code: 'UNNUMBERED_ACCREDITATION',
+  severity: SEVERITY.ERROR,
+  target: { type: TARGET_TYPE.ACCREDITATION, id: accreditationId },
+  message: `Accreditation ${accreditationId} has been accredited but carries no valid accreditation number`
```
timrutter, line 23: in this case of the bug that lead onto this, it was that an accreditation was cancelled and had its accreditation number removed.  its never correct to remove the accreditation number once its set, irrespective of status.
They start out with no number, which is set when first approved, after that it should always remain

timrutter (commented):
looks good

timrutter (approved):

PR #1609, opened 2026-08-10
PAE-1813 feat(organisations): search organisations by registration and accreditation ids and numbers

timrutter:
Ticket: [PAE-1813](https://eaflood.atlassian.net/browse/PAE-1813)
## What

Extends the paginated `GET /v1/organisations` so the admin UI can find an organisation from any identifier it holds, rather than only by name.

Five new optional query params, ANDed with each other and with the existing `search`:

| Param | Match |
|---|---|
| `orgId` | business `orgId` **or** the organisation's document `_id` |
| `registrationNumber` | anchored, case-insensitive, on `registrations[].registrationNumber` |
| `registrationId` | exact on `registrations[].id` |
| `accreditationNumber` | anchored, case-insensitive, on `accreditations[].accreditationNumber` |
| `accreditationId` | exact on `accreditations[].id` |

Criteria match at organisation level with no registration↔accreditation pairing, so an unmatched reg/acc combination correctly returns zero results — that "does this combination exist?" query is the point of ANDing them.

## Why it is shaped this way

**Criteria are bounded strings, not ObjectId or number types.** An admin pasting a truncated id should get an empty result set, not a 400 from the API. Value parsing happens in the repository, where an unparseable `orgId` becomes an unsatisfiable filter rather than an error.

**`isPaginatedRequest` is extended in lockstep with the query schema.** It is load-bearing: with the schema change but without it, a criteria-only request would fall through to `findAll()` and return every organisation as a bare array — silently wrong, and the wrong response shape. There is an explicit test asserting the response is not an array.

**The filter builders are shared between adapters.** `buildFindPageFilter` (Mongo) and `buildFindPageMatcher` (in-memory) sit together in `helpers.js` so they cannot drift. The `_id` branch is the one genuine asymmetry — MongoDB stores a real `ObjectId`, the in-memory adapter stores a plain string and filters the raw cache before `_id` → `id` mapping — so each adapter supplies its own comparison. The contract tests run against both, so parity is proven rather than assumed.

**`/^[0-9a-fA-F]{24}$/` rather than `ObjectId.isValid`.** The latter also returns true for any 12-character string, after which `createFromHexString` throws — turning a mistyped id into a 500 instead of zero results. This matches what `performFindById` already does.

## Notes for review

- Sorting, paging and the response envelope are unchanged; `totalItems` reflects the filtered count.
- The unpaginated legacy `findAll` path is untouched.
- The new filters hit unindexed embedded-array fields, and the anchored case-insensitive regex could not use an index anyway. Admin-only, low traffic, small collection — accepted deliberately. If it becomes slow, index `registrations.id`, `registrations.registrationNumber`, `accreditations.id`, `accreditations.accreditationNumber`.
- Over-length criteria return 422, matching this route's existing validation behaviour.

Deploy this before the admin frontend change (PAE-1813): `querySchema` is `.unknown(false)`, so the new frontend's params would 400 against an un-updated backend.

Jira: PAE-1813

[PAE-1813]: https://eaflood.atlassian.net/browse/PAE-1813?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/routes/v1/organisations/get.test.js
```diff
@@ -23,6 +23,37 @@ const buildOrgWithName = (name) => {
...
+ *
+ * @param {{
+ *   name: string,
+ *   orgId: number,
+ *   registrationNumber: string,
+ *   accreditationNumber: string
+ * }} criteria
+ * @returns {Omit<import('#domain/organisations/model.js').Organisation, 'status'>}
+ */
+const buildOrgWithCriteria = ({
```
GraemeF: This looks duplicated

src/repositories/organisations/port.js
```diff
@@ -61,7 +61,20 @@
  */
 
 /**
- * @typedef {{ search?: string, page: number, pageSize: number }} FindPageParams
+ * Paging plus the optional search criteria. Every criterion that carries a
```
GraemeF: "page" seems like it should be secondary - maybe just `find`?

src/repositories/organisations/helpers.js
```diff
@@ -223,6 +224,225 @@ const ORS_ADMIN_LIST_PROJECTION = {
...
+  }
+  if (documentId !== null) {
+    branches.push({ _id: ObjectId.createFromHexString(documentId) })
+  }
+
+  return branches.length === 0 ? UNSATISFIABLE_FILTER : { $or: branches }
+}
+
+/**
+ * Builds the MongoDB filter for a findPage query.
```
GraemeF: This should be private to the mongodb implementation I think

src/routes/v1/organisations/get.test.js
timrutter: Fixed. `buildOrgWithCriteria` was defined in both this file and `find-page.contract.js`; it now lives in `contract/test-data.js` and both import it.

The two copies had already drifted: the contract one guarded `!== undefined` before overriding a number, this one didn't — so it set `registrationNumber: undefined` on the first registration whenever a test didn't supply one. Kept the guarded version.

`buildOrgWithName` was duplicated across the same two files as well, so I extracted that alongside it.

src/repositories/organisations/helpers.js
timrutter: Agreed, and it was worse than just misplacement: `helpers.js` is imported by `inmemory.js`, and I'd added `import { ObjectId } from 'mongodb'` to it — so the in-memory adapter was transitively pulling in the Mongo driver.

Split it:

- `buildFindPageFilter`, `buildOrgIdFilter`, `anchoredCaseInsensitive` and `UNSATISFIABLE_FILTER` are now private to `mongodb.js`, along with the `ObjectId` import.
- `buildFindPageMatcher`, `buildOrgIdMatchers` and `anchoredMatcher` are private to `inmemory.js`.
- `helpers.js` keeps only the adapter-agnostic semantics — `normaliseCriteria`, `parseOrgIdCriterion` and `anchoredPattern` — so "what counts as a match" is still defined once, but neither adapter's query representation leaks into shared code. `helpers.js` no longer references `ObjectId`.

I'd originally colocated them to stop the two implementations drifting; the `findPage` contract tests are the real guard there, and they run against both adapters, so the split doesn't lose that.

src/repositories/organisations/port.js
timrutter: I take the point — after this change the method's job is "find organisations matching these criteria", and paging is how the results come back rather than what it's for.

Three things make me want your call before renaming, since it's shared port surface:

1. Paging isn't actually optional here. `page` and `pageSize` are required in `FindPageParams`, and the return is the `{ items, page, pageSize, totalItems, totalPages }` envelope. A bare `find` would read as though you could omit them.
2. There's a sibling, `findPageForOverseasSitesAdminList`, with the same shape and naming. Renaming one and not the other loses the parallel; renaming both is a wider change than this PR.
3. `find.contract.js` already uses `describe('find')` as the grouping for `findById` and friends, so the name is informally taken.

Options as I see them:

- **(a)** `findPage` → `find`, `FindPageParams` → `FindParams` — touches the port, both adapters, `mock-repositories.js`, the route, and the contract file/describe. Arguably also `findPageForOverseasSitesAdminList` for consistency.
- **(b)** Keep `findPage`, but rename the typedef to something criteria-led, e.g. `FindOrganisationsParams`, so the params read as a query rather than as pagination options.
- **(c)** Leave it.

Happy to do (a) in this PR or a follow-up if you think it's worth the churn — I'd lean (b) as the cheap version of your point, but it's your call.

GraemeF: I don't think the ORS naming or UX is particularly good either! You choose

timrutter: have made the name change

GraemeF (approved):

PR #1610, opened 2026-08-10
PAE-1783: add over-exported loads diagnostic and gate both on operator category

loveboat:

two changes sharing the same code path.

**1. new startup diagnostic for over-exported loads.** the not-exported figure clamps a load exporting more than it received to zero, so the overshoot is discarded. the deployed unexported-tonnage diagnostic counts 2,284 such rows across 83 reports and 31 registrations, but records neither the material nor how much.

this answers the ask on the ticket ahead of the Regulators call: *identify instances where T > S, broken down by material*.

read-only, flag-gated (`FEATURE_FLAG_OVER_EXPORTED_LOADS_REPORT`, default off), under a cross-instance lock — same shape as the unexported-tonnage diagnostic. logs a line per affected report naming each load, a rollup by material carrying the instance count, the affected exporters and the overshoot, and a summary.

**2. both diagnostics now gate on operator category.** they claimed to cover accredited exporters but only inferred it from monthly cadence plus a non-nil figure. an exporter whose accreditation was cancelled still reports monthly while being registered-only, so it passed. `getOperatorCategory(registration) === OPERATOR_CATEGORY.EXPORTER` is the real gate, and costs no extra read — the registration was already fetched for ledger partitions, then discarded.

**this moves the deployed diagnostic's population.** `scanned 1113` may change; if it does, the proxies were leaking and the sizing on PAE-1783 was computed over a slightly wrong population. the −217,049.2 figure itself is unaffected. reconcile the new run's load total against 2,284 before quoting either.

notes:

- the by-material rollup sums the loads themselves, not a summary log's netted figure, which would fold masked rows away. `loads` is the instance count and sums across materials to the run's own total
- the masking count is the population the settled per-load clamp will silently absorb: a log holding an over-exported row whose ΣS − ΣT still nets non-negative. computed over the full in-period row set, since a blank-received row contributes its −T
- `loadSourceRowStates` and `resolveLedgers` move out of `unexported-tonnage.js` into `source-row-states.js`, now two diagnostics need them
- `outOfScope` is a third `SourceRowStates` outcome: not a finding, not counted in `scanned`. a read that throws *is* counted — the gate has already passed
- `wasteProcessingType` is checked before `getOperatorCategory`, which throws on an unrecognised one and would surface as a bogus lookup-failure
- a load with no received tonnage is excluded rather than counted as an over-export — a blank reads as zero, which is a different defect
- `filterRecordsByDateField` is now generic in its record type; it was discarding `rowId`. `wasteRecordStatesForHead` takes a `Pick` of the repository rather than the whole thing. both accept strictly more than before

deliberately not shipped: a by-month breakdown and a largest-overshoots line. neither was asked for, both were copied from the unexported-tonnage diagnostic, and each costs a `LOGGING_EVENT_ACTIONS` value that is awkward to retract once a saved query references it. easy to add if the call asks for them.


src/reports/monitoring/over-exported-loads.js
```diff
@@ -0,0 +1,329 @@
...
+                loads.reduce(
+                  (sum, { overshoot }) =>
+                    addTonnage(sum, toRoundedTonnage(overshoot)),
+                  ZERO_TONNAGE
+                )
+              ),
+              net: netOf(inPeriod)
+            }
+    }
+  } catch {
```
dansweeting: Silent `catch`: a per-report error is swallowed with no log and no counter, but still counted in `scanned`. A systematic failure (bug in `netOf`, mid-scan outage) then reads as `scanned N, findings 0, masked 0, total overshoot 0` — indistinguishable from a clean estate. The sibling unexported-tonnage records these as `LOOKUP_FAILED`. Suggest counting error-skips and adding `unreadable N` to the summary (and/or `logger.warn` each).

src/reports/monitoring/source-row-states.js
```diff
@@ -0,0 +1,131 @@
...
+ * Errors propagate: a registration that cannot be read is a fact about the run,
+ * not about the exporter's data, and the caller records it as such.
+ *
+ * @param {Registration} registration
+ * @param {string} organisationId
+ * @param {string} registrationId
+ * @returns {WasteBalanceLedgerId[]}
+ */
+const resolveLedgers = (registration, organisationId, registrationId) => {
+  const accreditationIds = registration?.accreditationId
```
dansweeting: `registration?.accreditationId` — the optional chaining is now dead: `isAccreditedExporter` upstream already guarantees a non-null registration.

PR #1626, opened 2026-08-13
PAE-1420: Model summary-log passthrough columns as first-class fields

dansweeting:
Ticket: [PAE-1420](https://eaflood.atlassian.net/browse/PAE-1420)
## What

Audits the five GOV.UK summary-log templates against the modelled schema fields, then closes the passthrough gap by modelling every previously-unmodelled column as a **first-class field name**.

Our table schemas validate with Joi `.unknown(true)`, so template columns land in row-state data whether or not the schema declares them. The result was an implicit contract: some columns existed in the database only because we accept any field, not because we named them. This change gives those columns explicit, named identities.

### Audit result

Across all five processing types (15 tables), 12 were already fully modelled. The gap was **34 columns in 3 tables**:

- **Exporter · RECEIVED_LOADS_FOR_EXPORT** — 25 columns (supplier, loading-site, carrier, OSR-detail, refused/stopped, references)
- **Reprocessor-output · REPROCESSED_LOADS** — 7 columns (product/haulier/customer detail)
- **Reprocessor-output · SENT_ON_LOADS** — 2 columns (`EWC_CODE`, `WEIGHBRIDGE_TICKET`)

No modelled field was found to be absent from the current templates, so the V5→V5.1 / new-V2.1 refresh removed or renamed nothing — the only drift is additive.

## How each column is modelled

- **`fields.js`** — added as a named constant, so downstream code (e.g. report-creation gating) references `FIELDS.SUPPLIER_NAME` rather than a magic string.
- **`requiredHeaders`** — added, declaring the column as a known, present-required part of the contract.
- **`unfilledValues`** — dropdown columns normalise their `'Choose option'` placeholder to `null`. The dropdown set was verified exhaustively from each sheet's data-validation ranges (not sampled): the new dropdowns are exactly `OSR_COUNTRY`, `WAS_THE_WASTE_REFUSED`, `WAS_THE_WASTE_STOPPED` (exporter), `END_OF_WASTE_STANDARDS` (reprocessor-output REPROCESSED), and `EWC_CODE` (reprocessor-output SENT_ON).
- **`validationSchema`** — intentionally **not** touched. These columns keep flowing through `.unknown(true)` with no value-format rule. Adding one would newly reject an operator's import on a malformed value (accented text, over-long free text, a stray dropdown value) where it passes today; that is explicitly out of scope. Business validation can be layered on the named fields later where a rule is warranted.

### Behavioural change

Adding these columns to `requiredHeaders` means a file that **omits one of these header markers is now fatally rejected** (`HEADER_REQUIRED`, VAL008) where it previously passed through untouched. The **values** remain unvalidated; only header presence is now enforced. This is safe for known-good uploads: every new column is present in the current GOV.UK V5.1/V2.1 templates and in the V4/V5 repo fixtures, and template version is already enforced upstream. The residual risk is a file that omits template columns entirely, which now fails fast rather than silently dropping data. An integration test pins this contract so a future template that drops one of these columns is caught.

The guardrail is honoured: no column is added to any waste-balance required set, so classification / waste-balance outcomes for existing rows are unchanged.

## Why

This is the shared prerequisite for report-creation validation (PAE-1420, and the future PAE-1280 reprocessor work), which needs to gate on some of these columns (supplier group, final-destination group). Gating on unnamed passthrough columns is unsafe — this gives that validation a solid, explicit field contract to build on.

## Notes

- Test helpers and hand-built fixture tables were extended to include the new required headers.

Jira: https://eaflood.atlassian.net/browse/PAE-1420

[PAE-1420]: https://eaflood.atlassian.net/browse/PAE-1420?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


loveboat (commented):
Functionally looks ok. 

However there are constant references to `PAE-1420` in comments and even test names/descriptions. Also quite a few large comment blocks describing either things we're not doing, or are doing and not testing. Ideally we can add some tests to describe behaviour and knock out some of that bloviating.

loveboat (approved):

PR #1630, opened 2026-08-14
PAE-1783: parse calendar dates into types

loveboat:

consolidates ad-hoc calendar-date truncation onto shared helpers and brands the result. follows on from #1623.

- adds `utcCalendarDate()` / `toCalendarDate()`, removes three hand-rolled duplicates
- routes start-of-day construction through the existing `startOfDay()`
- `accreditationWindow()` smart constructor returns a branded pair
- brands `CalendarDate` through the reports repository port


GraemeF:
On the `CalendarDate` brand at `src/reports/repository/port.js:214-216`.

Branding `Report` is the right call: the port should promise strong domain types and leave the adapter to earn them. Two places where that does not hold yet, both pointing the same way.

`PeriodicReportDoc` at `port.js:270` is a database row shape rather than a port type. Its own docstring calls it "a raw report document as the periodic-report projections select it". Because it sits in the port and picks its three dates from `Report`, it now inherits `CalendarDate`, so the one type whose purpose is to describe the un-truncated input declares the dates already truncated. `buildPeriodSummary` then destructures those values and passes them to `backCompatCalendarDate` at `helpers.js:164-168`, which is declared `@param {string}`. tsc now sees a branded value going in and a branded value coming out, so the truncation reads as redundant, and the docstring above it already says the call is safe to reduce to a plain read. That call is the only thing keeping a legacy full datetime out of a lexicographic comparison.

`findReportById` shows the same gap from the other side. The port promises `Report`, so it promises `CalendarDate`, but `performFindReportById` returns the document straight from the collection and `mapReport` only normalises `stale`. Nothing truncates on that path, and nothing validates either: `createReportSchema` reaches the database only through `validateCreateReport`, on create. A report persisted before the bare-date fix therefore arrives at every caller branded and un-truncated.

The fix that keeps the port honest is to move the document shape and the truncation inside the mongodb adapter, so `port.js` carries domain types only and the adapter converts at its edge. Worth noting on that point that `backCompatCalendarDate` currently lives in `helpers.js`, which the in-memory adapter imports too, so it runs a migration it can never need.

GraemeF (approved):

PR #1637, opened 2026-08-18
PAE-1420: Add report-completeness startup diagnostic

dansweeting:
## What

Adds a startup diagnostic that scans every live summary log across the estate on boot and logs which would fail the report-data completeness rules, without changing any user-facing behaviour. It sits behind its own feature flag `FEATURE_FLAG_REPORT_DATA_COMPLETE_DIAGNOSTIC` (default off), separate from the blocking gate's flag, so it can run in an environment independently of the gate. It runs under a cross-instance lock so a single pod per deploy executes and logs it.

When enabled it logs, message-only for cdp-logs to collate:

- one line per summary log that would be blocked: its template, material, organisation/registration/accreditation (with the human-facing registration and accreditation numbers alongside the ids), and its counts of incomplete rows and missing fields;
- one line breaking the blocked summary logs down by template;
- one line breaking them down by material, across every material;
- one line breaking down how many times each mandatory field was missing across the estate, most-missing first;
- estate-wide totals: summary logs scanned, how many would be blocked and the total missing fields, and the distinct organisations / registrations / accreditations affected, plus which templates were evaluated;
- a warning per summary log whose registration can no longer be resolved (deleted or re-versioned away): the finding is still counted, with its material and numbers reported as `unknown`, so one missing registration degrades one line rather than aborting the scan.

### Example log output

Each line below is the `message` field of a structured log record. For an estate of 128 submitted summary logs where two would be blocked: one accredited exporter carrying plastic (three rows missing an overseas-site id, one of them also missing the export date), and one registered-only exporter carrying glass (one row missing all six supplier fields):

```
Report-data diagnostic: summary log 66f0a1c3e4b0a2f1d3c40a11 (template EXPORTER, material plastic) -- org 66ef2b7a9c1d4e0012a3b456 / registration 66ef2b7a9c1d4e0012a3b789 (no. R1234567) / accreditation 66ef2b7a9c1d4e0012a3bccd (no. A7654321) -- 3 incomplete row(s), 4 missing field(s)
Report-data diagnostic: summary log 66f0a1c3e4b0a2f1d3c40a22 (template EXPORTER_REGISTERED_ONLY, material glass) -- org 66ef2b7a9c1d4e0012a3b999 / registration 66ef2b7a9c1d4e0012a3baaa (no. R7654321) / accreditation registered-only -- 1 incomplete row(s), 6 missing field(s)
Report-data diagnostic missing data by template: EXPORTER:1, EXPORTER_REGISTERED_ONLY:1
Report-data diagnostic missing data by material: aluminium:0, fibre:0, glass:1, paper:0, plastic:1, steel:0, wood:0
Report-data diagnostic missing data by field: OSR_ID:3, ACTIVITIES_CARRIED_OUT_BY_SUPPLIER:1, DATE_OF_EXPORT:1, SUPPLIER_ADDRESS:1, SUPPLIER_EMAIL:1, SUPPLIER_NAME:1, SUPPLIER_PHONE_NUMBER:1, SUPPLIER_POSTCODE:1
Report-data diagnostic summary: scanned 128 summary log(s), 2 with incomplete data (10 missing field(s)) across 2 organisation(s) / 2 registration(s) / 1 accreditation(s). Evaluated templates: EXPORTER, EXPORTER_REGISTERED_ONLY (other templates have no completeness rules yet).
```

When nothing would be blocked the breakdowns still report (all zero), and the field line reads `(none)`.

When a violating summary log's registration cannot be resolved, its per-summary-log line carries `material unknown` and `(no. unknown)`, an `unknown` entry is appended to the material breakdown, and an additional warning is logged:

```
Report-data diagnostic: could not resolve registration 66ef2b7a9c1d4e0012a3bfff (org 66ef2b7a9c1d4e0012a3beee) for summary log 66f0a1c3e4b0a2f1d3c40a33; material reported as unknown
```

## Why

The whole-summary-log completeness gate can now block report creation on incomplete rows anywhere in the summary log. Before enabling the blocking flag in an environment we need visibility of the blast radius on real data: how many operators, registrations and rows would newly be blocked, and which fields are most often missing. This diagnostic surfaces that on boot without touching user-facing behaviour.

It reuses the gate's own rules engine (`findIssues` from the report-mandatory policy) rather than reimplementing the checks, so the diagnostic and the gate can never drift. The evaluated-template set is derived from the policy registry, so when the reprocessor templates gain completeness rules (PAE-1280) the diagnostic reports on them with no further change.

### DIFF
--- a/src/reports/monitoring/report-data-completeness-diagnostic.js
+++ b/src/reports/monitoring/report-data-completeness-diagnostic.js
@@ -17,7 +17,9 @@ import { MATERIAL } from '#domain/organisations/model.js'
  * @typedef {object} CompletenessFinding
  * @property {string} organisationId
  * @property {string} registrationId
+ * @property {string} registrationNumber - Human-facing registration number ('unknown' if the registration could not be resolved).
  * @property {string | null} accreditationId
+ * @property {string | null} accreditationNumber - Human-facing accreditation number; null for registered-only or unresolved.
  * @property {string} summaryLogId
  * @property {string} processingType - The SL's template (uniform across its rows).
  * @property {string} material
@@ -37,11 +39,12 @@ import { MATERIAL } from '#domain/organisations/model.js'
  */
 
 /**
- * Material reported for a violating summary log whose registration could not be
- * resolved. A real value would be one of the `MATERIAL` enum; this stands in so
- * the finding still counts toward the estate totals and the per-material rollup.
+ * Placeholder reported for a violating summary log whose registration could not
+ * be resolved, standing in for its material and its registration/accreditation
+ * numbers. The finding still counts toward the estate totals and the per-material
+ * rollup so the blast-radius numbers stay truthful.
  */
-export const UNKNOWN_MATERIAL = 'unknown'
+export const UNKNOWN = 'unknown'
 
 /**
  * The templates the completeness gate can evaluate today, derived from the
@@ -95,15 +98,21 @@ export const findReportDataCompletenessFindings = async ({
       fieldCounts.set(issue.field, (fieldCounts.get(issue.field) ?? 0) + 1)
     }
     // A deleted or re-versioned registration must not abort the estate scan:
-    // record it, report the material as unknown, and keep the finding so the
-    // blast-radius count stays truthful.
-    let material = UNKNOWN_MATERIAL
+    // record it, report the material and numbers as unknown, and keep the
+    // finding so the blast-radius count stays truthful.
+    let material = UNKNOWN
+    let registrationNumber = UNKNOWN
+    /** @type {string | null} */
+    let accreditationNumber = null
     try {
       const registration = await organisationsRepository.findRegistrationById(
         ledgerId.organisationId,
         ledgerId.registrationId
       )
       material = registration.material
+      registrationNumber = registration.registrationNumber ?? UNKNOWN
+      accreditationNumber =
+        registration.accreditation?.accreditationNumber ?? null
     } catch {
       unresolved.push({
         organisationId: ledgerId.organisationId,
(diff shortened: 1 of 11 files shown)


src/reports/monitoring/report-data-completeness-diagnostic.js
```diff
@@ -0,0 +1,244 @@
...
+}
+
+/**
+ * @param {CompletenessFinding} finding
+ * @returns {string}
+ */
+export const formatFinding = (finding) => {
+  const accreditation = finding.accreditationId ?? 'registered-only'
+  return (
+    `Report-data diagnostic: summary log ${finding.summaryLogId} ` +
```
senthilinfy: It would be good to log organisationName and registration id(not mongo id)

dansweeting: Not sure we can log the org name - we've shied away from it in the past. I can't really say why since it's public info rather than PII. But I can log the registration/accrediation _numbers_ in addition to what we already have here, which'd be friendlier?

dansweeting: Done in 0d1e842a. Each per-summary-log line now logs the human-facing registration number and accreditation number alongside the ids, e.g. `registration 66ef... (no. R1234567) / accreditation 66ef... (no. A7654321)`. Registered-only ledgers show `accreditation registered-only`; if a registration cannot be resolved the number reads `(no. unknown)`.

I have left the organisation name out: we have deliberately kept org names out of logs, so registration/accreditation numbers are the friendlier identifier we can safely add.

senthilinfy: Organisation name is not PII. Its sth we expose in public register. Unless I am mistaken?

senthilinfy (approved):

src/reports/monitoring/report-data-completeness-diagnostic.js
dansweeting: Yeah I agree, hence my doubt. The reg numbers should be fine, though? Although I'd be interested to find the reason we've tended to leave them out:)

PR #1640, opened 2026-08-19
PAE-1824: Converge epr-backend feature flags on config.get

dansweeting:
## What

Implements [ADR-0045](https://github.com/DEFRA/epr-re-ex-service/blob/main/docs/architecture/decisions/0045-feature-flag-mechanism-across-epr-services.md) (Proposal A). epr-backend now declares and reads feature flags the same way as epr-frontend: a convict entry in `config.js` read directly via `config.get('featureFlags.<name>')`. The bespoke `src/feature-flags/` ports-and-adapters module and its Hapi plugin are removed.

## Why

Adding a feature flag to epr-backend previously touched six files and restated the flag name in seven forms that had to stay in lockstep — the `FEATURE_FLAG_*` env var, the convict key, a method in each of two adapters, a property in each of two typedefs, and entries in two test arrays — and removal was the same in reverse. The frontend has no such cost because it has no wrapper, and `server.js` already read one flag directly. This converges both services on the simpler pattern already proven in production: adding or removing a flag becomes one convict entry plus its call sites.

## Changes

- Migrated the four production read sites — `plugins/router.js`, `server/run-stale-issued-tonnage-report.js`, and the two reads in `server/run-pre-cpa-resubmission-backfill.js` — from `server.featureFlags.isXEnabled()` to `config.get('featureFlags.<name>')` on the shared convict singleton.
- Deleted the adapter module (`feature-flags.port.js`, `feature-flags.config.js`, `feature-flags.inmemory.js` and their tests) and the `feature-flags` Hapi plugin. `log-feature-flag-states.js`, which reads the config object directly, is retained.
- Removed the `featureFlags` decorations from the Hapi type definitions, and the `featureFlags` option and plugin registration from `server.js` and the test-server helper.
- Test overrides now use `config.set('featureFlags.<name>', …)` with the default restored after each test, replacing in-memory adapter injection.

## Env vars unchanged

All `FEATURE_FLAG_*` environment variable names are byte-for-byte unchanged — `src/config.js` is untouched — so cdp-app-config, journey tests and CI are unaffected. This is a behaviour-preserving refactor.


dansweeting:
## Two implementations of ADR-0045 (Proposal A) — pick one, do not merge both

Both PRs land the same ADR goal and the same deletions (the `src/feature-flags/` adapter, the `feature-flags` Hapi plugin, the `FeatureFlags` typedef and decorations). They differ only in how config reaches the flag read sites and how tests override flags.

| | #1640 (singleton + `config.set`) | #1644 (injected per-server config) |
|---|---|---|
| Production reads | module-level convict singleton | per-server convict instance injected (router plugin option, run-* argument, request decoration) |
| Test overrides | `config.set(...)` on the singleton | `createTestServer({ config: { featureFlags: {…} } })` / `getConfig({…})` |
| Isolation | global `afterEach` reset hook in `.vite/setup-files.js` | structural: each server owns its convict instance, no reset hook |
| `createTestServer` `config` option | remains a silent no-op for flags | becomes the real flag override channel |
| Diff size | +237 / -561, 64 files | +345 / -695, 68 files |

**#1640 favours a smaller production diff** (read sites just import the singleton) at the cost of shared mutable flag state guarded by `config.set` discipline plus a global reset hook.

**#1644 favours no shared mutable state** (no `config.set`, no global reset, and the `createTestServer` `config` option stops being a silent no-op) at the cost of a larger production diff, since the router, the three startup jobs and the report handler each gain a config parameter or decoration.

Both are green (full suite, 100% coverage, types, lint). This is a design preference, not a correctness call.

Sibling PR: #1644.

dansweeting:
Superseded by #1644 (merged), which landed ADR-0045 Proposal A. Closing this draft.

PR #1644, opened 2026-08-19
PAE-1824: Read feature flags from injected per-server config

dansweeting:
Ticket: [PAE-1824](https://eaflood.atlassian.net/browse/PAE-1824)
## What

Implements ADR-0045 (Proposal A) in epr-backend: the bespoke feature-flag adapter is removed and every feature-flag read goes against the convict config the server already builds. Config is **injected** to each read site rather than read from a module singleton.

## Why

[ADR-0045](https://github.com/DEFRA/epr-re-ex-service/blob/main/docs/architecture/decisions/0045-feature-flag-mechanism-across-epr-services.md) converges the EPR services on `config.get('featureFlags.<name>')`, so adding or removing a flag becomes a single convict entry plus its call sites. Injecting the per-server convict instance (rather than mutating a shared singleton in tests) means flag overrides flow through the existing `createTestServer({ config })` option, there is no shared mutable flag state, and no global reset hook is needed: isolation is structural, since each server owns its own convict instance.

This is a behaviour-preserving refactor. The `FEATURE_FLAG_*` environment variable names and the six convict flag entries are unchanged.

## Changes

- Delete the `src/feature-flags/` adapter (port, config-backed and in-memory implementations) and the `feature-flags` Hapi plugin. `log-feature-flag-states.js` is kept (it already reads config directly).
- `router` reads `options.config.get('featureFlags.devEndpoints')` from a plugin option and registers routes synchronously, replacing the `server.dependency('feature-flags', …)` wrapper.
- `run-stale-issued-tonnage-report`, `run-pre-cpa-resubmission-backfill` and `run-report-data-completeness-diagnostic` take the config instance as a second argument and read `config.get('featureFlags.<name>')` (previously `server.featureFlags.*`).
- Report creation reads `request.config.get('featureFlags.reportDataValidation')` in the POST handler and passes a boolean into `createReportForPeriod`; `report-service` no longer depends on the adapter.
- `server.js` and `create-test-server.js` decorate the per-server config onto the request and register the router with the config option; the `featureFlags` plugin registration and the previously silent `featureFlags` test option are removed.
- `hapi-types`: remove the `FeatureFlags` typedef and its request/server decorations; add a `config` property to `HapiRequest`.
- Tests override flags through `createTestServer({ config: { featureFlags: { … } } })` or a `getConfig({ … })` instance for the startup-job unit tests; no `config.set`, no global `afterEach` reset.

## Trade-offs

This deliberately trades the deleted `FeatureFlags` port's named, type-checked accessors (`isReportDataValidationEnabled()` etc.) for stringly-typed convict paths (`config.get('featureFlags.reportDataValidation')`), which have no compile-time check that the key exists or returns a boolean. That is inherent to ADR-0045 Proposal A and matches epr-frontend; the convict schema in `config.js` remains the single source of truth for the flag set.

[PAE-1824]: https://eaflood.atlassian.net/browse/PAE-1824?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dansweeting:
## Two implementations of ADR-0045 (Proposal A) — pick one, do not merge both

Both PRs land the same ADR goal and the same deletions (the `src/feature-flags/` adapter, the `feature-flags` Hapi plugin, the `FeatureFlags` typedef and decorations). They differ only in how config reaches the flag read sites and how tests override flags.

| | #1640 (singleton + `config.set`) | #1644 (injected per-server config) |
|---|---|---|
| Production reads | module-level convict singleton | per-server convict instance injected (router plugin option, run-* argument, request decoration) |
| Test overrides | `config.set(...)` on the singleton | `createTestServer({ config: { featureFlags: {…} } })` / `getConfig({…})` |
| Isolation | global `afterEach` reset hook in `.vite/setup-files.js` | structural: each server owns its convict instance, no reset hook |
| `createTestServer` `config` option | remains a silent no-op for flags | becomes the real flag override channel |
| Diff size | +237 / -561, 64 files | +345 / -695, 68 files |

**#1640 favours a smaller production diff** (read sites just import the singleton) at the cost of shared mutable flag state guarded by `config.set` discipline plus a global reset hook.

**#1644 favours no shared mutable state** (no `config.set`, no global reset, and the `createTestServer` `config` option stops being a silent no-op) at the cost of a larger production diff, since the router, the three startup jobs and the report handler each gain a config parameter or decoration.

Both are green (full suite, 100% coverage, types, lint). This is a design preference, not a correctness call.

Sibling PR: #1640.

senthilinfy (approved):

PR #1646, opened 2026-08-20
PAE-1826: Remove the regulator scope

GraemeF:
`regulator` is a role, not a scope. The role is already carried by `REGULATOR_ROLE`, so the entry in `SCOPES` duplicated it in a list that exists to name capabilities, and no route gated on it. This removes it from `SCOPES` and from the regulator's scope bundle. It was verified dead across epr-backend, epr-frontend, epr-re-ex-admin-frontend, epr-re-ex-journey-tests and cdp-app-config: nothing reads it as an authorisation decision, and epr-frontend's `src/server/auth/scopes.js` already records that it gates no route there. The regulator's other scopes and both role constants are untouched, so nothing a regulator can reach changes.

No journey-test change accompanies this. The scope gated nothing, so no scenario can observe its removal, and no journey fixture mints it into a token.

### DIFF
--- a/src/common/helpers/auth/constants.test.js
+++ b/src/common/helpers/auth/constants.test.js
@@ -75,10 +75,6 @@ describe('the regulator role', () => {
     expect(REGULATOR_SCOPES).toContain(SCOPES.organisationRead)
   })
 
-  test('carries the coarse regulator scope for what only a regulator does', () => {
-    expect(REGULATOR_SCOPES).toContain(SCOPES.regulator)
-  })
-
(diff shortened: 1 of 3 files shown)


src/common/helpers/auth/constants.test.js
```diff
@@ -8,6 +8,14 @@ import {
   SCOPES
 } from './constants.js'
 
+describe('SCOPES', () => {
+  test('every scope names a resource and an action, never a role', () => {
```
GraemeF: pointless test

atzn-dev (approved):

PR #1652, opened 2026-08-24
PAE-1826: Move the waste balance ledger reads off the admin path

GraemeF:
Ticket: [PAE-1826](https://eaflood.atlassian.net/browse/PAE-1826)
Both waste balance ledger reads move off /v1/admin onto /v1/organisations/{organisationId}/registrations/{registrationId}/waste-balance-ledger and the accreditation address beneath it. They no longer respond with the raw stored event documents: instead a `ledger` object carrying the org/reg/acc ids that identify the partition, and an `events` array whose entries each state a number, a kind, a timestamp, an actor, the opening and closing balances, and either a `summaryLog` or a `prn`. An event is addressed by its ledger and its number, so the store's own document id is no longer read, returned by the repository, or stated on the wire.

This breaks the address and the body together. The endpoint serves regulators and service maintainers and no operator, so it migrates in place rather than expanding and contracting, and it must deploy alongside DEFRA/epr-frontend#1068 and DEFRA/epr-re-ex-admin-frontend#545, which read the new address and shape.

[PAE-1826]: https://eaflood.atlassian.net/browse/PAE-1826?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/waste-balances/repository/ledger-validation.js
```diff
@@ -23,17 +20,26 @@ export const validateLedgerEventInsert = (data) => {
 }
 
 /**
+ * A read validates the shape a write states, because the store adds nothing to
```
GraemeF: No human uses words this way.

src/waste-balances/repository/ledger-mongodb.js
```diff
@@ -70,13 +70,12 @@ export async function ensureLedgerCollection(db) {
...
 
-const toLedgerEvent = (doc) => {
-  const { _id, ...rest } = doc
-  return validateLedgerEventRead({
-    id: _id.toString(),
-    ...rest
-  })
-}
+/**
+ * A reader addresses an event by its ledger and its `number`, so `_id` is the
```
GraemeF: This is unnecessary.

senthilinfy (approved):

PR #1654, opened 2026-08-25
PAE-1859: Allow admin cancellation of awaiting-acceptance PRNs/PERNs

senthilinfy:
Ticket: [PAE-1859](https://eaflood.atlassian.net/browse/PAE-1859)
Extends the admin PRN/PERN cancellation path (previously accepted-only, PAE-1823) to also cover notes stuck at awaiting acceptance. A regulator can now recover stranded tonnage when the recipient cannot accept or reject a PRN/PERN issued to them in error.

The existing balance-reversal command, audit trail, and 31-January-following-year cancellation deadline all apply unchanged to this new path.

[PAE-1859]: https://eaflood.atlassian.net/browse/PAE-1859?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/packaging-recycling-notes/domain/cancellation.js
```diff
@@ -1,11 +1,23 @@
...
 import { assertBeforeEndOfRelevantYear } from '#packaging-recycling-notes/domain/relevant-year.js'
 
+/**
+ * Prior statuses a PRN/PERN can be admin-cancelled from, and so must respect
+ * the relevant-year deadline below. `accepted` since PAE-1823; `awaiting_acceptance`
+ * added for PAE-1859 (cancelling a note stuck awaiting the recipient's response).
+ *
+ * @type {Set<import('#packaging-recycling-notes/domain/model.js').PrnStatus>}
+ */
+const ADMIN_CANCELLABLE_PREVIOUS_STATUSES = new Set([
```
dansweeting: The cancellable-status set is now declared here and again as `CANCELLABLE_STATUSES` in `routes/admin-cancel.js`. These two have to stay identical, or the deadline check and the 409 guard disagree: add a status to one only, and either a note becomes cancellable with no relevant-year enforcement, or the deadline guards a status the route never admits (dead branch). The failure is silent. Could we export the single set from this module and import it in the route, so the two guards are provably the same? Happy to block on this one.

src/packaging-recycling-notes/domain/cancellation.js
```diff
@@ -1,11 +1,23 @@
...
+  PRN_STATUS.ACCEPTED,
+  PRN_STATUS.AWAITING_ACCEPTANCE
```
dansweeting: Worth confirming this is deliberate rather than inherited: PAE-1823 tied the 31 January deadline to *accepted* notes, i.e. tonnage a producer has actually claimed against an obligation year. An `awaiting_acceptance` note was never accepted, so applying the same hard deadline means a stranded, never-accepted note becomes permanently un-cancellable after 31 January. The tests assert this (`2027-02-01` rejects), so it reads as intended, but it's a policy decision, not a mechanical extension. If it is intended, a one-line comment here saying so would stop a future reader assuming it was copied by reflex.

dansweeting (commented):
One scope question while we're in this logic: the set covers `ACCEPTED` and `AWAITING_ACCEPTANCE` but not `AWAITING_CANCELLATION`, so the signatory-completed `awaiting_cancellation -> cancelled` path still credits full tonnage back with no relevant-year check. That's pre-existing and out of scope here, but this PR is precisely about which cancellation paths respect the deadline, so: is it intended that a producer-initiated rejection completed by a signatory can settle after 31 January while an admin cancel of the same note cannot? No change needed if the answer is yes, just flagging.

For context, the balance maths checks out: an `awaiting_acceptance` note has had both `PRN_CREATED` and `PRN_ISSUED` applied and acceptance is a balance no-op, so routing through `cancelIssuedPrn`/`credit_full` lands on the same net-zero as the accepted path. Reusing the existing command is the right call, and the mapper assertions (no `acceptedAt`/`rejectedAt`) guard a real regression. The only thing I'd hold merge on is de-duplicating the status set.

src/packaging-recycling-notes/domain/cancellation.js
senthilinfy: confirmed https://defra-digital-team.slack.com/archives/C0A9YGVMM7G/p1787667971625239

dansweeting (approved):
Both review threads resolved: the cancellable-status set is now a single source of truth in `domain/cancellation.js`, imported by the route's 409 guard (no more duplication), and the 31 January deadline on awaiting-acceptance notes is confirmed intended. Approving.

